PR #12, opened 2021-05-05
console: add interactive sorting

### LINKED ISSUE #4: console: sort task list
right now, in the `top` view, it's not possible to sort the tasks by anything --- they are stored in a hashmap, and rendered in iteration order.

we should probably find a nice way to make them sortable, and (eventually) let the user control which key they're sorted by, like in real `top(1)`. it's useful to store the tasks in a hashmap for applying updates, since we can index a particular task and update its stats, or quickly remove them. ideally, we should find a way to sort them without having to copy from the hashmap into a vec every time we render the list of tasks.

one thing we could do is make the actual task _data_ refcounted, the actual task data, and then construct btrees for every sortable parameter as different sorted views of the task list, so that every sort is ready to go when the user selects it...

hawkw:
this branch adds support for sorting tasks in the top tasks view. tasks can
be sorted by their task ID, total runtime, busy time, idle time, and
number of polls, and the sorted column can be selected using the left
and right arrow keys (similar to `top(1)`). the sort direction
(ascending/descending) can be inverted using the `i` key.

in order to add sorting, it was necessary to add rudimentary plumbing
for interactivity by setting up `crossterm`'s input-handling utilities
and handling keypress events. we can add mouse support later.

here's a quick demo of sorting tasks:

[asciicast][1]

[1]: https://asciinema.org/a/X0xi6hzpEPVUaVVVgP26xdATG

closes: #4

### DIFF
--- a/console/Cargo.toml
+++ b/console/Cargo.toml
@@ -15,3 +15,5 @@ tui = { version = "0.15.0", features = ["crossterm"] }
 chrono = "0.4"
 tracing = "0.1"
 prost-types = "0.7"
+crossterm = { version = "0.19", features = ["event-stream"] }
+color-eyre = "0.5"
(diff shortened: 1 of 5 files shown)


carllerche:
Awesome :+1:

PR #16, opened 2021-05-06
fix(console): disable tui's default features

memoryruins:
otherwise, `cargo check` fails on windows with errors like the following:
```
error[E0433]: failed to resolve: maybe a missing crate `sys`?
  --> termion-1.5.6\src\lib.rs:27:9
   |
24 | pub use sys::size::terminal_size;
   |         ^^^ maybe a missing crate `sys`?

### DIFF
--- a/console/Cargo.toml
+++ b/console/Cargo.toml
@@ -11,7 +11,7 @@ tonic = { version = "0.4", features = ["transport"] }
 console-api = { path = "../console-api", features = ["transport"]}
 tokio = { version = "1", features = ["full", "rt-multi-thread"]}
 futures = "0.3"
-tui = { version = "0.15.0", features = ["crossterm"] }
+tui = { version = "0.15.0", default-features = false, features = ["crossterm"] }
 chrono = "0.4"
 tracing = "0.1"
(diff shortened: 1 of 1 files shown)


hawkw (approved):
great, thank you for the fix!

this makes me think that we probably need CI builds on Windows and macOS as well as on Linux, to ensure that the TUI builds successfully for all platforms. i'd happily merge a PR adding that as well, if you're interested...otherwise, i should make a note to do that later.

also, eventually, i think it could be good to make it possible to build the console using any TUI backend; right now, it's tightly coupled to `crossterm`. i'm not *really* sure if there are any use-cases for being able to swap out the backend, but maybe people would want to be able to build it with e.g. `curses`? so, we could look into that in a follow-up as well.

memoryruins:
> i'd happily merge a PR adding that as well, if you're interested

sounds good! I'll open one soon that adds windows and macos.

> i'm not really sure if there are any use-cases for being able to swap out the backend, but maybe people would want to be able to build it with e.g. curses?

I expect crossterm will be sufficient for most cases, but that's an option to keep in mind if someone ever wants it.

PR #21, opened 2021-05-08
Fix example on README

gnieto:
Server needs to be run via `serve` to make it work.


hawkw (approved):
thanks for catching that!

PR #22, opened 2021-05-08
Require tokio >= 1.5 on console-subscriber

gnieto:
This crate only compiles with Tokio >= 1.5, as it requires the capacity
function on `mpsc::Sender` here:

https://github.com/tokio-rs/console/blob/84c9928a7ec62bfa8b953a8fc029f33306d6cb69/console-subscriber/src/lib.rs#L154

This function was added on Tokio 1.5, according to the changelog:

https://github.com/tokio-rs/tokio/releases/tag/tokio-1.5.0

### DIFF
--- a/console-subscriber/Cargo.toml
+++ b/console-subscriber/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2018"
 
 [dependencies]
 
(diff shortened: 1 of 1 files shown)


hawkw (approved):
looks good to me --- thanks for catching that!

PR #23, opened 2021-05-09
Add example targets

zaharidichev:
Makes things easier to run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,12 @@ tools**:
 * `dump.rs`: a simple CLI program that dumps the data stream from a `Tasks`
   server
 
+Examples can be executed with:
+
+```
+cargo run --example $name
+```
+
 [`tracing`]: https://lib.rs/crates/tracing
 [`tracing-subscriber`]: https://lib.rs/crates/tracing-subscriber
 [`console-api`]: ../console-api
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ tools**:
 
 Examples can be executed with:
 
-```
+```shell
 cargo run --example $name
 ```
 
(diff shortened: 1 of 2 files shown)


console-subscriber/Cargo.toml
```diff
@@ -22,4 +22,12 @@ futures = { version = "0.3", default-features = false }
...
\ No newline at end of file
+tracing-subscriber = { version = "0.2.17", features = ["fmt", "registry", "env-filter"] }
+
+[[example]]
+name = "app"
+path = "examples/app.rs"
+
+[[example]]
+name = "dump"
+path = "examples/dump.rs"
```
hawkw: are these necessary for the examples to work? i know i've used `cargo run --example` to run them locally without having to add this to the Cargo.toml...

README.md
```diff
@@ -126,6 +126,12 @@ tools**:
 * `dump.rs`: a simple CLI program that dumps the data stream from a `Tasks`
   server
 
+Examples can be executed with:
+
+```
```
hawkw: nit:
```suggestion
```shell
```

hawkw (approved):

console-subscriber/Cargo.toml
seanmonstar: Same, I also run (from the `console-subscriber` directory) `cargo run --example app` and it works...

hawkw: > Same, I also run (from the `console-subscriber` directory) `cargo run --example app` and it works...

you should be able to do it even from the repo root, at least on Rust 1.52.0 --- i've been doing this

zaharidichev: yes in deed, this is the case. Will update

PR #24, opened 2021-05-10
Setup Task Details view

seanmonstar:
This starts adding the concept of there being multiple possible "views" of the console. To add a second view, this includes a basic "task details" view. If you hit "enter" when scrolling through the task list, it will go to the details view. It's pretty bare-bones for now, just showing the same info you would see in the task list view, but with a different layout. The point of opening this now is to simply start the process of having multiple views.

Some things we _could_ show are listed in [this doc](https://hackmd.io/a4P75rj0RP6qettxAVzv8w?view#Task-instance). I'm thinking about another PR soon that would include a list of logs (tracing events) scoped to the task to show on the details page, so you can see the logs of just a single task easily.

### DIFF
--- a/console/src/view/mod.rs
+++ b/console/src/view/mod.rs
@@ -10,6 +10,15 @@ pub(crate) enum View {
     TaskInstance(self::task::TaskView),
 }
 
+macro_rules! key {
+    ($code:ident) => {
+        input::Event::Key(input::KeyEvent {
+            code: input::KeyCode::$code,
+            ..
+        })
+    };
+}
+
 impl View {
     pub(crate) fn update_input(&mut self, event: input::Event, tasks: &mut crate::tasks::State) {
         match self {
@@ -17,10 +26,7 @@ impl View {
                 // The enter key changes views, so handle here since we can
                 // mutate the currently selected view.
                 match event {
-                    input::Event::Key(input::KeyEvent {
-                        code: input::KeyCode::Enter,
-                        ..
-                    }) => {
+                    key!(Enter) => {
                         if let Some(task) = tasks.selected_task().upgrade() {
                             *self = View::TaskInstance(self::task::TaskView::new(task));
                         }
@@ -31,7 +37,19 @@ impl View {
                     }
                 }
             }
-            View::TaskInstance(view) => view.update_input(event),
+            View::TaskInstance(view) => {
+                // The escape key changes views, so handle here since we can
+                // mutate the currently selected view.
+                match event {
+                    key!(Esc) => {
+                        *self = View::TasksList;
+                    }
+                    _ => {
+                        // otherwise pass on to view
+                        view.update_input(event);
+                    }
+                }
+            }
         }
     }
 
@@ -44,13 +62,13 @@ impl View {
         match self {
             View::TasksList => {
                 tasks.render(frame, area);
-                tasks.retain_active();
             }
             View::TaskInstance(view) => {
                 view.render(frame, area);
-                // retain_active()? should that always be done?
             }
         }
+
+        tasks.retain_active();
     }
 }
 
--- a/console/src/view/mod.rs
+++ b/console/src/view/mod.rs
@@ -1,0 +1,61 @@
+use crate::input;
+use tui::layout;
+
+mod task;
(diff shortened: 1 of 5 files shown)


console/src/view/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+        tasks: &mut crate::tasks::State,
+    ) {
+        match self {
+            View::TasksList => {
+                tasks.render(frame, area);
+                tasks.retain_active();
+            }
+            View::TaskInstance(view) => {
+                view.render(frame, area);
+                // retain_active()? should that always be done?
```
hawkw: `retain_active` drops data stored for completed tasks after a period of time...we probably want to run that here so that we don't use increasingly large amounts of memory to store data for tasks that have ended?

since the task view clones the `Rc` for the task that's being actively looked at, we won't yank it out from under the user if we continue deleting completed tasks in the background...

console/src/view/task.rs
```diff
@@ -0,0 +1,54 @@
...
+        // Rows with the following info:
+        // - Task main attributes
+        // - task metadata
+        // - metrics
+        // - logs?
+
+        let task = &*self.task.borrow();
+        const DUR_PRECISION: usize = 4;
+
+        let attrs = Spans::from(vec![Span::raw("ID: "), Span::raw(task.id_hex())]);
```
hawkw: probably worth including the task's fields as well --- these will include (for example) the location it was spawned from. also, we could show whether or not the task has completed.

console/src/view/task.rs
```diff
@@ -0,0 +1,54 @@
...
+
+        let metrics = Spans::from(vec![
```
hawkw, line 48: other metrics that could be nice to show here include:
 - the number of polls
 - time from when the task was spawned to when it was first polled (which we can compute based on data we already have stored)
 - when the task was spawned, and if it has completed, when it finished

console/src/view/task.rs
```diff
@@ -0,0 +1,54 @@
...
+            Span::raw("Total Time: "),
+            Span::from(format!("{:.prec$?}", task.total(), prec = DUR_PRECISION,)),
+            Span::raw(", "),
+            Span::raw("Busy: "),
+            Span::from(format!("{:.prec$?}", task.busy(), prec = DUR_PRECISION,)),
+            Span::raw(", "),
+            Span::raw("Idle: "),
+            Span::from(format!("{:.prec$?}", task.idle(), prec = DUR_PRECISION,)),
```
hawkw: nit: it might be nice to bold the headings ("Total Time", "Busy", "Idle" etc)  in this view?

console/src/view/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+                            *self = View::TaskInstance(self::task::TaskView::new(task));
+                        }
+                    }
+                    _ => {
+                        // otherwise pass on to view
+                        tasks.update_input(event);
+                    }
+                }
+            }
+            View::TaskInstance(view) => view.update_input(event),
```
hawkw: i feel like it could be good to have a way to get back out of the single task view and return to the task list...the escape key seems like a natural choice, or pressing enter again?

hawkw (approved):
looks good to me, this is a great start! i had some suggestions for potential future additions; we don't have to add them all right now

console/src/view/mod.rs
seanmonstar: Yea, escape is what I was thinking...

hawkw:
@seanmonstar are you interested in making any of the additional changes i suggested on this branch, or shall i merge this as-is so we can keep working on it in follow-up PRs?

(i'm fine either way, i just don't want to yank it out from under you while you're working on it)

seanmonstar:
Ya I'm going to add a few things and then merge. Just distracted :)

seanmonstar:
Or we can merge as-is and I can file a new PR. They're cheap, if you wanted to add things to the view too. :shrug:

seanmonstar:
Addressed most things. I didn't add more metrics yet, but that's an easy PR for anyone interested in joining in.

I'll merge once it's green.

PR #26, opened 2021-05-16
feat: send structured fields on the wire

### LINKED ISSUE #6: subscriber: structured span fields
right now, we just capture all span fields as a string. we should probably fix this

zaharidichev:
This PR adds structured fields to the wire format.

Fix: #6

### DIFF
--- a/console-api/src/common.rs
+++ b/console-api/src/common.rs
@@ -45,6 +45,7 @@ impl<'a> From<&'a tracing_core::Metadata<'a>> for Metadata {
             location: Some(location),
             kind: kind as i32,
             level: metadata::Level::from(*meta.level()) as i32,
+            field_names: Vec::new(),
             ..Default::default()
         }
     }
@@ -64,11 +65,11 @@ impl<'a> From<&'a std::panic::Location<'a>> for Location {
 impl fmt::Display for field::Value {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            field::Value::BoolVal(v) => write!(f, "{}", v)?,
-            field::Value::StrVal(v) => write!(f, "{}", v)?,
-            field::Value::U64Val(v) => write!(f, "{}", v)?,
-            field::Value::DebugVal(v) => write!(f, "{}", v)?,
-            field::Value::I64Val(v) => write!(f, "{}", v)?,
+            field::Value::BoolVal(v) => fmt::Display::fmt(v, f)?,
+            field::Value::StrVal(v) => fmt::Display::fmt(v, f)?,
+            field::Value::U64Val(v) => fmt::Display::fmt(v, f)?,
+            field::Value::DebugVal(v) => fmt::Display::fmt(v, f)?,
+            field::Value::I64Val(v) => fmt::Display::fmt(v, f)?,
         }
 
         Ok(())
--- a/console-api/src/common.rs
+++ b/console-api/src/common.rs
@@ -61,6 +61,31 @@ impl<'a> From<&'a std::panic::Location<'a>> for Location {
     }
 }
 
+impl fmt::Display for field::Value {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            field::Value::BoolVal(v) => write!(f, "{}", v)?,
+            field::Value::StrVal(v) => write!(f, "{}", v)?,
+            field::Value::U64Val(v) => write!(f, "{}", v)?,
+            field::Value::DebugVal(v) => write!(f, "{}", v)?,
+            field::Value::I64Val(v) => write!(f, "{}", v)?,
+        }
+
+        Ok(())
+    }
+}
+
+impl fmt::Display for Field {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let name_val = (self.name.as_ref(), self.value.as_ref());
+        if let (Some(field::Name::StrName(name)), Some(val)) = name_val {
+            write!(f, "{}={}", name, val)?;
+        }
+
+        Ok(())
+    }
+}
+
 impl fmt::Display for Location {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match (self.module_path.as_ref(), self.file.as_ref()) {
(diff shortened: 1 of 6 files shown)


proto/tasks.proto
```diff
@@ -74,11 +74,7 @@ message Task {
     // The category of task this task belongs to.
     Kind kind = 3;
 
-    // A string representation of any fields recorded about this task.
-    //
-    // NOTE: eventually, it would be nice to support structured fields in tasks;
-    // we can deprecate this when we add that.
-    string string_fields = 4;
+    repeated common.Field fields = 4;
```
hawkw, line 67: nit: it would be nice to have a comment here

proto/common.proto
```diff
@@ -25,15 +25,23 @@ message SpanId {
...
+    oneof name {
+        string str_name = 1;
+        uint64 name_idx = 2;
+    }
+    oneof value {
+        string debug_val = 3;
+        string str_val = 4;
+        uint64 u64_val = 5;
+        sint64 i64_val = 6;
+        bool bool_val = 7;
```
hawkw: nit: it would be nice to have some comments here explaining the different potential fields here. in particular, it would be nice to explain how names work, since that's potentially not clear?

proto/common.proto
```diff
@@ -25,15 +25,23 @@ message SpanId {
 }
 
 message Field {
-    oneof field {
-        string debug = 1;
+    oneof name {
+        string str_name = 1;
+        uint64 name_idx = 2;
```
hawkw, line 33: in order to actually use this, we'll also want to include the list of field names in the metadata message, so that these indices are meaningful. but we don't have to do that in this PR, that can be added in a follow-up.

hawkw: also, just a thought: it _might_ be worth having this message also include the metadata ID that the index refers to, so that a `Field` message with a name index can always be hydrated? but, since the `Field` message is only ever sent in the `Span` and `Task` messages, which also include metadata IDs, this isn't strictly _necessary_...

console-subscriber/src/lib.rs
```diff
@@ -304,3 +287,37 @@ impl proto::tasks::tasks_server::Tasks for Server {
         Ok(tonic::Response::new(stream))
     }
 }
+
+impl Visit for FieldsCollector {
```
hawkw: nit, take it or leave it: I might call this `FieldVisitor` or something --- in `tracing` 0.2, the name "Collector" will refer to what are currently called "Subscribers", so this could be a little confusing in the future.

console-subscriber/src/lib.rs
```diff
@@ -304,3 +287,37 @@ impl proto::tasks::tasks_server::Tasks for Server {
...
+            name: Some(field.name().into()),
+            value: Some(value.into()),
+        });
+    }
+    fn record_str(&mut self, field: &tracing_core::Field, value: &str) {
+        self.fields.push(proto::Field {
+            name: Some(field.name().into()),
+            value: Some(value.into()),
+        });
+    }
```
hawkw, line 333: note that [`&dyn std::error::Error` is also a potential primitive `tracing` field value](https://docs.rs/tracing/0.1.26/tracing/field/trait.Visit.html#method.record_error). 

this implementation will just record `Error`s with `fmt::Debug`, which is fine for now, but it might be nice to also allow recording `Error`s in a more structured way as a follow-up? we could have an `Error` message which includes the `Error` type's `fmt::Display` output and an optional [`source` field](https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.source), or maybe the entire chain of `source`s as a protobuf list?

right now this wouldn't be particularly useful, since we only look at the `tokio::task` spans which never record `Error` fields. but in the future, we could potentially support nice error reporting in the console, so it's something that could be worth thinking about as a follow-up change...

console/src/tasks.rs
```diff
@@ -115,13 +116,21 @@ impl State {
...
+                })
+                .trim_end()
+                .into();
             let id = task.id?.id;
             let stats = stats_update.remove(&id)?.into();
             let task = Task {
                 id,
                 id_hex: format!("{:x}", id),
-                fields: task.string_fields,
+                fields,
```
hawkw: :+1: for now, since the purpose of this change was just to add structured fields to the wire format. 

a good follow-up would be to change the console TUI to store task fields as structured values as well (maybe a hashmap of names -> values for starters?). that would enable some additional things we could do, such as sorting the task list by the values of various fields, filtering which fields are displayed, and nicer formatting (e.g. bolding field names or displaying them in a different color).

console-subscriber/src/lib.rs
hawkw: nitpicky, but: let's prefer to always put a newline between these methods:
```suggestion

    fn record_i64(&mut self, field: &tracing_core::Field, value: i64) {
        self.fields.push(proto::Field {
            name: Some(field.name().into()),
            value: Some(value.into()),
        });
    }
    
    fn record_u64(&mut self, field: &tracing_core::Field, value: u64) {
        self.fields.push(proto::Field {
            name: Some(field.name().into()),
            value: Some(value.into()),
        });
    }
    
    fn record_bool(&mut self, field: &tracing_core::Field, value: bool) {
        self.fields.push(proto::Field {
            name: Some(field.name().into()),
            value: Some(value.into()),
        });
    }
    
    fn record_str(&mut self, field: &tracing_core::Field, value: &str) {
        self.fields.push(proto::Field {
            name: Some(field.name().into()),
            value: Some(value.into()),
        });
    }
```

console-api/src/common.rs
```diff
@@ -61,6 +61,31 @@ impl<'a> From<&'a std::panic::Location<'a>> for Location {
...
 }
 
+impl fmt::Display for field::Value {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            field::Value::BoolVal(v) => write!(f, "{}", v)?,
+            field::Value::StrVal(v) => write!(f, "{}", v)?,
+            field::Value::U64Val(v) => write!(f, "{}", v)?,
+            field::Value::DebugVal(v) => write!(f, "{}", v)?,
+            field::Value::I64Val(v) => write!(f, "{}", v)?,
```
hawkw: nit: it might be slightly simpler to just call the inner value's `fmt::Display` impl directly, rather than using the `write!` macro...i think there may be a very small amount of additional overhead introduced by going through a `format_args!`, and also i think this will ensure that any format parameters (e.g. padding/width/etc) are passed through to the inner value
```suggestion
            field::Value::BoolVal(v) => fmt::Display::fmt(v, f)?,
            field::Value::StrVal(v) => fmt::Display::fmt(v, f)?,
            field::Value::U64Val(v) => fmt::Display::fmt(v, f)?,
            field::Value::DebugVal(v) => fmt::Display::fmt(v, f)?,
            field::Value::I64Val(v) => fmt::Display::fmt(v, f)?,
```

hawkw (approved):
overall, this looks great!

i commented on a couple minor style nits you may want to address, and on potential follow-up work we can do once this is merged. also, it would be nice to see comments for the additions to the protobuf...but that's also not really a blocker as i haven't been consistent about documenting the protos either. :)

proto/common.proto
```diff
@@ -54,6 +66,10 @@ message Metadata {
     Kind kind = 5;
     Level level = 6;
 
+    // The names of the key-value fields attached to the
+    // span or event this metadata is associated with.
+    repeated string field_names = 7;
```
hawkw, line 71: as a follow-up, we should actually populate this field...

hawkw (approved):
this looks great, thank you! i'll open issues for some of the follow-up changes.

PR #28, opened 2021-05-20
fix: update uncompleted tasks total time every update

seanmonstar:
The "total" time of a task was sent from the subscriber to the console
each time there was any _other_ change for the task. However, tasks that
didn't have changes wouldn't send updates, and so their total time would
lag behind. This fixes it so all tasks in the console have an up-to-date
total time, by:

- The `total` time is now only sent in a `Task` proto when the task is
  completed.
- The subscriber's "current" time is sent in every `TaskUpdate`.
- The console uses that time to calculate "total" time if
  the task isn't completed yet.

### DIFF
--- a/console/src/tasks.rs
+++ b/console/src/tasks.rs
@@ -21,6 +21,7 @@ pub(crate) struct State {
     table_state: TableState,
     selected_column: usize,
     sort_descending: bool,
+    last_updated_at: Option<SystemTime>,
 }
 
 #[derive(Debug)]
@@ -48,8 +49,8 @@ struct Stats {
     polls: u64,
     created_at: SystemTime,
     busy: Duration,
-    idle: Duration,
-    total: Duration,
+    idle: Option<Duration>,
+    total: Option<Duration>,
 }
 
 impl State {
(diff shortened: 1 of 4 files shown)


console/src/tasks.rs
```diff
@@ -105,6 +110,9 @@ impl State {
     }
 
     pub(crate) fn update_tasks(&mut self, update: proto::tasks::TaskUpdate) {
+        if let Some(now) = update.now {
+            self.last_updated_at = Some(now.into());
+        }
         let mut stats_update = update.stats_update;
```
hawkw, line 116: nit:
```suggestion
        }
        
        let mut stats_update = update.stats_update;
```

hawkw (approved):
very nice, lgtm!

in a future branch, we might also want to display the last update time in the task list view?

hawkw:
> * The `total` time is now only sent in a `Task` proto when the task is completed.

one thought: would it make more sense to just send a completion _timestamp_ in `StatsUpdate`, and have the client always calculate total times? we can change that in an additional follow-up though.

console/src/tasks.rs
```diff
@@ -105,6 +110,10 @@ impl State {
     }
 
     pub(crate) fn update_tasks(&mut self, update: proto::tasks::TaskUpdate) {
+        if let Some(now) = update.now {
+            self.last_updated_at = Some(now.into());
+        }
+        
         let mut stats_update = update.stats_update;
```
hawkw: oh okay, i guess rustfmt doesn't agree with me about this?
```suggestion
        }
        let mut stats_update = update.stats_update;
```

PR #29, opened 2021-05-21
fix: detect completed tasks even if console connects after

seanmonstar:
The "completion" of a task would result in that being sent to all watchers in a task update. However, after a flush, that status wasn't stored in a way that new watchers could detect. This changes it so that a task is determined "completed" when it includes a "total" duration, instead of being in a separate part of the `TasksUpdate` proto message.

### DIFF
--- a/console/src/tasks.rs
+++ b/console/src/tasks.rs
@@ -148,16 +148,6 @@ impl State {
                 t.updated();
             }
         }
-
-        for proto::SpanId { id } in update.completed {
-            if let Some(task) = self.tasks.get_mut(&id) {
-                let mut task = task.borrow_mut();
-                task.kind = "!";
-                task.completed_for = 1;
-            } else {
-                tracing::warn!(?id, "tried to complete a task that didn't exist");
-            }
-        }
     }
 
     pub(crate) fn render<B: tui::backend::Backend>(
--- a/console/src/tasks.rs
+++ b/console/src/tasks.rs
@@ -126,7 +126,7 @@ impl State {
 
             let id = task.id?.id;
             let stats = stats_update.remove(&id)?.into();
-            let task = Task {
+            let mut task = Task {
                 id,
                 id_hex: format!("{:x}", id),
                 fields: task.string_fields,
@@ -134,6 +134,7 @@ impl State {
                 stats,
                 completed_for: 0,
             };
+            task.updated();
             let task = Rc::new(RefCell::new(task));
             sorted.push(Rc::downgrade(&task));
             Some((id, task))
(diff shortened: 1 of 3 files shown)


console/src/tasks.rs
```diff
@@ -362,6 +355,14 @@ impl Task {
             .idle
             .unwrap_or_else(|| self.total(since) - self.busy())
     }
+
+    fn updated(&mut self) {
```
hawkw: very nitpicky, but, since this mutates the task in place, i would probably call it
```suggestion
    fn update(&mut self) {
```
i would probably expect something called `updated` to take a `Task` by value and return it.

hawkw: also a nit, but it looks like this currently only handles marking tasks as completed. my guess is that the intention is for it to also eventually do other task state updates, but if that's not the case, it might be worth giving it a name that's specific to completion?

proto/tasks.proto
```diff
@@ -37,22 +37,11 @@ message TaskUpdate {
...
-    // list in a future `TaskUpdate`.
-    //
-    // When a task ID is present in this list, the `stats_update` field of this
-    // `TaskUpdate` will contain the final statistics recorded for that task.
-    repeated common.SpanId completed = 4;
     // The system time when this update was recorded.
     //
     // This is the timestamp any durations in the included `Stats` were
     // calculated relative to.
-    google.protobuf.Timestamp now = 5;
```
hawkw, line 55: just a note: when we stabilize the protobufs (e.g. when we publish a release of `console-api` on crates.io), we will want to stop making proto changes like this; instead, we will want to remove fields [by marking their field numbers as `reserved`](https://developers.google.com/protocol-buffers/docs/proto3#updating). for now, this is fine, since we aren't yet guaranteeing interoperability across git hashes.

hawkw (approved):
this looks good to me!

as a follow-up, we will probably want to make a change on the aggregator side to remove completed tasks once they have been sent in a task update with the `closed_at` time set, so that the hashmap of span data doesn't grow infinitely...

console/src/tasks.rs
```diff
@@ -126,14 +126,15 @@ impl State {
...
-            let task = Task {
+            let mut task = Task {
                 id,
                 id_hex: format!("{:x}", id),
                 fields: task.string_fields,
                 kind,
                 stats,
                 completed_for: 0,
             };
+            task.updated();
```
hawkw: again, i would call this
```suggestion
            task.update();
```

console/src/tasks.rs
```diff
@@ -142,17 +143,9 @@ impl State {
...
-        for proto::SpanId { id } in update.completed {
-            if let Some(task) = self.tasks.get_mut(&id) {
-                let mut task = task.borrow_mut();
-                task.kind = "!";
-                task.completed_for = 1;
-            } else {
-                tracing::warn!(?id, "tried to complete a task that didn't exist");
+                let mut t = task.borrow_mut();
+                t.stats = stats.into();
+                t.updated();
```
hawkw:
```suggestion
                t.update();
```

PR #31, opened 2021-05-25
feat(subscriber): drop data for completed tasks 

hawkw:
Currently, the console subscriber will store task data for tasks even
after they complete. This means that it will use an increasing amount of
memory over time. This branch updates the aggregator task to drop data
for completed spans once it has been sent to a watcher.

If no watcher is connected, data isn't currently dropped; I was planning
to address that in a follow-up.

### DIFF
--- a/console-subscriber/examples/app.rs
+++ b/console-subscriber/examples/app.rs
@@ -3,7 +3,9 @@ use tracing_subscriber::prelude::*;
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
-    let (layer, server) = console_subscriber::TasksLayer::new();
+    let (layer, server) = console_subscriber::TasksLayer::builder()
+        .retention(Duration::from_secs(60))
+        .build();
     let filter =
         tracing_subscriber::EnvFilter::from_default_env().add_directive("tokio=trace".parse()?);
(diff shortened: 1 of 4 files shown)


seanmonstar:
A downside to this change is that any client that subscribes a second too late will never get to know about the task. If we ever wanted to include rewinding, this would make that trickier.

hawkw:
> A downside to this change is that any client that subscribes a second too late will never get to know about the task. If we ever wanted to include rewinding, this would make that trickier.

Only in the case where another client is also connected. When no client is connected, data is never dropped at all...which I also think isn't ideal, due to memory usage...but, if the client is the only one to connect, it will receive every task that has ever been spawned.

For now, I think it's pretty important that using the console isn't essentially an endless memory leak, especially when no clients are connected. If we want to do rewinding, we could change the logic for cleaning up completed tasks to only drop them if they have been closed for some time window that's long enough that a client can catch recently closed tasks.

hawkw:
@seanmonstar I've updated this PR so that we'll now retain closed tasks for a configurable time interval. By default, it's an hour, which seems long enough to allow replaying a reasonable amount of history, but still places some bound on how much memory the aggregator will consume. The builder allows tuning this if necessary.

seanmonstar (approved):

PR #35, opened 2021-05-27
console: bold field names

### LINKED ISSUE #33: console: store structured task fields
Now that task fields are transmitted over the wire, we can store them in their structured format in the console and introduce features such as filtering or nicer formatting.

Context: https://github.com/tokio-rs/console/pull/26#discussion_r637434155

zaharidichev:
This PR delays the formatting of the fields and instead
stores them on the `Task` struct. This allows for nicer formatting
options. Currently we just bold the field names in the task tables view.

fix #33

### DIFF
--- a/console-api/src/common.rs
+++ b/console-api/src/common.rs
@@ -64,6 +64,31 @@ impl<'a> From<&'a std::panic::Location<'a>> for Location {
     }
 }
 
+impl fmt::Display for field::Value {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            field::Value::BoolVal(v) => fmt::Display::fmt(v, f)?,
+            field::Value::StrVal(v) => fmt::Display::fmt(v, f)?,
+            field::Value::U64Val(v) => fmt::Display::fmt(v, f)?,
+            field::Value::DebugVal(v) => fmt::Display::fmt(v, f)?,
+            field::Value::I64Val(v) => fmt::Display::fmt(v, f)?,
+        }
+
+        Ok(())
+    }
+}
+
+impl fmt::Display for Field {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let name_val = (self.name.as_ref(), self.value.as_ref());
+        if let (Some(field::Name::StrName(name)), Some(val)) = name_val {
+            write!(f, "{}={}", name, val)?;
+        }
+
+        Ok(())
+    }
+}
+
 impl fmt::Display for Location {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match (self.module_path.as_ref(), self.file.as_ref()) {
--- a/console-api/src/common.rs
+++ b/console-api/src/common.rs
@@ -62,31 +62,6 @@ impl<'a> From<&'a std::panic::Location<'a>> for Location {
     }
 }
 
-impl fmt::Display for field::Value {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            field::Value::BoolVal(v) => fmt::Display::fmt(v, f)?,
-            field::Value::StrVal(v) => fmt::Display::fmt(v, f)?,
-            field::Value::U64Val(v) => fmt::Display::fmt(v, f)?,
-            field::Value::DebugVal(v) => fmt::Display::fmt(v, f)?,
-            field::Value::I64Val(v) => fmt::Display::fmt(v, f)?,
-        }
-
-        Ok(())
-    }
-}
-
-impl fmt::Display for Field {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let name_val = (self.name.as_ref(), self.value.as_ref());
-        if let (Some(field::Name::StrName(name)), Some(val)) = name_val {
-            write!(f, "{}={}", name, val)?;
-        }
-
-        Ok(())
-    }
-}
-
 impl fmt::Display for Location {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match (self.module_path.as_ref(), self.file.as_ref()) {
--- a/console/src/tasks.rs
+++ b/console/src/tasks.rs
@@ -6,6 +6,7 @@ use std::{
     convert::TryFrom,
     fmt,
     rc::{Rc, Weak},
+    sync::Arc,
     time::{Duration, SystemTime},
 };
 use tui::{
@@ -41,6 +42,7 @@ pub(crate) struct Task {
     id: u64,
     id_hex: String,
     fields: Vec<Field>,
+    formatted_fields: Vec<Span<'static>>,
     kind: &'static str,
     stats: Stats,
     completed_for: usize,
@@ -48,7 +50,7 @@ pub(crate) struct Task {
 
 #[derive(Debug)]
 pub(crate) struct Metadata {
-    field_names: Vec<String>,
+    field_names: Vec<Arc<str>>,
     //TODO: add more metadata as needed
 }
 
@@ -63,7 +65,7 @@ struct Stats {
 
 #[derive(Debug)]
 pub(crate) struct Field {
-    pub(crate) name: String,
+    pub(crate) name: Arc<str>,
     pub(crate) value: FieldValue,
 }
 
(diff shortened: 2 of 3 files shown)


console-api/src/common.rs
```diff
@@ -62,31 +62,6 @@ impl<'a> From<&'a std::panic::Location<'a>> for Location {
...
-impl fmt::Display for Field {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let name_val = (self.name.as_ref(), self.value.as_ref());
-        if let (Some(field::Name::StrName(name)), Some(val)) = name_val {
-            write!(f, "{}={}", name, val)?;
-        }
-
-        Ok(())
-    }
-}
```
hawkw: I think it's _probably_ worth keeping these in the `console-api` crate, even if we're not currently using them in the console, but TIOLI.

console/src/tasks.rs
```diff
@@ -114,8 +136,23 @@ impl State {
         if let Some(now) = update.now {
             self.last_updated_at = Some(now.into());
         }
+
+        if let Some(new_metada) = update.new_metadata {
+            let metas: HashMap<_, Metadata> = new_metada
```
hawkw: nit: i'd prefer either
```suggestion
        if let Some(new_metadata) = update.new_metadata {
            let metas: HashMap<_, Metadata> = new_metadata
```
or
```suggestion
        if let Some(new_meta) = update.new_metadata {
            let metas: HashMap<_, Metadata> = new_meta
```
as i feel like metadata is rarely, if ever, abbreviated as "metada"...

console/src/tasks.rs
```diff
@@ -114,8 +136,23 @@ impl State {
...
+                .metadata
+                .into_iter()
+                .map(|meta| {
+                    let id = meta.id.expect("no id").id;
+                    let metadata = meta.metadata.expect("no metadata");
+                    (id, metadata.into())
+                })
+                .collect();
+            self.metas.extend(metas);
```
hawkw: nit: [the `extend` method takes a `T: IntoIterator`](https://doc.rust-lang.org/std/iter/trait.Extend.html#tymethod.extend), and [all types implementing `Iterator` also implement `IntoIterator`](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html#impl-IntoIterator-25). so, we don't need to call `collect` here and could skip allocating an intermediate hashmap --- making the code slightly simpler and removing an unnecessary allocation.

```suggestion
            let metas = new_metada
                .metadata
                .into_iter()
                .map(|meta| {
                    let id = meta.id.expect("no id").id;
                    let metadata = meta.metadata.expect("no metadata");
                    (id, metadata.into())
                });
            self.metas.extend(metas);
```

console/src/tasks.rs
```diff
@@ -127,12 +164,23 @@ impl State {
...
+                .filter_map(|f| {
+                    let field_name = f.name.as_ref().expect("no name");
+                    let name = match field_name {
+                        proto::field::Name::StrName(n) => Some(n.clone()),
+                        proto::field::Name::NameIdx(idx) => {
+                            let meta_id = f.metadata_id.as_ref().expect("no id");
+                            metas
+                                .get(&meta_id.id)
+                                .and_then(|meta| meta.field_names.get(*idx as usize))
+                                .cloned()
```
hawkw, line 175: it might be worth representing the metadata's field names as `Arc<str>`s, so that they are cheap to clone --- once the subscriber populates the metadata field names, we can expect most spans to have field names stored as indices rather than strings, so we could significantly reduce the number of strings the console allocates if we only stored a single copy of those strings in the metadata, and cloned an `Arc` into the task's fields.

it would be fine to do this in a follow-up PR, though --- this is fine for now.

console/src/tasks.rs
```diff
@@ -127,12 +164,23 @@ impl State {
...
+                        }
+                    };
+                    let value = f.value.as_ref().expect("no value").clone().into();
+                    name.map(|name| Field { name, value })
```
hawkw: we never reuse the protobuf `Task`'s fields, so we shouldn't have to clone them here. if we moved the fields by value, we could write something like this instead:
```rust
            let fields = task
                .fields
                .drain(..)
                .filter_map(|f| {
                    let field_name = f.name.expect("no name");
                    let name = match field_name {
                        proto::field::Name::StrName(n) => Some(n),
                        proto::field::Name::NameIdx(idx) => {
                            let meta_id = f.metadata_id.as_ref().expect("no id");
                            metas
                                .get(&meta_id.id)
                                .and_then(|meta| meta.field_names.get(*idx as usize))
                                .cloned()
                        }
                    };
                    let value = f.value.expect("no value").into();
                    name.map(|name| Field { name, value })
                })
                .collect();
```

console/src/tasks.rs
```diff
@@ -182,6 +230,19 @@ impl State {
...
+                acc.extend(vec![
+                    Span::styled(
+                        f.name.clone(),
+                        Style::default().add_modifier(Modifier::BOLD),
+                    ),
+                    Span::from("="),
+                    Span::from(format!("{} ", f.value)),
+                ]);
+                acc
+            }));
```
hawkw: hmm, this change is actually a performance hit, because we are now formatting the fields again *every* time we update the view, rather than just once.

it might be worth having the `Task` struct store the formatted representation as well as the structured representation (which we would use for things like sorting/filtering by fields). we could have some logic for determining when to re-format the fields if the task span records new fields (which currently won't happen, but we may add in the future).

again, we don't _have_ to do that in this PR, but it might be nice to do in the future as an optimization?

console/src/view/task.rs
```diff
@@ -35,6 +35,16 @@ impl TaskView {
...
 
+        let fields: String = task
+            .fields()
+            .iter()
+            .fold(String::new(), |mut res, f| {
+                write!(&mut res, "{}={} ", f.name, f.value).unwrap();
+                res
+            })
+            .trim_end()
+            .into();
```
hawkw: it would be nice to do the same formatting here as we do in the tasks list view? but again, we can add that in a follow-up if you prefer.

hawkw (approved):
overall, this looks pretty good, although I had some suggestions you may want to address.

a number of these are things we could also change in follow-up PRs if you prefer --- feel free to merge this now if you'd rather address some of those suggestions in a follow-up, or let me know once you've updated this branch?

console/src/tasks.rs
```diff
@@ -39,12 +41,19 @@ enum SortBy {
 pub(crate) struct Task {
     id: u64,
     id_hex: String,
-    fields: String,
+    fields: Vec<Field>,
+    formatted_fields: Vec<Span<'static>>,
```
zaharidichev, line 45: is that `'static` correct.. 😕

console/src/tasks.rs
```diff
@@ -114,31 +138,67 @@ impl State {
         if let Some(now) = update.now {
             self.last_updated_at = Some(now.into());
         }
+
+        if let Some(new_metadata) = update.new_metadata {
+            let metas = new_metadata.metadata.into_iter().map(|meta| {
+                let id = meta.id.expect("no id").id;
+                let metadata = meta.metadata.expect("no metadata");
+                (id, metadata.into())
+            });
```
hawkw: nit/TIOLI: it's _probably_ better to skip any metadata that's empty (and maybe log a warning?), rather than using `expect` and panicking if it's empty. we could write this instead:
```suggestion
            let metas = new_metadata.metadata.into_iter().filter_map(|meta| {
                let id = meta.id?.id;
                let metadata = meta.metadata?;
                Some((id, metadata.into()))
            });
```
this will silently skip any metadata where the `id` or `metadata` field is none --- you could add logging as well.

the subscriber *should* never send us empty metadata, but if there's a bug in the subscriber implementation, or if we make backwards compatible proto changes and those fields become optional, it's probably better to keep running and skip them, rather than panicking?

not a big deal for now though.

console/src/tasks.rs
```diff
@@ -114,31 +138,67 @@ impl State {
...
+
+            let fields: Vec<Field> = task
                 .fields
-                .iter()
-                .fold(String::new(), |mut res, f| {
-                    write!(&mut res, "{} ", f).unwrap();
-                    res
+                .drain(..)
+                .filter_map(|f| {
+                    let field_name = f.name.as_ref().expect("no name");
```
hawkw: should probably just skip it if there's no name, since we are already inside of a `filter_map`...
```suggestion
                    let field_name = f.name.as_ref()?;
```

console/src/tasks.rs
```diff
@@ -114,31 +138,67 @@ impl State {
...
+                    let name: Option<Arc<str>> = match field_name {
+                        proto::field::Name::StrName(n) => Some(n.clone().into()),
+                        proto::field::Name::NameIdx(idx) => {
+                            let meta_id = f.metadata_id.as_ref().expect("no id");
```
hawkw: similarly, it might be more robust to bail if we can't find the metadata
```suggestion
                            let meta_id = f.metadata_id.as_ref()?;
```
(though we should probably log a warning saying that the subscriber sent us nonsensical data)

hawkw (approved):

PR #44, opened 2021-06-11
subscriber: emit waker stats

### LINKED ISSUE #42: subscriber: emit data about Waker events
Next step after #37.

We want to detect a few things with regards to the Wakers related to a Task:

- Time from wake to poll
-  If poll returns Pending but the Waker wasn't cloned.
   - While not fool proof, this could be a good-enough heuristic to detect a "forgot to wake".
- If wake was called from a different thread, or same thread.
- If a wake is called from within poll
  - While not wrong, could be a sign to help someone notice a task that "busy loops".

seanmonstar:
Closes #42

### DIFF
--- a/console-subscriber/src/lib.rs
+++ b/console-subscriber/src/lib.rs
@@ -259,7 +259,7 @@ where
                     self.send(Event::Waker { id, op, at });
                 }
                 _ => {
-                    tracing::warn!("unknown waker event?");
+                    tracing::warn!("unknown waker event: {:?}", event);
                 }
             }
         }
@@ -399,6 +399,7 @@ impl Visit for WakerVisitor {
             self.id = Some(span::Id::from_u64(value));
         }
     }
+
     fn record_str(&mut self, field: &tracing_core::Field, value: &str) {
         if field.name() == "op" {
             self.op = Some(match value {
--- a/console-subscriber/src/lib.rs
+++ b/console-subscriber/src/lib.rs
@@ -42,6 +42,11 @@ struct FieldVisitor {
     meta_id: proto::MetaId,
 }
 
+struct WakerVisitor {
+    id: Option<span::Id>,
+    op: Option<WakeOp>,
+}
+
 struct Watch(mpsc::Sender<Result<proto::tasks::TaskUpdate, tonic::Status>>);
 
 enum Event {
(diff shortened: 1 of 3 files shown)


console-subscriber/src/lib.rs
```diff
@@ -226,6 +243,28 @@ where
         }
     }
 
+    fn on_event(&self, event: &tracing::Event<'_>, _ctx: Context<'_, S>) {
+        let meta = event.metadata();
+        // make faster like spawn metadata pointer check?
```
hawkw, line 248: hmm, i think if we wanted to do that, we would need to store a separate metadata pointer for _each_ waker event. also, that would rely on the assumption that there's only one callsite for each waker op?

console-subscriber/src/lib.rs
```diff
@@ -226,6 +243,28 @@ where
...
+
+            match visitor {
+                WakerVisitor {
+                    id: Some(id),
+                    op: Some(op),
+                } => {
+                    self.send(Event::Waker { id, op, at });
+                }
+                _ => {
+                    tracing::warn!("unknown waker event?");
```
hawkw: should we log the unknown op in this case? not a blocker.

console-subscriber/src/lib.rs
```diff
@@ -349,3 +388,26 @@ impl Visit for FieldVisitor {
...
+    fn record_debug(&mut self, _: &field::Field, _: &dyn std::fmt::Debug) {
+        // don't care (yet?)
+    }
+
+    fn record_u64(&mut self, field: &tracing_core::Field, value: u64) {
+        if field.name() == "task.id" {
+            self.id = Some(span::Id::from_u64(value));
+        }
+    }
+    fn record_str(&mut self, field: &tracing_core::Field, value: &str) {
```
hawkw: nit: missing whitespace here
```suggestion
    }
    
    fn record_str(&mut self, field: &tracing_core::Field, value: &str) {
```

hawkw (approved):
lovely, this is great! i had some tiny, non-blocking nitpicks.

console-subscriber/src/lib.rs
seanmonstar: Yea exactly. One thing that I noticed _is_ staying the same is the `meta.target()` pointer, since that static string is only in 1 place in the Tokio crate. But it could also be fragile...

PR #47, opened 2021-06-15
feat(subscriber): Record and send poll times with HdrHistogram

### LINKED ISSUE #36: subscriber: record and send a histogram of Task poll times
We currently send the total elapsed time that a task spends inside its `poll` function, but it'd be helpful to also include more statistics. This would help to see average poll time, which would make it easier to identify tasks that are "blocking" too long in their `poll` functions. It would also help to identify when only _sometimes_ a poll takes too long. And we could eventually use this data in the console to show a warning of tasks with poll times that are too high.

Using a histogram to record this seems like a good balance, since a simple average can lose the spikiness of some polls. I imagine we can use something like [hdrhistogram](https://crates.io/crates/hdrhistogram).

oguzbilgener:
Closes #36 

As suggested in the issue, this PR uses `HdrHistogram` to record task poll times as nanoseconds. The histogram is serialized into bytes using the `V2Serializer`. 

The subscriber now offers a second RPC stream called TaskDetails for a given task ID. 

The console app starts a details stream when you view the details of a task and stops the stream when you leave it.

To demonstrate that it works, I also added some stats and a small chart to the task view using the histogram data.
But I won't insist on merging this part, since you may have other plans for the task details view. 

Screenshot from 2021-06-22 21-20-43

I've got one TODO left and I would appreciate your suggestions on it.

### DIFF
--- a/console-subscriber/src/aggregator.rs
+++ b/console-subscriber/src/aggregator.rs
@@ -110,7 +110,7 @@ impl Default for Stats {
             waker_clones: 0,
             waker_drops: 0,
             last_wake: None,
-            poll_times_histogram: Histogram::<u64>::new(1).unwrap(),
+            poll_times_histogram: Histogram::<u64>::new(2).unwrap(),
         }
     }
 }
--- a/console-subscriber/src/aggregator.rs
+++ b/console-subscriber/src/aggregator.rs
@@ -207,7 +207,7 @@ impl Aggregator {
                                     }
 
                                 }
-                                // If task is not found, drop the subscription
+                                // If task is not found, drop the subscription which should close the stream
                             },
                         };
 
--- a/console-subscriber/src/aggregator.rs
+++ b/console-subscriber/src/aggregator.rs
@@ -163,7 +163,7 @@ impl Aggregator {
                 subscription = self.rpcs.recv() => {
                     if let Some(subscription) = subscription {
                         match subscription {
-                            WatchKind::TaskUpdate(subscription) => {
+                            WatchKind::Task(subscription) => {
                                 tracing::debug!("new tasks subscription");
                                 let new_tasks = self.tasks.all().map(|(id, task)| {
                                     task.to_proto(id.clone())
@@ -185,7 +185,7 @@ impl Aggregator {
                                     self.watchers.push(subscription)
                                 }
                             },
-                            WatchKind::TaskDetailUpdate(task_id, subscription) => {
+                            WatchKind::TaskDetail(task_id, subscription) => {
                                 tracing::debug!(id = ?task_id, "new task details subscription");
                                 let task_id: span::Id = task_id.into();
                                 if let Some(stats) = self.stats.find(&task_id) {
--- a/console-subscriber/src/aggregator.rs
+++ b/console-subscriber/src/aggregator.rs
@@ -164,53 +164,12 @@ impl Aggregator {
                     if let Some(subscription) = subscription {
                         match subscription {
                             WatchKind::Task(subscription) => {
-                                tracing::debug!("new tasks subscription");
-                                let new_tasks = self.tasks.all().map(|(id, task)| {
-                                    task.to_proto(id.clone())
-                                }).collect();
-                                let now = SystemTime::now();
-                                let stats_update = self.stats.all().map(|(id, stats)| {
-                                    (id.into_u64(), stats.to_proto())
-                                }).collect();
-                                // Send the initial state --- if this fails, the subscription is
-                                // already dead.
-                                if subscription.update(&proto::tasks::TaskUpdate {
-                                    new_metadata: Some(proto::RegisterMetadata {
-                                        metadata: self.all_metadata.clone(),
-                                    }),
-                                    new_tasks,
-                                    stats_update,
-                                    now: Some(now.into()),
-                                }) {
-                                    self.watchers.push(subscription)
-                                }
+                                self.add_task_subscription(subscription);
                             },
                             WatchKind::TaskDetail(task_id, subscription) => {
-                                tracing::debug!(id = ?task_id, "new task details subscription");
-                                let task_id: span::Id = task_id.into();
-                                if let Some(stats) = self.stats.find(&task_id) {
-                                    let now = SystemTime::now();
-                                    // Send the initial state --- if this fails, the subscription is
-                                    // already dead.
-                                    if subscription.update(&proto::tasks::TaskDetails {
-                                        task_id: Some(task_id.clone().into()),
-                                        now: Some(now.into()),
-                                        details: Some(proto::tasks::Details {
-                                            poll_times_histogram: serialize_histogram(&stats.poll_times_histogram).ok()
-                                        })
-                                    }) {
-                                        self
-                                            .details_watchers
-                                            .entry(task_id)
-                                            .or_insert_with(Vec::new)
-                                            .push(subscription);
-                                    }
-
-                                }
-                                // If task is not found, drop the subscription which should close the stream
+                                self.add_task_detail_subscription(task_id, subscription);
                             },
                         };
-
                     } else {
                         tracing::debug!("rpc channel closed, terminating");
                         return;
@@ -251,6 +210,61 @@ impl Aggregator {
             // been sent off.
             self.drop_closed_tasks();
         }
+    }
+
+    /// Add the task subscription to the watchers after sending the first update
+    fn add_task_subscription(&mut self, subscription: Watch<proto::tasks::TaskUpdate>) {
+        tracing::debug!("new tasks subscription");
+        let new_tasks = self
+            .tasks
+            .all()
+            .map(|(id, task)| task.to_proto(id.clone()))
+            .collect();
+        let now = SystemTime::now();
+        let stats_update = self
+            .stats
+            .all()
+            .map(|(id, stats)| (id.into_u64(), stats.to_proto()))
+            .collect();
+        // Send the initial state --- if this fails, the subscription is already dead
+        if subscription.update(&proto::tasks::TaskUpdate {
+            new_metadata: Some(proto::RegisterMetadata {
+                metadata: self.all_metadata.clone(),
+            }),
+            new_tasks,
+            stats_update,
+            now: Some(now.into()),
+        }) {
+            self.watchers.push(subscription)
+        }
+    }
+
+    /// Add the task details subscription to the watchers after sending the first update,
+    /// if the task is found.
+    fn add_task_detail_subscription(
+        &mut self,
+        task_id: proto::SpanId,
+        subscription: Watch<proto::tasks::TaskDetails>,
+    ) {
+        tracing::debug!(id = ?task_id, "new task details subscription");
+        let task_id: span::Id = task_id.into();
+        if let Some(stats) = self.stats.find(&task_id) {
+            let now = SystemTime::now();
+            // Send the initial state --- if this fails, the subscription is already dead.
+            if subscription.update(&proto::tasks::TaskDetails {
+                task_id: Some(task_id.clone().into()),
+                now: Some(now.into()),
+                details: Some(proto::tasks::Details {
+                    poll_times_histogram: serialize_histogram(&stats.poll_times_histogram).ok(),
+                }),
+            }) {
+                self.details_watchers
+                    .entry(task_id)
+                    .or_insert_with(Vec::new)
+                    .push(subscription);
+            }
+        }
+        // If the task is not found, drop the subscription which should close the stream.
     }
 
     /// Publish the current state to all active watchers.
--- a/console-subscriber/src/aggregator.rs
+++ b/console-subscriber/src/aggregator.rs
@@ -5,6 +5,7 @@ use tokio::sync::{mpsc, Notify};
 use futures::FutureExt;
 use std::{
     collections::HashMap,
+    convert::TryInto,
     mem,
     ops::{Deref, DerefMut},
     sync::{
@@ -14,6 +15,12 @@ use std::{
     time::{Duration, SystemTime},
 };
 use tracing_core::{span, Metadata};
+
+use hdrhistogram::{
+    serialization::{Serializer, V2DeflateSerializeError, V2DeflateSerializer},
+    Histogram,
+};
+
 pub(crate) struct Aggregator {
     /// Channel of incoming events emitted by `TaskLayer`s.
     events: mpsc::Receiver<Event>,
@@ -56,7 +63,6 @@ pub(crate) struct Flush {
     pub(crate) triggered: AtomicBool,
 }
 
-#[derive(Default)]
 struct Stats {
     // task stats
     polls: u64,
@@ -72,6 +78,8 @@ struct Stats {
     waker_clones: u64,
     waker_drops: u64,
     last_wake: Option<SystemTime>,
+
+    poll_times_histogram: Histogram<u64>,
 }
 
 #[derive(Default)]
@@ -82,6 +90,25 @@ struct TaskData<T> {
 struct Task {
     metadata: &'static Metadata<'static>,
     fields: Vec<proto::Field>,
+}
+
+impl Default for Stats {
+    fn default() -> Self {
+        Stats {
+            polls: 0,
+            current_polls: 0,
+            created_at: None,
+            first_poll: None,
+            last_poll: None,
+            busy_time: Default::default(),
+            closed_at: None,
+            wakes: 0,
+            waker_clones: 0,
+            waker_drops: 0,
+            last_wake: None,
+            poll_times_histogram: Histogram::<u64>::new(1).unwrap(),
+        }
+    }
 }
 
 impl Aggregator {
@@ -274,7 +301,12 @@ impl Aggregator {
                 stats.current_polls -= 1;
                 if stats.current_polls == 0 {
                     if let Some(last_poll) = stats.last_poll {
-                        stats.busy_time += at.duration_since(last_poll).unwrap();
+                        let elapsed = at.duration_since(last_poll).unwrap();
+                        stats.busy_time += elapsed;
+                        stats
+                            .poll_times_histogram
+                            .record(elapsed.as_nanos().try_into().unwrap_or(u64::MAX))
+                            .unwrap();
                     }
                 }
             }
@@ -482,6 +514,7 @@ impl Stats {
             waker_clones: self.waker_clones,
             waker_drops: self.waker_drops,
             last_wake: self.last_wake.map(Into::into),
+            poll_times_histogram: serialize_histogram(&self.poll_times_histogram).ok(),
         }
     }
 }
@@ -497,4 +530,11 @@ impl Task {
             fields: self.fields.clone(),
         }
     }
+}
+
+fn serialize_histogram(histogram: &Histogram<u64>) -> Result<Vec<u8>, V2DeflateSerializeError> {
+    let mut serializer = V2DeflateSerializer::new();
+    let mut buf = Vec::new();
+    serializer.serialize(histogram, &mut buf)?;
+    Ok(buf)
 }
--- a/console-subscriber/src/aggregator.rs
+++ b/console-subscriber/src/aggregator.rs
@@ -110,6 +110,8 @@ impl Default for Stats {
             waker_clones: 0,
             waker_drops: 0,
             last_wake: None,
+            // significant figures should be in the [0-5] range and memory usage
+            // grows exponentially with higher a sigfig
             poll_times_histogram: Histogram::<u64>::new(2).unwrap(),
         }
     }
@@ -163,19 +165,18 @@ impl Aggregator {
 
                 // a new client has started watching!
                 subscription = self.rpcs.recv() => {
-                    if let Some(subscription) = subscription {
-                        match subscription {
-                            WatchKind::Task(subscription) => {
-                                self.add_task_subscription(subscription);
-                            },
-                            WatchKind::TaskDetail(watch_request) => {
-                                self.add_task_detail_subscription(watch_request);
-                            },
-                        };
-                    } else {
-                        tracing::debug!("rpc channel closed, terminating");
-                        return;
-                    }
+                    match subscription {
+                        Some(WatchKind::Tasks(subscription)) => {
+                            self.add_task_subscription(subscription);
+                        },
+                        Some(WatchKind::TaskDetail(watch_request)) => {
+                            self.add_task_detail_subscription(watch_request);
+                        },
+                        _ => {
+                            tracing::debug!("rpc channel closed, terminating");
+                            return;
+                        }
+                    };
 
                     false
                 }
@@ -254,7 +255,7 @@ impl Aggregator {
         } = watch_request;
         tracing::debug!(id = ?id, "new task details subscription");
         let task_id: span::Id = id.into();
-        if let Some(stats) = self.stats.find(&task_id) {
+        if let Some(stats) = self.stats.get(&task_id) {
             let (tx, rx) = mpsc::channel(buffer);
             let subscription = Watch(tx);
             let now = SystemTime::now();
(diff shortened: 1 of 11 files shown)


console-subscriber/src/aggregator.rs
```diff
@@ -498,3 +531,10 @@ impl Task {
         }
     }
 }
+
+fn serialize_histogram(histogram: &Histogram<u64>) -> Result<Vec<u8>, V2DeflateSerializeError> {
+    let mut serializer = V2DeflateSerializer::new();
```
hawkw: what's the motivation for the V2 + DEFLATE serializer, versus the V2 without compression? I'm assuming the intention is to reduce the wire size of the histogram. looking at the [`hdrhistogram` docs, it seems like adding the compression has a significant performance penalty](https://docs.rs/hdrhistogram/7.3.0/hdrhistogram/serialization/index.html#performance-concerns) --- they seem to mostly recommend using compression for archival storage.

not necessarily saying this is the wrong choice, but it seems like it would be good to make sure we have a clear understanding of the performance tradeoffs, before we commit to having a particular serialization format be part of the API. of course, we can break the protobuf format as much as we like, currently, because we haven't committed to a stable release yet...but this will be harder in the future.

proto/tasks.proto
```diff
@@ -127,4 +127,7 @@ message Stats {
     //
     // If this is `None`, the task has not yet been woken.
     optional google.protobuf.Timestamp last_wake = 10;
+
+    // HdrHistogram.rs `Histogram` serialized to binary in the V2 + DEFLATE format
+    optional bytes poll_times_histogram = 11;
```
hawkw: adding the poll time histogram to `every` task stats update we send will probably increase the wire size of the update messages significantly --- although `hdrhistogram` is fairly compact, this is still potentially a lot of data.

here's an alternative proposal: when viewing a list of all tasks, i don't think we're going to actually care about the entire histogram. i think that we'll mostly want the histograms when we are displaying more detailed stats about individual task(s). one idea i've been considering is adding a new RPC to the [`Tasks` service](https://github.com/tokio-rs/console/blob/4ba86c027294503b3444ccd725bcc7a88562ba50/proto/tasks.proto#L9-L11) in the protobuf for requesting detailed data on an individual task. maybe something like this:
```protobuf
service Tasks {
    rpc WatchTasks(TasksRequest) returns (stream TaskUpdate) {}
    rpc TaskDetails(DetailsRequest) returns (stream TaskDetails) {}
}

// ...

message DetailsRequest {
    Task id = 1;
}

message TaskDetails {
    // HdrHistogram.rs `Histogram` serialized to binary in the V2 + DEFLATE format
    bytes poll_times_histogram = 1;
}

// ...
```
this way, we could always calculate histograms for each task in the aggregator (so that we can have historical data) but only serialize them on request. in the future, we may also want to add additional data to the task details view that we might not want to send for every task, so i think it'll probably be useful to have the additional RPC for other purposes as well.

if we want to include some data from the histogram for all tasks in the `WatchTasks` RPC, we could add a couple percentile values to the `TasksUpdate` message (maybe the p50, p90, and p99 poll times?). those would be useful to display in overview views like the task list, and then when we want to show details about a specific task, we could start requesting its complete histogram.

what do you think? this would be a larger change, but it might be more efficient in terms of data on the wire. we could probably avoid paying the performance penalty from compressing the histograms in that case, because we wouldn't be sending the complete histogram for most tasks.

hawkw (commented):
thanks for working on this, this is really cool!

i had some thoughts on how we could potentially save some data on the wire by only sending the complete histogram on-demand. although `hdrhistogram` is pretty space-efficient, sending one for each task is still potentially a lot of data in a production system, where there might be a very large number of tasks.

let me know what you think of my suggestions?

console-subscriber/src/aggregator.rs
oguzbilgener: I just went with the smallest size option initially. Thank you for pointing this out, the serializer format was definitely something I wanted to discuss. After re-reading the paragraph you linked I definitely agree. Also compression is something that can be applied at a lower layer.

So I'm going to change this to V2Serializer.

proto/tasks.proto
oguzbilgener: This is a great idea, I was thinking that it could be made more efficient when playing with it on the TaskView side and trying to determine when to deserialize.  

I'm okay with it being a larger change. Though I jinxed myself when I said in the issue that I had lots of free time, I should be able to tackle this this week :)

hawkw: okay, that sounds great, i think this would be the right approach, especially when we add other detailed stats. let me know if you have any questions?

seanmonstar:
Thanks for the screenshot, that looks super cool! What exactly is the graph showing?

oguzbilgener:
It should show... the entire histogram :sweat_smile: with the bucket count equal to the number of horizontal pixels in the render area.

It looks like tui's `Sparkline` omits relatively very small values, so that's why both ends of the graph are empty, otherwise the graph should fill the box horizontally. The sparkline should give a general idea of how the poll times are distributed.

I also played with `BarChart` but it requires more space and its value labels were sort of ugly when some buckets have so small values that the value label is taller than the bar:



Also, there may be [better ways](http://hdrhistogram.github.io/HdrHistogram/plotFiles.html) to visualize this since it's a latency histogram. So I'm okay with taking it out.

console/src/view/task.rs
```diff
@@ -123,13 +160,95 @@ impl TaskView {
         let mut fields = Text::default();
         fields.extend(task.formatted_fields().iter().cloned().map(Spans::from));
 
+        let chart_data = make_chart_data(self.details.clone(), task.id(), sparkline_area.width - 2);
+        let histogram_sparkline = Sparkline::default()
+            .block(Block::default().title("Poll Times").borders(Borders::ALL))
+            .data(&chart_data);
```
hawkw: It would be really nice to get some kind of label for bucket values on the histogram...we don't have to label each bucket (which, if the bars are single character wide, we don't really have space for), but it would be nice to label at least the minimum and maximum buckets with their durations (and maybe also the bucket with the highest value?). Without labels, the histogram is kinda hard to interpret...

I don't think `tui`'s `Sparkline` has a built-in way to add axis labels to histograms, but maybe we could do it manually by adding a single line of text underneath the histogram?

Ideally, in a GUI, we might draw one widget for the sparkline *and* percentiles, and label the histogram's x-axis with the percentiles, but because each bucket's bar is one character wide, I don't think that would work in practice.

console/src/view/task.rs
```diff
@@ -123,13 +160,95 @@ impl TaskView {
...
+        })
+        .unwrap_or_default()
+}
+
+fn make_percentiles_widgets(details: DetailsRef, task_id: u64) -> (Text<'static>, Text<'static>) {
+    let percentiles_iter = details
+        .borrow()
+        .as_ref()
+        .and_then(|details| filter_same_task(task_id, details))
+        .and_then(|details| details.poll_times_histogram())
```
hawkw: nit, take it or leave it: this could be combined into one `and_then`
```suggestion
        .and_then(|details| Some(filter_same_task(task_id, details)?.poll_times_histogram()))
```
but, what you have now might be clearer, so it would be fine to leave it as-is...

console/src/view/task.rs
```diff
@@ -123,13 +160,95 @@ impl TaskView {
...
+}
+
+fn make_chart_data(details: DetailsRef, task_id: u64, width: u16) -> Vec<u64> {
+    details
+        .borrow()
+        .as_ref()
+        .and_then(|details| filter_same_task(task_id, details))
+        .and_then(|details| details.poll_times_histogram())
+        .map(|histogram| {
+            // This is probably very buggy
```
hawkw: what's potentially buggy here? it would be nice if the comment explained the potential issue...

console/src/view/task.rs
```diff
@@ -123,13 +160,95 @@ impl TaskView {
...
+            let steps = ((histogram.max() - histogram.min()) as f64 / width as f64).ceil() as u64;
+            if steps > 0 {
+                let data: Vec<u64> = histogram
+                    .iter_linear(steps)
+                    .map(|it| it.count_since_last_iteration())
+                    .collect();
+                data
+            } else {
+                Vec::new()
+            }
```
hawkw: i'm not sure if the `if` is _strictly_ necessary here; I imagine `iter_linear` returns an empty iterator when `steps` is 0, and `collect`ing an empty iterator will just give us an empty `Vec`. this might be _slightly_ more efficient, since the iterator is never created, but i dunno if this has a meaningful performance impact. 

i'm fine with this though.

console-subscriber/src/aggregator.rs
```diff
@@ -131,26 +162,55 @@ impl Aggregator {
...
+                                        self
+                                            .details_watchers
+                                            .entry(task_id)
+                                            .or_insert_with(Vec::new)
+                                            .push(subscription);
+                                    }
+
+                                }
+                                // If task is not found, drop the subscription which should close the stream
+                            },
```
hawkw: take it or leave it: this is now enough code that it _might_ be worth trying to factor it out into a method, or a pair of methods for starting `Tasks` and `TaskDetails` watches? not a blocker though.

console-subscriber/src/lib.rs
```diff
@@ -50,7 +51,12 @@ struct WakerVisitor {
     op: Option<WakeOp>,
 }
 
-struct Watch(mpsc::Sender<Result<proto::tasks::TaskUpdate, tonic::Status>>);
+struct Watch<T>(mpsc::Sender<Result<T, tonic::Status>>);
+
+enum WatchKind {
+    TaskUpdate(Watch<proto::tasks::TaskUpdate>),
+    TaskDetailUpdate(SpanId, Watch<proto::tasks::TaskDetails>),
```
hawkw: nit, not a blocker: i would personally have called these 
```suggestion
    Tasks(Watch<proto::tasks::TaskUpdate>),
    TaskDetails(SpanId, Watch<proto::tasks::TaskDetails>),
```
based on the name of the RPCs, rather than the names of the update message.

console-subscriber/src/lib.rs
```diff
@@ -343,11 +351,32 @@ impl proto::tasks::tasks_server::Tasks for Server {
...
+            .ok_or_else(|| tonic::Status::invalid_argument("missing task_id"))?;
+        let permit = self.subscribe.reserve().await.map_err(|_| {
+            tonic::Status::internal("cannot start new watch, aggregation task is not running")
+        })?;
+        let (tx, rx) = mpsc::channel(self.client_buffer);
+        permit.send(WatchKind::TaskDetailUpdate(task_id, Watch(tx)));
+        tracing::debug!("task details watch started");
+        // At this point we don't know if the task with the given ID exists or not.
+        // But the only Watch sender will be dropped if the task doesn't exist
+        // so that should close the stream.
```
hawkw: hmm...i think it might technically be more correct to return the gRPC `NOT_FOUND` status code in the case where the requested task doesn't exist...but that would require more refactoring.

if we wanted to do that, we could instead send the aggregator task a message containing a task ID and a `tokio::sync::oneshot::Sender<Watch<proto::tasks::TaskDetails>>`. if a task for that ID exists, the aggregator task could send back a new watch, otherwise, it would drop the `oneshot::Sender`. then, we could await the `oneshot::Receiver` here. if it returns a watch stream, we return that to the client, otherwise, if the oneshot is closed, we fail the RPC with the `NOT_FOUND` status code.

this is, of course, a bit more complex, but it's probably more semantically correct. it may not matter in practice, though...

jonhoo:
Sideline comment — definitely provide an option to render a CDF if you can. It's a little harder to read than a histogram, but provides much more data (e.g., what is the 95th percentile latency) at a glance.

console/src/view/task.rs
oguzbilgener: So, I've been working on this today: (commit https://github.com/tokio-rs/console/pull/47/commits/924e1757f61a2e821b2126b3368b750c0f912c18)
Screenshot from 2021-06-26 18-20-28

(also modified sparkline to always render something if count > 0.)

However, after looking at the raw output of `iter_linear` and wondering why the bucket count is not always what I intended and why there are sometimes 0s at both ends, I realized that what I'm doing is probably wrong and I should probably give up on this traditional histogram visualization idea. `HdrHistogram` doesn't seem to be designed for this, none of the iterators it offers allows getting counts for a deterministic number of buckets, like the chart area width. 

Does this make sense? What do you think?

console/src/view/task.rs
oguzbilgener: I tried this, but `histogram` in the next line in `.map(..)` becomes an `Option<Histogram>` so I'd need to make more changes.

console/src/view/task.rs
oguzbilgener: I think I intuitively added the `if` here but turns out `iter_linear` [panics](https://docs.rs/hdrhistogram/7.3.0/src/hdrhistogram/iterators/linear.rs.html#21-24) if the given `steps` is less than or equal to 0

console-subscriber/src/lib.rs
oguzbilgener: Pushed https://github.com/tokio-rs/console/pull/47/commits/fc982d25e2fa25937ca79c83f368a5909bd85e24 to solve this. I believe this is slightly different from what you suggested, so please let me know if you like this. I wanted to avoid sending a separate message before the actual `permit.send(WatchKind)` call, in order to avoid performing error handling twice.

console-subscriber/src/lib.rs
oguzbilgener: Done in https://github.com/tokio-rs/console/pull/47/commits/d4d643f197df4c1c082dbb7665eeea5b516ac35a

console-subscriber/src/aggregator.rs
oguzbilgener: Done in https://github.com/tokio-rs/console/pull/47/commits/12c19b92d79211417247c5ac1ea4494b35f17704

console/src/view/task.rs
hawkw: > Does this make sense? What do you think?

hmm, i'd ping @jonhoo about this, i'm not personally all that familiar with `hdrhistogram`, while he's the crate author. 

the current implementation *looks* reasonable to me, but we should make sure it's a correct representation of the data.

console/src/view/task.rs
hawkw: oh i see, nvm!

console/src/view/task.rs
hawkw: oh, huh, that's fine, then. maybe worth adding a comment noting that we need the `if`, in that case?

console-subscriber/src/aggregator.rs
```diff
@@ -131,26 +164,14 @@ impl Aggregator {
...
+                            WatchKind::Task(subscription) => {
+                                self.add_task_subscription(subscription);
+                            },
+                            WatchKind::TaskDetail(watch_request) => {
+                                self.add_task_detail_subscription(watch_request);
+                            },
+                        };
                     } else {
                         tracing::debug!("rpc channel closed, terminating");
                         return;
```
hawkw: style nit, take it or leave it: we could combine the `match` and `if let` here, to something like:
```rust
match subscription {
    Some(WatchKind::Task(subscription)) = {
         self.add_task_subscription(subscription);
    },
    Some(WatchKind::TaskDetail(watch_request)) => {
        self.add_task_detail_subscription(watch_request);
    },
    None => {
        tracing::debug!("rpc channel closed, terminating");
        return;
    },
}
```

not a big deal either way, but this might be a bit neater.

console-subscriber/src/aggregator.rs
```diff
@@ -427,6 +539,10 @@ impl<T> TaskData<T> {
     fn all(&self) -> impl Iterator<Item = (&span::Id, &T)> {
         self.data.iter().map(|(id, (data, _))| (id, data))
     }
+
+    fn find(&self, id: &span::Id) -> Option<&T> {
```
hawkw: nit, take it or leave it: might have called this `get` to keep it consistent with the `HashMap` method? not a blocker

console-subscriber/src/lib.rs
```diff
@@ -50,7 +51,18 @@ struct WakerVisitor {
     op: Option<WakeOp>,
 }
 
-struct Watch(mpsc::Sender<Result<proto::tasks::TaskUpdate, tonic::Status>>);
+struct Watch<T>(mpsc::Sender<Result<T, tonic::Status>>);
+
+enum WatchKind {
+    Task(Watch<proto::tasks::TaskUpdate>),
```
hawkw: nit: i think this should be called "task*s*", rather than "task", "task" suggests that it is a watch on a  _single_ task

console-subscriber/src/lib.rs
```diff
@@ -343,11 +357,40 @@ impl proto::tasks::tasks_server::Tasks for Server {
...
+        let (stream_sender, stream_recv) = oneshot::channel();
+        permit.send(WatchKind::TaskDetail(WatchRequest {
+            id: task_id.clone(),
+            stream_sender,
+            buffer: self.client_buffer,
+        }));
+        // If the aggregator drops the sender, the task doesn't exist.
+        let rx = stream_recv
+            .await
+            .map_err(|_| tonic::Status::not_found("task not found"))?;
```
hawkw: :+1: 

(but, i might log a warning in this case, maybe in the `map_err` block?)

console/src/view/mini_histogram.rs
```diff
@@ -0,0 +1,242 @@
...
+        };
+        self.render_bars(bars_area, buf);
+    }
+}
+
+impl<'a> MiniHistogram<'a> {
+    // fn bars_area_width() -> u64 {
+    //     max_qty_label.len() as u16;
+    // }
+
```
oguzbilgener:
```suggestion
```

proto/tasks.proto
```diff
@@ -44,6 +49,16 @@ message TaskUpdate {
...
 }
 
+// A task details update
+message TaskDetails {
+    // The task's ID which the details belong to.
+    common.SpanId task_id = 1;
+
+    google.protobuf.Timestamp now = 2;
+
+    Details details = 3;
```
hawkw: nit, take it or leave it: i'm not convinced this needs to be a separate message --- i think it would be fine just to put the histogram (and any other data we might add) in the `TaskDetails` message directly. should make the generated code a bit easier to work with that way...

console/src/view/task.rs
```diff
@@ -123,13 +163,134 @@ impl TaskView {
...
+                (75, histogram.value_at_percentile(75.0)),
+                (90, histogram.value_at_percentile(90.0)),
+                (95, histogram.value_at_percentile(95.0)),
+                (99, histogram.value_at_percentile(99.0)),
+            ]
+        })
+        .map(|pairs| {
+            pairs.into_iter().map(|pair| {
+                format!(
+                    "p{}: {:.prec$?}",
```
hawkw: nit: it would be nice to bold the percentile label ("pNN") like we do for labels elsewhere?

console/src/view/task.rs
```diff
@@ -123,13 +163,134 @@ impl TaskView {
...
         frame.render_widget(wakers_widget, stats_area[1]);
         frame.render_widget(fields_widget, fields_area);
+        frame.render_widget(percentiles_widget, histogram_area[0]);
+        frame.render_widget(percentiles_widget_1, percentiles_columns[0]);
+        frame.render_widget(percentiles_widget_2, percentiles_columns[1]);
+        frame.render_widget(histogram_sparkline, sparkline_area);
+    }
+}
+
+fn make_chart_data(details: DetailsRef, task_id: u64, width: u16) -> (Vec<u64>, HistogramMetadata) {
```
hawkw: seems like this (and `make_percentiles_widgets`) _could_  be methods on `DetailsRef`...not a blocker though

console/src/view/task.rs
oguzbilgener: Updated:
Screen Shot 2021-06-29 at 2 52 32 PM

To add more percentiles with different lengths like p5 or p99.9 I need to account for the spacing better, but not sure if we care about those at the moment? It would be nice to fill up those two columns of percentiles though 😄

console/src/view/task.rs
oguzbilgener: Unfortunately `DetailsRef` is a type alias and the `Details` struct definition lives in `console/src/tasks.rs`. Not sure if we want to move this implementation there? Having an `impl Details {}` here might also be confusing as well.

console/src/main.rs
```diff
@@ -82,3 +108,56 @@ async fn main() -> color_eyre::Result<()> {
...
+                            _ => {}
+                        }
+                    } else {
+                        break;
+                    }
+                },
+            }
+        }
+    } else {
+        // TODO: handle connection error, print details somewhere? Related to Issue #30
```
oguzbilgener, line 161: I can make another small PR to address this after #57

console/src/view/task.rs
hawkw: i think it's fine to have an `impl Details` here.

oguzbilgener: Done in https://github.com/tokio-rs/console/pull/47/commits/992f01c561148b35612eaefe05bb9dfaa072f957. Sorry about the back and forth

console/src/tasks.rs
```diff
@@ -184,6 +195,30 @@ impl State {
         }
     }
 
+    pub(crate) fn get_task_details_ref(&self) -> DetailsRef {
```
hawkw: i would probably have just called this
```suggestion
    pub(crate) fn task_details_ref(&self) -> DetailsRef {
```
or even
```suggestion
    pub(crate) fn details_ref(&self) -> DetailsRef {
```
in general, it's [considered more idiomatic](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter) to only use the `get_` prefix for functions that do some kind of data structure lookup, and avoid it for field accessors like this method.

console/src/view/task.rs
```diff
@@ -123,13 +167,128 @@ impl TaskView {
...
+        let percentiles_iter = self
+            .poll_times_histogram()
+            .map(|histogram| {
+                [10f64, 25f64, 50f64, 75f64, 90f64, 95f64, 99f64]
+                    .iter()
+                    .map(|i| (*i, histogram.value_at_percentile(*i)))
+                    .collect::<Vec<(f64, u64)>>()
+            })
+            .map(|pairs| {
+                pairs.into_iter().map(|pair| {
```
hawkw: nit: i *think* it should be possible to avoid the `collect`/`into_iter` here, and just have the first `map` closure return an iterator? that would be a bit more efficient, as we would be able to avoid allocating an intermediate `Vec`
```suggestion
            })
            .map(|pairs| {
                pairs.map(|pair| {
```

console/src/view/task.rs
```diff
@@ -123,13 +167,128 @@ impl TaskView {
...
+                            // Because HdrHistogram can return empty buckets depending
+                            // on its internal state, as it approximates values.
+                            if *count == 0 && !found_first_nonzero {
+                                false
+                            } else {
+                                found_first_nonzero = true;
+                                true
+                            }
+                        })
+                        .collect();
```
hawkw: nit/take it or leave it: could the `map` and `filter` be combined with `filter_map`? not a big deal either way.

hawkw (approved):
this looks good to me! would be nice to get a :+1: from @seanmonstar as well, if you have a minute to look it over?

console/src/tasks.rs
oguzbilgener: This round of suggestions is also taken care of. Thank you for the reviews!

console/src/view/task.rs
hawkw: alternatively, we could just make the poll times section a couple lines longer, and make one column of percentiles. then, the percentiles box wouldn't have to be 50% of the section, and we could make the histogram wider...and get more vertical space, which would (hopefully) allow us to draw the histogram with a little more vertical resolution.

but, we can tweak the layout and stuff in subsequent PRs. the important part for this branch, IMO, is getting the histogram wired through from the subscriber to the console, and having a basic setup for drawing it. which we've done! so, let's just merge this and keep working on making it look nice as a follow-up. :)

seanmonstar (approved):
Let's goooooo 🚀

hawkw:
yay! thanks for all your hard work on this, @oguzbilgener!

PR #53, opened 2021-06-18
add an environment-configured console_subscriber::init

### LINKED ISSUE #49: subscriber: provide a simpler 'init' function
The current way to setup the subscriber in an app is flexible, but complex. It'd be useful to have a simple `console_subscriber::init()` function that does the most common thing for you. Notably:

- Builds the `TaskLayer` with defaults.
- Inits a tracing registry and adds the layer.
- Starts a new thread to run a single-threaded Tokio runtime.
- Spawns the `server` portion of the task layer into that runtime.

jbr:
closes #49

### DIFF
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -113,7 +113,6 @@ impl Builder {
     ///   localhost:1234 or similar. Default: 127.0.0.1:6669
     /// * `TOKIO_CONSOLE_PUBLISH_INTERVAL`: The number of milliseconds to wait
     ///   between sending updates to the console. Default: 1000ms (1s)
-
     pub fn from_default_env(mut self) -> Self {
         if let Ok(retention) = std::env::var("TOKIO_CONSOLE_RETENTION") {
             self.retention = Duration::from_secs(
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -112,7 +112,7 @@ impl Builder {
     /// | `TOKIO_CONSOLE_RETENTION_SECS`      | The number of seconds to accumulate completed tracing data                | 3600s (1h)        |
     /// | `TOKIO_CONSOLE_BIND`                | a HOST:PORT description, such as `localhost:1234`                         | `127.0.0.1:6669`  |
     /// | `TOKIO_CONSOLE_PUBLISH_INTERVAL_MS` | The number of milliseconds to wait between sending updates to the console | 1000ms (1s)       |
-    pub fn from_default_env(mut self) -> Self {
+    pub fn with_default_env(mut self) -> Self {
         if let Ok(retention) = std::env::var("TOKIO_CONSOLE_RETENTION_SECS") {
             self.retention = Duration::from_secs(
                 retention
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -108,7 +108,7 @@ impl Builder {
     /// Configures this builder from a standard set of environment variables:
     ///
     /// * `TOKIO_CONSOLE_RETENTION`: The number of seconds to accumulate
-    ///   completed tracing data. Default: 60s.
+    ///   completed tracing data. Default: 3600s (1h).
     /// * `TOKIO_CONSOLE_BIND`: a HOST:PORT description, such as
     ///   localhost:1234 or similar. Default: 127.0.0.1:6669
     /// * `TOKIO_CONSOLE_PUBLISH_INTERVAL`: The number of milliseconds to wait
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -107,11 +107,11 @@ impl Builder {
 
     /// Configures this builder from a standard set of environment variables:
     ///
-    /// * `TOKIO_CONSOLE_RETENTION`: The number of seconds to accumulate
+    /// * `TOKIO_CONSOLE_RETENTION_SECS`: The number of seconds to accumulate
     ///   completed tracing data. Default: 3600s (1h).
     /// * `TOKIO_CONSOLE_BIND`: a HOST:PORT description, such as
     ///   localhost:1234 or similar. Default: 127.0.0.1:6669
-    /// * `TOKIO_CONSOLE_PUBLISH_INTERVAL`: The number of milliseconds to wait
+    /// * `TOKIO_CONSOLE_PUBLISH_INTERVAL_MS`: The number of milliseconds to wait
     ///   between sending updates to the console. Default: 1000ms (1s)
     pub fn from_default_env(mut self) -> Self {
         if let Ok(retention) = std::env::var("TOKIO_CONSOLE_RETENTION_SECS") {
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -114,11 +114,11 @@ impl Builder {
     /// * `TOKIO_CONSOLE_PUBLISH_INTERVAL`: The number of milliseconds to wait
     ///   between sending updates to the console. Default: 1000ms (1s)
     pub fn from_default_env(mut self) -> Self {
-        if let Ok(retention) = std::env::var("TOKIO_CONSOLE_RETENTION") {
+        if let Ok(retention) = std::env::var("TOKIO_CONSOLE_RETENTION_SECS") {
             self.retention = Duration::from_secs(
                 retention
                     .parse()
-                    .expect("TOKIO_CONSOLE_RETENTION must be an integer"),
+                    .expect("TOKIO_CONSOLE_RETENTION_SECS must be an integer"),
             );
         }
 
@@ -130,11 +130,11 @@ impl Builder {
                 .expect("tokio console could not resolve TOKIO_CONSOLE_BIND");
         }
 
-        if let Ok(interval) = std::env::var("TOKIO_CONSOLE_PUBLISH_INTERVAL") {
+        if let Ok(interval) = std::env::var("TOKIO_CONSOLE_PUBLISH_INTERVAL_MS") {
             self.publish_interval = Duration::from_millis(
                 interval
                     .parse()
-                    .expect("TOKIO_CONSOLE_PUBLISH_INTERVAL must be an integer"),
+                    .expect("TOKIO_CONSOLE_PUBLISH_INTERVAL_MS must be an integer"),
             );
         }
 
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -110,7 +110,7 @@ impl Builder {
     /// | **Environment Variable**            | **Purpose**                                                               | **Default Value** |
     /// |-------------------------------------|---------------------------------------------------------------------------|-------------------|
     /// | `TOKIO_CONSOLE_RETENTION_SECS`      | The number of seconds to accumulate completed tracing data                | 3600s (1h)        |
-    /// | `TOKIO_CONSOLE_BIND`                | a HOST:PORT description, such as localhost:1234                           | 127.0.0.1:6669    |
+    /// | `TOKIO_CONSOLE_BIND`                | a HOST:PORT description, such as `localhost:1234`                         | `127.0.0.1:6669`  |
     /// | `TOKIO_CONSOLE_PUBLISH_INTERVAL_MS` | The number of milliseconds to wait between sending updates to the console | 1000ms (1s)       |
     pub fn from_default_env(mut self) -> Self {
         if let Ok(retention) = std::env::var("TOKIO_CONSOLE_RETENTION_SECS") {
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -107,12 +107,11 @@ impl Builder {
 
     /// Configures this builder from a standard set of environment variables:
     ///
-    /// * `TOKIO_CONSOLE_RETENTION_SECS`: The number of seconds to accumulate
-    ///   completed tracing data. Default: 3600s (1h).
-    /// * `TOKIO_CONSOLE_BIND`: a HOST:PORT description, such as
-    ///   localhost:1234 or similar. Default: 127.0.0.1:6669
-    /// * `TOKIO_CONSOLE_PUBLISH_INTERVAL_MS`: The number of milliseconds to wait
-    ///   between sending updates to the console. Default: 1000ms (1s)
+    /// | **Environment Variable**            | **Purpose**                                                               | **Default Value** |
+    /// |-------------------------------------|---------------------------------------------------------------------------|-------------------|
+    /// | `TOKIO_CONSOLE_RETENTION_SECS`      | The number of seconds to accumulate completed tracing data                | 3600s (1h)        |
+    /// | `TOKIO_CONSOLE_BIND`                | a HOST:PORT description, such as localhost:1234                           | 127.0.0.1:6669    |
+    /// | `TOKIO_CONSOLE_PUBLISH_INTERVAL_MS` | The number of milliseconds to wait between sending updates to the console | 1000ms (1s)       |
     pub fn from_default_env(mut self) -> Self {
         if let Ok(retention) = std::env::var("TOKIO_CONSOLE_RETENTION_SECS") {
             self.retention = Duration::from_secs(
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -1,5 +1,8 @@
 use super::{Server, TasksLayer};
-use std::{net::SocketAddr, time::Duration};
+use std::{
+    net::{SocketAddr, ToSocketAddrs},
+    time::Duration,
+};
 
 /// Builder for configuring [`TasksLayer`]s.
 #[derive(Clone, Debug)]
@@ -100,5 +103,42 @@ impl Builder {
     /// Completes the builder, returning a [`TasksLayer`] and [`Server`] task.
     pub fn build(self) -> (TasksLayer, Server) {
         TasksLayer::build(self)
+    }
+
+    /// Configures this builder from a standard set of environment variables:
+    ///
+    /// * `TOKIO_CONSOLE_RETENTION`: The number of seconds to accumulate
+    ///   completed tracing data. Default: 60s.
+    /// * `TOKIO_CONSOLE_BIND`: a HOST:PORT description, such as
+    ///   localhost:1234 or similar. Default: 127.0.0.1:6669
+    /// * `TOKIO_CONSOLE_PUBLISH_INTERVAL`: The number of milliseconds to wait
+    ///   between sending updates to the console. Default: 1000ms (1s)
+
+    pub fn from_default_env(mut self) -> Self {
+        if let Ok(retention) = std::env::var("TOKIO_CONSOLE_RETENTION") {
+            self.retention = Duration::from_secs(
+                retention
+                    .parse()
+                    .expect("TOKIO_CONSOLE_RETENTION must be an integer"),
+            );
+        }
+
+        if let Ok(bind) = std::env::var("TOKIO_CONSOLE_BIND") {
+            self.server_addr = bind
+                .to_socket_addrs()
+                .expect("TOKIO_CONSOLE_BIND must be formatted as HOST:PORT, such as localhost:4321")
+                .next()
+                .expect("tokio console could not resolve TOKIO_CONSOLE_BIND");
+        }
+
+        if let Ok(interval) = std::env::var("TOKIO_CONSOLE_PUBLISH_INTERVAL") {
+            self.publish_interval = Duration::from_millis(
+                interval
+                    .parse()
+                    .expect("TOKIO_CONSOLE_PUBLISH_INTERVAL must be an integer"),
+            );
+        }
+
+        self
     }
 }
--- a/console-subscriber/src/init.rs
+++ b/console-subscriber/src/init.rs
@@ -1,0 +1,83 @@
+use crate::TasksLayer;
+use std::thread;
+use tokio::runtime;
+use tracing_core::{subscriber::Subscriber, Dispatch};
+use tracing_subscriber::{fmt, layer::Layered, prelude::*, EnvFilter, Layer, Registry};
+
+type ConsoleSubscriberLayer =
+    Layered<TasksLayer, Layered<EnvFilter, Layered<fmt::Layer<Registry>, Registry>>>;
+
+/// Starts the console subscriber server on its own thread
+///
+/// This function represents the easiest way to get started using
+/// tokio-console.
+///
+/// ## Configuration
+///
+/// Tokio console subscriber is configured with sensible defaults for most
+/// use cases. If you need to tune these parameters, several environmental
+/// configuration variables are available:
+///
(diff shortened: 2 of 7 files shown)


console-subscriber/src/builder.rs
```diff
@@ -101,4 +104,43 @@ impl Builder {
...
+    /**
+    Configures this builder from a standard set of environment variables:
+
+    * `TOKIO_CONSOLE_RETENTION`: The number of seconds to accumulate
+      completed tracing data. Default: 60s.
+    * `TOKIO_CONSOLE_BIND`: a HOST:PORT description, such as
+      localhost:1234 or similar. Default: 127.0.0.1:6669
+    * `TOKIO_CONSOLE_PUBLISH_INTERVAL`: The number of milliseconds to wait
+      between sending updates to the console. Default: 1000ms (1s)
+    */
```
hawkw: style nit: elsewhere we use `///`-style doc comments exclusively...i could be convinced to switch to using block comments for doc comments, but i'd prefer to be consistent about it.

console-subscriber/src/builder.rs
```diff
@@ -101,4 +104,41 @@ impl Builder {
...
+    /// Configures this builder from a standard set of environment variables:
+    ///
+    /// * `TOKIO_CONSOLE_RETENTION`: The number of seconds to accumulate
+    ///   completed tracing data. Default: 60s.
+    /// * `TOKIO_CONSOLE_BIND`: a HOST:PORT description, such as
+    ///   localhost:1234 or similar. Default: 127.0.0.1:6669
+    /// * `TOKIO_CONSOLE_PUBLISH_INTERVAL`: The number of milliseconds to wait
+    ///   between sending updates to the console. Default: 1000ms (1s)
+
+    pub fn from_default_env(mut self) -> Self {
```
hawkw: nit: in general, we don't put whitespace between doc comments and functions elsewhere in the project
```suggestion
    ///   between sending updates to the console. Default: 1000ms (1s)
    pub fn from_default_env(mut self) -> Self {
```

console-subscriber/src/builder.rs
```diff
@@ -101,4 +104,41 @@ impl Builder {
...
+        if let Ok(retention) = std::env::var("TOKIO_CONSOLE_RETENTION") {
+            self.retention = Duration::from_secs(
+                retention
+                    .parse()
+                    .expect("TOKIO_CONSOLE_RETENTION must be an integer"),
+            );
+        }
```
hawkw: since we have two different duration values that are configured from an env var, and they're in different units, i think we ought to make the env var names include the unit. it's good that we list the units in the docs, but i think it's probably good to also ensure that it's clear when you see the variables _used_ --- you might be, for example,  reading a shell script that says something like

```bash
TOKIO_CONSOLE_RETENTION=60 TOKIO_CONSOLE_PUBLISH_INTERVAL=1000 ./myapp
```
and get the wrong impression that the retention period is significantly shorter than the publish interval.

so, i'd suggest:
```suggestion
        if let Ok(retention) = std::env::var("TOKIO_CONSOLE_RETENTION_SECS") {
            self.retention = Duration::from_secs(
                retention
                    .parse()
                    .expect("TOKIO_CONSOLE_RETENTION_SECS must be an integer"),
            );
        }
```

console-subscriber/src/builder.rs
```diff
@@ -101,4 +104,41 @@ impl Builder {
...
+                .expect("tokio console could not resolve TOKIO_CONSOLE_BIND");
+        }
+
+        if let Ok(interval) = std::env::var("TOKIO_CONSOLE_PUBLISH_INTERVAL") {
+            self.publish_interval = Duration::from_millis(
+                interval
+                    .parse()
+                    .expect("TOKIO_CONSOLE_PUBLISH_INTERVAL must be an integer"),
+            );
+        }
```
hawkw: ...and similarly, i'd change this to
```suggestion
        if let Ok(interval) = std::env::var("TOKIO_CONSOLE_PUBLISH_INTERVAL_MS") {
            self.publish_interval = Duration::from_millis(
                interval
                    .parse()
                    .expect("TOKIO_CONSOLE_PUBLISH_INTERVAL_MS must be an integer"),
            );
        }
```

console-subscriber/src/builder.rs
```diff
@@ -101,4 +104,41 @@ impl Builder {
...
     }
+
+    /// Configures this builder from a standard set of environment variables:
+    ///
+    /// * `TOKIO_CONSOLE_RETENTION`: The number of seconds to accumulate
+    ///   completed tracing data. Default: 60s.
+    /// * `TOKIO_CONSOLE_BIND`: a HOST:PORT description, such as
+    ///   localhost:1234 or similar. Default: 127.0.0.1:6669
+    /// * `TOKIO_CONSOLE_PUBLISH_INTERVAL`: The number of milliseconds to wait
+    ///   between sending updates to the console. Default: 1000ms (1s)
```
hawkw: take it or leave it: it might be nice if this was a markdown table like
```markdown
+---------------------------------+-----------------+-----------------------+
| **Environment Variable** | **Purpose** | **Default Value** |
+---------------------------------+-----------------+-----------------------+
...
```
rather than a bulleted list? but, not a big deal.

console-subscriber/src/lib.rs
```diff
@@ -16,13 +16,69 @@ use tracing_core::{
...
+  completed tracing data. Default: 60s.
+* `TOKIO_CONSOLE_BIND`: a HOST:PORT description, such as
+  localhost:1234 or similar. Default: 127.0.0.1:6669
+* `TOKIO_CONSOLE_PUBLISH_INTERVAL`: The number of milliseconds to wait
+  between sending updates to the console. Default: 1000ms (1s)
+* `RUST_LOG`: configure the tracing filter. Default: `tokio=trace`,
+  and any additional filtering directives will be appended to this
+  default. See [`EnvFilter`] for further information on the format
+  for this variable.
+*/
```
hawkw: nit: again, we generally prefer line comments for rustdoc elsewhere.

console-subscriber/src/lib.rs
```diff
@@ -16,13 +16,69 @@ use tracing_core::{
...
 use aggregator::Aggregator;
 mod builder;
 pub use builder::Builder;
 
+/**
+Starts the console subscriber server on its own thread
+
+This function represents the easiest way to get started using
+tokio-console.
+
```
hawkw: these docs should probably state explicitly that this function also sets the default `tracing` subscriber (and maybe include a link to the `tracing` docs?). otherwise, i can see people being surprised if they try to set their own subscriber and it fails because the default subscriber is already set.

console-subscriber/src/builder.rs
```diff
@@ -101,4 +104,41 @@ impl Builder {
...
+
```
hawkw: take it or leave it: it would be nice if the docs for this also included an example of using the builder to configure a console layer from the default environment variables , and then using it to build a subscriber, spawn the background task, etc? but, we can come back and add that later.

console-subscriber/src/builder.rs
```diff
@@ -101,4 +104,41 @@ impl Builder {
     pub fn build(self) -> (TasksLayer, Server) {
         TasksLayer::build(self)
     }
+
+    /// Configures this builder from a standard set of environment variables:
+    ///
+    /// * `TOKIO_CONSOLE_RETENTION`: The number of seconds to accumulate
+    ///   completed tracing data. Default: 60s.
```
hawkw: the default retention is 60 _minutes_ (1 hour), not 60 seconds: https://github.com/tokio-rs/console/blob/103021e5dfb7d5a25381b2edbe006e9e85a676a9/console-subscriber/src/lib.rs#L196-L197

we should probably fix the docs:
```suggestion
    ///   completed tracing data. Default: 3600s (1h).
```

console-subscriber/src/builder.rs
hawkw: there are also a set of public associated constants for the default values of these parameters, in case users need to set them programmatically: https://github.com/tokio-rs/console/blob/103021e5dfb7d5a25381b2edbe006e9e85a676a9/console-subscriber/src/builder.rs#L31-L35

it might be worth linking to them here, like the other `Builder` methods do?

console-subscriber/src/lib.rs
```diff
@@ -16,13 +16,69 @@ use tracing_core::{
...
+* `TOKIO_CONSOLE_RETENTION`: The number of seconds to accumulate
+  completed tracing data. Default: 60s.
+* `TOKIO_CONSOLE_BIND`: a HOST:PORT description, such as
+  localhost:1234 or similar. Default: 127.0.0.1:6669
+* `TOKIO_CONSOLE_PUBLISH_INTERVAL`: The number of milliseconds to wait
+  between sending updates to the console. Default: 1000ms (1s)
+* `RUST_LOG`: configure the tracing filter. Default: `tokio=trace`,
+  and any additional filtering directives will be appended to this
+  default. See [`EnvFilter`] for further information on the format
+  for this variable.
```
hawkw: take it or leave it: similarly, this might want to be a table?

console-subscriber/src/lib.rs
```diff
@@ -16,13 +16,69 @@ use tracing_core::{
...
+                .expect("console subscriber runtime initialization failed");
+
+            runtime.block_on(async move {
+                server
+                    .serve()
+                    .await
+                    .expect("console subscriber server failed")
+            });
+        })
+        .expect("console subscriber could not spawn thread");
```
hawkw: we'll hit this `expect` if the background thread panics due to setting the default subscriber failing, but it will have the method `console subscriber could not spawn thread`, which doesn't seem super obvious. it might be worth calling `registry().with(...)....init()` on the calling thread before spawning the background thread, so that the panic doesn't end up wrapped in this panic message?

hawkw (changes requested):
overall, this looks like the right general direction; i had some minor style and documentation suggestions.

console-subscriber/src/builder.rs
jbr: Sweet, this was also my first instinct but wasn't sure if the variable names were getting too long. Self-describing units are great 👍

console-subscriber/src/builder.rs
jbr: Do you think it would make sense to use minutes for the retention period? Is it likely people will need <1min retention?

console-subscriber/src/builder.rs
jbr: Before I saw your review on this PR, I attempted to address this use case by adding a single in-code extension point: `init_with_layer`, with the hope that we can accommodate all common initialization patterns and discourage anyone who would need docs on how to eject from doing so. With the addition of init_with_layer, have we covered all of those use cases? An advantage of this approach is that if the majority of users are going through `init` or `init_with_layer`, future changes to those functions wouldn't require app code change, whereas if we encourage a copy-paste initializer through documentation, we have no way of pushing out changes to that "ejected" code.

console-subscriber/src/builder.rs
jbr: I'm a big fan of `/**` / `/*!` for doc comments because it's less visually distracting, but in this case it was just habit and I should have checked for project convention

console-subscriber/src/init.rs
```diff
@@ -0,0 +1,83 @@
...
+/// ## Adding additional layers
+///
+/// To add an additional [`Layer`], see [`init_with_layer`]
+pub fn init() {
+    init_inner(|console_subscriber| console_subscriber)
+}
+
+/// Starts the console subscriber server with an additional subscriber
+/// layer, for example a log layer. This interface can be combined
+/// with any of the environment configuration documented at [`init`].
```
hawkw: it would be nice to show an example here, but it's okay to add that in a follow-up branch if you prefer?

console-subscriber/src/init.rs
```diff
@@ -0,0 +1,83 @@
...
+    thread::Builder::new()
+        .name("console_subscriber".into())
+        .spawn(move || {
+            let (layer, server) = TasksLayer::builder().from_default_env().build();
+
+            let filter =
+                EnvFilter::from_default_env().add_directive("tokio=trace".parse().unwrap());
+
+            let console_subscriber = tracing_subscriber::registry()
+                .with(fmt::layer())
```
hawkw: do we always want to include a `fmt` layer? this means that users can't customize how events are logged, even if they call `init_with_layer`. i might just only add the `fmt` layer in `init`.

console-subscriber/src/init.rs
```diff
@@ -0,0 +1,83 @@
...
+pub fn init() {
+    init_inner(|console_subscriber| console_subscriber)
+}
+
+/// Starts the console subscriber server with an additional subscriber
+/// layer, for example a log layer. This interface can be combined
+/// with any of the environment configuration documented at [`init`].
+pub fn init_with_layer<AdditionalLayer>(additional_layer: AdditionalLayer)
+where
+    AdditionalLayer: Layer<ConsoleSubscriberLayer> + Send + Sync + 'static,
```
hawkw: an alternative API design is to have a function that spawns the background thread, and just returns a `Registry` with the `EnvFilter` and `TasksLayer` layers already added. that way, the user can call any methods on that `Registry` to add additional layers, and set it as the default subscriber themselves. the `init` function could just call that and then immediately call `init()` on the returned registry. 

i feel like this design could be more intuitive than taking a layer as an argument, but it might be _slightly_ wordier to use...i could be convinced either way.

console-subscriber/src/init.rs
```diff
@@ -0,0 +1,83 @@
...
+use tracing_subscriber::{fmt, layer::Layered, prelude::*, EnvFilter, Layer, Registry};
+
+type ConsoleSubscriberLayer =
+    Layered<TasksLayer, Layered<EnvFilter, Layered<fmt::Layer<Registry>, Registry>>>;
+
+/// Starts the console subscriber server on its own thread
+///
+/// This function represents the easiest way to get started using
+/// tokio-console.
+///
```
hawkw, line 12: can we also document that this calls `tracing::subscriber::set_global_default`? i think i mentioned that earlier but the comment may have disappeared when moving files around.

console-subscriber/src/init.rs
```diff
@@ -0,0 +1,83 @@
...
+/// ## Configuration
+///
+/// Tokio console subscriber is configured with sensible defaults for most
+/// use cases. If you need to tune these parameters, several environmental
+/// configuration variables are available:
+///
+/// * `TOKIO_CONSOLE_RETENTION_SECS`: The number of seconds to accumulate
+///   completed tracing data. Default: 60s.
+/// * `TOKIO_CONSOLE_BIND`: a HOST:PORT description, such as
+///   localhost:1234 or similar. Default: 127.0.0.1:6669
```
hawkw: nit, take it or leave it:
```suggestion
///   `localhost:1234` or similar. Default: `127.0.0.1:6669`
```

console-subscriber/src/init.rs
jbr: That does feel more flexible. What if instead of init_with_layer it was `build() -> impl Layer + Into<Dispatch>`, with `#[must_use]`.

```
console_subscriber::build().with(…).init()
console_subscriber::init(); // shorthand for console_subscriber::build().init();
```

console-subscriber/src/init.rs
jbr: Addressed this in f2b6de51565b504f83b784f6de8f891fab929578 — sorry for moving stuff around during review

console-subscriber/src/init.rs
jbr: addressed in a1ef76b (rebased in concept on the tables)

console-subscriber/src/init.rs
jbr: I'm new to tracing, but layers seem to be order-dependent, and there's no way to prepend because everything is composed through wrapping, so I'm not sure how to extract the fmt layer. Any suggestions?

Do you imagine replacing the fmt layer to be a commonly-needed extension point? I'm trying to figure out how to balance keeping the usage simple with providing the right sort of escape hatches for common customization

console-subscriber/src/lib.rs
jbr: addressed in  f2b6de5

console-subscriber/src/lib.rs
jbr: addressed in f2b6de51565b504f83b784f6de8f891fab929578

console-subscriber/src/builder.rs
jbr: tables addressed in f2b6de51565b504f83b784f6de8f891fab929578

Regarding the constants, it seems like not particularly relevant to the from_default_env interface, as there's no way to programmatically use them here, but they are documented where they can be programmatically used. Does that make sense?

console-subscriber/src/init.rs
jbr: What's a good example layer that people might want to add on here?  My inexperience with tracing means I don't have a good intuition for what people might want

console-subscriber/src/builder.rs
hawkw: yeah, i don't think it's terribly important to have them there, just a thought, not a blocker.

console-subscriber/src/init.rs
hawkw: > I'm new to tracing, but layers seem to be order-dependent, and there's no way to prepend because everything is composed through wrapping, so I'm not sure how to extract the fmt layer. Any suggestions?

layers are not order dependent :)

> Do you imagine replacing the fmt layer to be a commonly-needed extension point? I'm trying to figure out how to balance keeping the usage simple with providing the right sort of escape hatches for common customization

the `fmt` layer is [very configurable](https://docs.rs/tracing-subscriber/0.2.18/tracing_subscriber/fmt/index.html#configuration).  i think wanting to change how events are formatted by providing your own `fmt` layer is probably going to be one of the most common use-cases for overriding the default configuration. for example, if you want to change the log format from textual to JSON, the current API means that you would have to use the builder and do everything manually.

console-subscriber/src/init.rs
```diff
@@ -0,0 +1,81 @@
+use crate::TasksLayer;
+use std::thread;
+use tokio::runtime;
+use tracing_core::{subscriber::Subscriber, Dispatch};
+use tracing_subscriber::{fmt, layer::Layered, prelude::*, EnvFilter, Layer, Registry};
+
+type ConsoleSubscriberLayer =
+    Layered<TasksLayer, Layered<EnvFilter, Layered<fmt::Layer<Registry>, Registry>>>;
+
+/// Starts the console subscriber server on its own thread
```
hawkw: typo: missing period :)
```suggestion
/// Starts the console subscriber server on its own thread,
```

console-subscriber/src/init.rs
```diff
@@ -0,0 +1,81 @@
...
+///
+/// This function represents the easiest way to get started using
+/// tokio-console.
+///
+/// Please note that this registers the exclusive tracing subscriber
+/// for your application. If you need to add additional layers to this
+/// subscriber, see [`init_with_layer`].
```
hawkw: couple minor suggestionshere: 
- i would use the terminology "global default subscriber", as it's still possible for the default subscriber to be temporarily overridden.
- in general, we tend to prefer prefixing notes like this with `**Note**:` in documentation (but it's not a big deal).
- it could also be nice to link to the `tracing` docs here to make things easier for users who aren't familiar with `tracing`.

with all of that said, what about something like this?
```suggestion
/// **Note**: this function sets the [default `tracing` subscriber][default]
/// for your application. If you need to add additional layers to this
/// subscriber, see [`init_with_layer`].
///
/// [default]: https://docs.rs/tracing/0.1.26/tracing/dispatcher/index.html#setting-the-default-subscriber
```

console-subscriber/src/init.rs
jbr: addressed in bb2c2dbef5d03e6d23cca0161b03e1f5471b8119 — there's now `init` and `build` and `build` doesn't add the fmt layer. this does involve a bit of heavy-handed documentation because that might be surprising

console-subscriber/src/init.rs
jbr: addressed in bb2c2dbef5d03e6d23cca0161b03e1f5471b8119 — build and init

console-subscriber/src/init.rs
jbr: addresssed in 572e4547927f77061b6727e89a385f5a2691b341

console-subscriber/src/init.rs
```diff
@@ -0,0 +1,104 @@
+use crate::TasksLayer;
+use std::thread;
+use tokio::runtime;
+use tracing_subscriber::{fmt, layer::Layered, prelude::*, EnvFilter, Registry};
+
+type ConsoleSubscriberLayer = Layered<TasksLayer, Layered<EnvFilter, Registry>>;
```
hawkw, line 6: nit (and this isn't _that important_, since this isn't part of the public API), i would just call this 
```suggestion
type ConsoleSubscriber = Layered<TasksLayer, Layered<EnvFilter, Registry>>;
```
this type _isn't_ actually a `Layer`, since it includes a subscriber --- it doesn't implement the `Layer` trait, it implements the `Subscriber` trait.

console-subscriber/src/init.rs
```diff
@@ -0,0 +1,104 @@
...
+/// ## Examples
+///
+/// ```rust
+/// use tracing_subscriber::prelude::*;
+/// console_subscriber::build()
+///     .with(tracing_subscriber::fmt::layer())
+/// //  .with(...)
+///     .init();
+/// ```
+
```
hawkw, line 81: nit: extra whitespace here
```suggestion
/// ```
```

console-subscriber/src/init.rs
```diff
@@ -0,0 +1,104 @@
...
+#[must_use = "build() without init() will not attach a tracing subscriber"]
```
hawkw: nit: let's use consistent terminology:
```suggestion
#[must_use = "build() without init() will not set the default tracing subscriber"]
```

console-subscriber/src/init.rs
```diff
@@ -0,0 +1,104 @@
...
+pub fn build() -> ConsoleSubscriberLayer {
```
hawkw, line 83: take it or leave it: i'm not *totally* sold on the name `build()` for this --- it doesn't feel super clear to me how this differs from `init()`.. i might have called this `subscriber()` or something, because it returns a `Subscriber`. but, i'm not attached to that, this is fine.

console-subscriber/src/init.rs
```diff
@@ -0,0 +1,104 @@
...
+///     .init();
+/// ```
+
+pub fn init() {
+    build().with(fmt::layer()).init()
+}
+
+/// Builds a new tracing subscriber, but does not yet initialize the
+/// subscriber, allowing for customization by adding additional
+/// layers.
```
hawkw: how about something like:
```suggestion
/// Returns a new `tracing` [subscriber] configured with a [`TasksLayer`]
/// and a [filter] that enables the spans and events required by the console.
///
/// Unlike [`init`], this function does not set the default subscriber, allowing
/// additional [`Layer`s] to be added.
///
/// [subscriber]: https://docs.rs/tracing/latest/tracing/subscriber/trait.Subscriber.html
/// [filter]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html
```

hawkw (approved):

PR #54, opened 2021-06-25
example: add some misbehaving options to example app

seanmonstar:
I've used these in demos to be able to highlight how you can find "bad" tasks in the UI.

- `blocks`: Will show an extremely high busy time compared to number of polls (and with the new poll histogram, even better).
- `coma`: Shows a task that only has 1 poll, and waker stats will show it was never cloned.
- `burn`: While the stats don't yet show it, I plan to include a stat around "self-wakes", which this would highlight.

### DIFF
--- a/console-subscriber/examples/app.rs
+++ b/console-subscriber/examples/app.rs
@@ -5,7 +5,25 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     console_subscriber::init();
 
     let task1 = tokio::spawn(spawn_tasks(1, 10));
-    let task2 = tokio::spawn(spawn_tasks(10, 100));
+    let task2 = tokio::spawn(spawn_tasks(10, 30));
+
+    // spawn optional extras from CLI args
+    // skip first which is command name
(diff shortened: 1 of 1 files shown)


console-subscriber/examples/app.rs
```diff
@@ -5,7 +5,25 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     console_subscriber::init();
 
     let task1 = tokio::spawn(spawn_tasks(1, 10));
-    let task2 = tokio::spawn(spawn_tasks(10, 100));
+    let task2 = tokio::spawn(spawn_tasks(10, 30));
+
+    // spawn optional extras from CLI args
```
hawkw, line 19: it might be kind of nice to have a comment at the top of the file (and/or a `--help` flag?) that mentions the CLI args, and summarizes the kind of bug that each one represents, so that people trying out the console with the example app know what they're doing?

not a blocker though!

hawkw (approved):
this is very nifty, thanks! a brief summary of what the different options do could be nice, but it isn't a huge deal.

PR #55, opened 2021-06-25
Emit and show self-wake count for a Task

seanmonstar:
This counts all wakes of a task that happen while the task is "currently polled". This could be from `task::yield_now()`, from `tokio::coop`, or any other pattern that triggers the task to wake even though it's currently being polled.

### DIFF
--- a/console-subscriber/src/aggregator.rs
+++ b/console-subscriber/src/aggregator.rs
@@ -71,6 +71,7 @@ struct Stats {
     waker_clones: u64,
     waker_drops: u64,
     last_wake: Option<SystemTime>,
+    self_wakes: u64,
 }
 
 #[derive(Default)]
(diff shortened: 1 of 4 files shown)


console-subscriber/src/aggregator.rs
```diff
@@ -295,6 +296,12 @@ impl Aggregator {
                             stats.wakes += 1;
                             stats.last_wake = Some(at);
 
+                            // If currently "inside" this task's poll, then the
+                            // task has woken itself.
+                            if stats.current_polls > 0 {
+                                stats.self_wakes += 1;
+                            }
```
hawkw, line 303: hmm, is this accurate? 

the task might be being polled on one worker thread, and woken from a task on _another_ worker. if we determine a wake to be a "self-wake" any time a task is woken while it is being polled, we will count wakes from another task that occur concurrently with a poll.

i'm not sure how frequently this will cause false positives, so it may not be a huge deal, but i felt like it was worth pointing out.

seanmonstar: Oh yes, true. I suppose the proper way would be to crawl the event `parent`s. Or, or do something with a thread local in the subscriber...

hawkw:
@seanmonstar any plans to move forward with this?

console-subscriber/src/aggregator.rs
hawkw: FWIW, PR #77 will add a thread-local array tracking the current span we're in to the subscriber. We could use this to properly detect self-wakes without false positives, if we're still interested in working on this.

We'd have to do the self-wake detection to the `TasksLayer`, though, rather than in the aggregator (since the layer has the thread-local list of current spans), but we could just add an `is_self_wake` field to the wake event, or something.

Issue #30, opened 2021-05-24
console: handle disconnection more gracefully


hawkw:
right now, if the console is connected to a server, and the server disconnects abruptly, we immediately crash with a big error message:
```
:; cargo run
   Compiling console-subscriber v0.1.0 (/home/eliza/Code/consolation/console-subscriber)
    Finished dev [unoptimized + debuginfo] target(s) in 0.69s
     Running `target/debug/console`
using default address (http://127.0.0.1:6669)
Error:
   0: status: Unknown, message: "h2 protocol error: broken pipe", details: [], metadata: MetadataMap { headers: {} }

Location:
   console/src/main.rs:52

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 5 frames hidden ⋮
   6: console::main::{{closure}}::h0f02eae7993e30df
      at /home/eliza/Code/consolation/console/src/main.rs:52
   7: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll::h89be4d09d64d3ea1
      at /home/eliza/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/future/mod.rs:80
   8: tokio::park::thread::CachedParkThread::block_on::{{closure}}::h8f240001a5901c1b
      at /home/eliza/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.5.0/src/park/thread.rs:263
   9: tokio::coop::with_budget::{{closure}}::h3c95765523460448
      at /home/eliza/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.5.0/src/coop.rs:106
  10: std::thread::local::LocalKey<T>::try_with::hab4856318f2607b4
      at /home/eliza/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/thread/local.rs:272
  11: std::thread::local::LocalKey<T>::with::h3d4d9732c38fcac6
      at /home/eliza/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/thread/local.rs:248
  12: tokio::coop::with_budget::h54da439ec06757dd
[...]

PR #57, opened 2021-06-29
console: handle disconnection more gracefully

zaharidichev:
This PR aims to make it so we do not panic on disconnects.
Instead of panicking the connection state indicator changes
and browsing the last state is still possible.

Screenshot 2021-06-29 at 12 15 14
Screenshot 2021-06-29 at 12 15 20

Related: #30

### DIFF
--- a/console/src/main.rs
+++ b/console/src/main.rs
@@ -20,20 +20,6 @@ enum ConnectionState {
     Disconnected,
 }
 
-async fn connect(
-    target: String,
-    should_delay: bool,
-) -> Result<tonic::Streaming<TaskUpdate>, Box<dyn std::error::Error>> {
-    if should_delay {
-        tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
-    }
-
-    let mut client = TasksClient::connect(target).await?;
-    let request = tonic::Request::new(TasksRequest {});
-    let rsp = client.watch_tasks(request).await?;
-    Ok(rsp.into_inner())
-}
-
 #[tokio::main]
 async fn main() -> color_eyre::Result<()> {
     color_eyre::install()?;
@@ -135,4 +121,18 @@ async fn main() -> color_eyre::Result<()> {
             view.render(f, chunks[1], &mut tasks);
         })?;
     }
+}
+
+async fn connect(
+    target: String,
+    is_reconnect: bool,
+) -> Result<tonic::Streaming<TaskUpdate>, Box<dyn std::error::Error>> {
+    if is_reconnect {
+        tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
+    }
+
+    let mut client = TasksClient::connect(target).await?;
+    let request = tonic::Request::new(TasksRequest {});
+    let rsp = client.watch_tasks(request).await?;
+    Ok(rsp.into_inner())
 }
--- a/console/src/main.rs
+++ b/console/src/main.rs
@@ -105,10 +105,8 @@ async fn main() -> color_eyre::Result<()> {
                 .constraints([Constraint::Length(2), Constraint::Percentage(95)].as_ref())
                 .split(f.size());
 
-            let target = target.clone();
-
             let header_block = Block::default().title(vec![
-                Span::raw(format!("connection: {} ", target.clone())),
+                Span::raw(format!("connection: {} ", target)),
                 match connection_state {
                     Some(ConnectionState::Connected) => Span::styled(
                         "(CONNECTED)",
--- a/console/src/main.rs
+++ b/console/src/main.rs
@@ -1,7 +1,7 @@
 use color_eyre::{eyre::eyre, Help, SectionExt};
-use console_api::tasks::{tasks_client::TasksClient, TasksRequest};
+use console_api::tasks::{tasks_client::TasksClient, TaskUpdate, TasksRequest};
 use futures::stream::StreamExt;
-use tasks::{ConnectionState, State};
+use tasks::State;
 
 use tui::{
     layout::{Constraint, Direction, Layout},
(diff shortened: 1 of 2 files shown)


console/src/main.rs
```diff
@@ -46,10 +48,20 @@ async fn main() -> color_eyre::Result<()> {
...
-                let update = task_update
-                    .ok_or_else(|| eyre!("data stream closed by server"))
-                    .with_section(|| "in the future, this should be reconnected automatically...".header("Note:"))?;
-                tasks.update_tasks(update?);
+                match task_update {
+                    Some(Ok(update)) => {
+                        tasks.update_tasks(update);
+                    },
+                    Some(Err(status)) => {
+                        tracing::error!(%status, "error fro stream");
```
hawkw: typo:
```suggestion
                        tracing::error!(%status, "error from stream");
```

console/src/main.rs
```diff
@@ -60,11 +72,25 @@ async fn main() -> color_eyre::Result<()> {
...
+                        "DISCONNECTED",
+                        Style::default().add_modifier(Modifier::BOLD).fg(Color::Red),
+                    ),
+                    ConnectionState::Pending => Span::styled(
+                        "PENDING",
+                        Style::default()
+                            .add_modifier(Modifier::BOLD)
+                            .fg(Color::Yellow),
+                    ),
+                },
```
hawkw: nit, take it or leave it: it might be worth keeping the remote address in all states, and adding disconnected/pending in addition to the address. for example
```rust
connection: 127.0.0.1:6669 (DISCONNECTED)
```
rather than
```
connection: DISCONNECTED
```

that way, if i have multiple consoles open connected to different instances of a program, i know which is which even if one of them is disconnected.

console/src/tasks.rs
```diff
@@ -11,12 +11,20 @@ use std::{
 };
 use tui::text::Span;
 
+#[derive(Debug)]
+pub(crate) enum ConnectionState {
+    Pending,
```
hawkw: do we ever actually display the pending state? AFAICT, we await the initial connection before we draw the first screen, so while we are in the pending state while waiting to connect, we don't actually draw anything:
https://github.com/tokio-rs/console/blob/ba60d04a71285f96fd1cbc9b86065c1855792681/console/src/main.rs#L32-L38

not opposed to adding this state for future use, but i don't think it's actually displayed to the user currently

console/src/tasks.rs
```diff
@@ -11,12 +11,20 @@ use std::{
...
+    Disconnected,
+}
+
 #[derive(Default, Debug)]
 pub(crate) struct State {
     tasks: HashMap<u64, Rc<RefCell<Task>>>,
     metas: HashMap<u64, Metadata>,
     last_updated_at: Option<SystemTime>,
     new_tasks: Vec<TaskRef>,
+    connection_state: ConnectionState,
```
hawkw: i'm not sure if this state belongs in the tasks module --- it's really application-wide, not specific to the task state on a runtime. although, if we ever support connecting to multiple separate remotes from one `console` process, it would make sense for this to be part of the per-remote state...but that seems a ways away...

hawkw (commented):
overall, this looks good. i had a couple minor suggestions, and some notes, but i think once the typo is fixed (and maybe we also display the remote address in all states?) i'll be happy to merge this.

the next step will probably be to add reconnection with a fixed interval (say, wait one or two seconds before trying to reconnect, in a loop). we can probably just stick that in the main app loop in the case where we lose the connection, rather than adding `tower::reconnect` or similar.

a backoff can be added afterwards in a follow-up branch.

console/src/main.rs
```diff
@@ -15,6 +15,25 @@ mod tasks;
...
 mod view;
 
+enum ConnectionState {
+    Connected,
+    Disconnected,
+}
+
+async fn connect(
+    target: String,
+    should_delay: bool,
```
hawkw: take it or leave it: this essentially indicates whether or not a connection attempt is a reconnect, right? the delay is the behavior that `connect` does when it's reconnecting --- it might, eventually, also do other behavior. personally, i might call this `is_reconnect` or something.

not a blocker though!

console/src/main.rs
```diff
@@ -15,6 +15,25 @@ mod tasks;
 mod term;
 mod view;
 
+enum ConnectionState {
+    Connected,
+    Disconnected,
+}
+
+async fn connect(
```
hawkw: minor nit: i would generally put this below `main` --- i like to sort of topologically sort functions in a module so that major ones that call the others are at the top of the file, and the smaller helper functions that the main ones call are below them. IMO it makes more sense to read the file that way.

but this is not a big deal and i'd be happy to merge this as-is.

console/src/main.rs
```diff
@@ -15,6 +15,25 @@ mod tasks;
...
+) -> Result<tonic::Streaming<TaskUpdate>, Box<dyn std::error::Error>> {
+    if should_delay {
+        tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
+    }
+
+    let mut client = TasksClient::connect(target).await?;
+    let request = tonic::Request::new(TasksRequest {});
+    let rsp = client.watch_tasks(request).await?;
+    Ok(rsp.into_inner())
+}
```
hawkw: eventually i imagine we might want to move the connection-management code out of `main.rs` into its own module? since there's not a ton of it right now, it's fine here for now, but just a thought.

hawkw (approved):
lgtm --- couple additional comments if you're interested in addressing them, but not blockers.

console/src/main.rs
zaharidichev: Yeah I think so, and if we do that, it seems to make sense to beef it a bit to provide things like backoff

hawkw (approved):
great, looks good to me! thanks for working on this one!

PR #64, opened 2021-07-12
subscriber: remove fmt layer from init

seanmonstar:
With the `fmt` layer being added by default, plus this layer filtering for `RUST_LOG=tokio=trace`, it can be rather noisy. I think we can remove the fmt layer from the default.

### DIFF
--- a/console-subscriber/src/init.rs
+++ b/console-subscriber/src/init.rs
@@ -1,7 +1,7 @@
 use crate::TasksLayer;
 use std::thread;
 use tokio::runtime;
-use tracing_subscriber::{fmt, layer::Layered, prelude::*, EnvFilter, Registry};
+use tracing_subscriber::{layer::Layered, prelude::*, EnvFilter, Registry};
 
 type ConsoleSubscriberLayer = Layered<TasksLayer, Layered<EnvFilter, Registry>>;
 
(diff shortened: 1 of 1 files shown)


hawkw:
Hmm, I'm not sure if we really want this. As discussed in https://github.com/tokio-rs/console/pull/53#discussion_r654644081, because `init` sets the global default tracing subscriber, additional attempts to set a default subscriber in user code will fail or panic...which I think a lot of people may not be happy about out of the box. 

I wonder if a better approach is to use the [per-writer filtering API](https://docs.rs/tracing-subscriber/0.2.19/tracing_subscriber/fmt/writer/trait.MakeWriterExt.html#method.with_filter) added in `tracing-subscriber` 0.2.19? This would allow us to configure the `fmt` subscriber to not *print* `trace` events from Tokio to stdout, but they would still be recorded by the console. Other logging would still work. We could even check if the user-provided `RUST_LOG` env var includes `tokio=trace`, and skip adding the output filter in that case.

This *would* still include the console's _spans_ in the logs for user-generated events, in some cases, since the writer doesn't control whether a span is recorded. But, this is probably an acceptable compromise until tracing gets proper per-layer filtering.

seanmonstar:
Well, I can say having shown the console to various people, I always pipe away the fmt output, since it's super noisy and distracting. I don't believe we want _that_ as the default. I'd even say that just because someone reaches for the easy `init()` function for the console, that doesn't mean they want all that output too. And after I had someone using this branch, when they _did_ want output, it was easy to switch to `builder().with(fmt).init()`.

Do we really expect people to add `console_subscriber::init()` and also try to to do some other crates `init`? And that they won't be able to work out how to fix it?

seanmonstar:
I still think this is probably better than leaving the status quo.

hawkw (approved):
sure, okay!

PR #65, opened 2021-07-12
console: set task poll percentile constraint to exactly 24

seanmonstar:
With it only having a "max" of 24, if the chart needed to take up more space, it would use up all of it and the percentile window would be gone.

### DIFF
--- a/console/src/view/task.rs
+++ b/console/src/view/task.rs
@@ -81,7 +81,7 @@ impl TaskView {
             .constraints(
                 [
                     // 24 chars is long enough for the title "Poll Times Percentiles"
-                    layout::Constraint::Max(24),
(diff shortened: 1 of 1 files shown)


console/src/view/task.rs
```diff
@@ -81,7 +81,7 @@ impl TaskView {
             .constraints(
                 [
                     // 24 chars is long enough for the title "Poll Times Percentiles"
-                    layout::Constraint::Max(24),
+                    layout::Constraint::Length(24),
```
hawkw, line 84: we should probably double-check that the format specs for formatting the percentiles will never exceed a total of 24 chars...but we can do that in a follow-up.

hawkw (approved):
great, thanks for fixing this!

PR #67, opened 2021-07-13
feat(console): factor out conn mgmt, add backoffs

hawkw:
This branch factors out the connection state management (reconnects)
into a new module, so it's not part of `main`. Additionally, it adds a
simple backoff on connection errors.

### DIFF
--- a/console/src/conn.rs
+++ b/console/src/conn.rs
@@ -1,0 +1,147 @@
+use console_api::tasks::{
+    tasks_client::TasksClient, DetailsRequest, TaskDetails, TaskUpdate, TasksRequest,
+};
+use futures::stream::StreamExt;
+use std::{error::Error, pin::Pin, time::Duration};
+use tonic::{transport::Channel, Streaming};
+
+#[derive(Debug)]
+pub struct Connection {
+    target: String,
+    state: State,
+}
+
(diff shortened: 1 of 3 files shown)


console/src/conn.rs
```diff
@@ -0,0 +1,147 @@
...
+        Self {
+            target,
+            state: State::Disconnected(Duration::from_secs(0)),
+        }
+    }
+
+    async fn connect(&mut self) {
+        const MAX_BACKOFF: Duration = Duration::from_secs(5);
+
+        while let State::Disconnected(backoff) = self.state {
```
seanmonstar, line 35: I'd have thought we could make use of tower's `Reconnect` and a backoff layer to do this (doesn't tonic already handle reconnecting?).

seanmonstar (approved):
Either way, factoring this into a separate file is an improvement, so I'm r+.

console/src/conn.rs
hawkw: yeah, we probably could --- i wanted to be able to display to the user whether we are currently connected or reconnecting, and it felt like tracking that in a separate tower middleware would be more work than just writing a simple reconnect loop. if we end up wanting to compose more tower middleware, we can always switch later?

Issue #71, opened 2021-07-15
wish: more distinguished units on numbers


pnkfelix:
Its great that we're using units on our values so that we don't have to fill up too many columns when people will only want 2 or 3 sig figs of info per value.

But: I worry that its too easy to overlook the difference when you see "21ms 34μs" in a column or even "21s 3ms"

What are good ways to mitigate this? Main ideas I can think of:

1. When sorting by column, add blank lines between rows when the unit changes. (The main drawbacks here: it wastes space on aforementioned blank lines. Also, it only works for sorting by the column where confusion occurs, which means people can still overlook it when it happens in an unsorted column.)
2. Use colors to distinguish the units. I.e. color `s` with red, `ms` with yellow, `μs` with green.

hawkw:
Color-coding definitely seems like a good idea. I think it's very important that we never *rely* on colors to communicate anything, so that all the information is still available on terminals that don't support colors; however, it's a great idea to use colors in cases like this, where the information is *present* regardless but the color-coding makes it easier to visually distinguish. I wonder if we should also bold the units so that they stand out better?

I'm less *sure* about the idea of adding blank lines, but I think it's definitely worth trying it out and seeing how it looks? I feel like it's actually more important to be able to distinguish different units in cases where the column _isn't_ the sorting key, since they'll all be mixed together, but anything that makes the data easier to read is definitely worth trying!

hawkw:
Okay, I threw together a quick draft of coloring durations based on their units: 
![image](https://user-images.githubusercontent.com/2796466/126686509-e5fe37f7-82e7-422e-a486-c8d60e97ed6a.png)

seanmonstar:
@pnkfelix Eliza added some colors for different units in #72 (and see picture), seem good enough to mark this done?

hawkw:
> @pnkfelix Eliza added some colors for different units in #72 (and see picture), seem good enough to mark this done?

I left it open at the time since i thought we might also want to investigate adding blank line separators when sorting tables... but i'm still on the fence about whether that's actually a good idea. 

We can close this and look into that idea separately.

PR #72, opened 2021-07-22
feat(console): color durations based on units 

hawkw:
This branch adds color-coding for durations based on their unit, as
described in #71. The colors are a simple gradient, from picoseconds to
whole seconds.

Currently, the unit is detected by formatting the value using the
`fmt::Debug` impl for `Duration`, and looking at the last two characters
to determine the unit. Admittedly, this is kind of janky, but it seemed
like re-implementing all of the `Duration` formatting logic would be
worse.

Right now, this assumes the terminal supports the ANSI 256-color
palette. I'm working on an additional change to detect what color
palettes the terminal supports and select colors based on that, but I
figured I'd do that in a separate PR.

Here's a little example screenshot:

image

### DIFF
--- a/console/src/view/mod.rs
+++ b/console/src/view/mod.rs
@@ -122,3 +122,21 @@ impl Default for View {
 pub(crate) fn bold<'a>(text: impl Into<Cow<'a, str>>) -> Span<'a> {
     Span::styled(text, Style::default().add_modifier(style::Modifier::BOLD))
 }
+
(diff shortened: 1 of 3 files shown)


seanmonstar (approved):
This seems fine, though how does it look on terminals with just 16 colors? I guess if it just looks the same, NBD.

hawkw:
> This seems fine, though how does it look on terminals with just 16 colors? I guess if it just looks the same, NBD.

I have another branch in progress to detect how many colors the terminal supports and pick color palettes based on that...it turns out this is surprisingly complex, though!

PR #75, opened 2021-07-25
refactor: Use u64 task IDs

### LINKED ISSUE #61: Pretty task IDs
Currently, task IDs can be big:

We could use an `AtomicUsize` to generate incrementing task IDs that start off shorter.

oguzbilgener:
Closes  #61 

### Subscriber
- Assign sequential ID numbers to tasks, keep a HashMap of span ID --> task ID.
- Change the task id type to uint64 in protos.

### Console
- Use the u64 task IDs, display them in decimal.
- Remove now-unused `Task::id_hex()`
- Shrink the TID column to a minimum of 4 chars.

### DIFF
--- a/console-subscriber/src/aggregator.rs
+++ b/console-subscriber/src/aggregator.rs
@@ -274,7 +274,7 @@ impl Aggregator {
             // Then send the initial state --- if this fails, the subscription is already dead.
             if stream_sender.send(rx).is_ok()
                 && subscription.update(&proto::tasks::TaskDetails {
-                    task_id: id,
+                    task_id: Some(id.into()),
                     now: Some(now.into()),
                     poll_times_histogram: serialize_histogram(&stats.poll_times_histogram).ok(),
                 })
@@ -324,10 +324,10 @@ impl Aggregator {
         let stats = &self.stats;
         // Assuming there are much fewer task details subscribers than there are
         // stats updates, iterate over `details_watchers` and compact the map.
-        self.details_watchers.retain(|id, watchers| {
-            if let Some(task_stats) = stats.get(id) {
+        self.details_watchers.retain(|&id, watchers| {
+            if let Some(task_stats) = stats.get(&id) {
                 let details = proto::tasks::TaskDetails {
-                    task_id: *id,
+                    task_id: Some(id.into()),
                     now: Some(now.into()),
                     poll_times_histogram: serialize_histogram(&task_stats.poll_times_histogram)
                         .ok(),
@@ -631,7 +631,7 @@ impl Stats {
 impl Task {
     fn to_proto(&self, id: u64) -> proto::tasks::Task {
         proto::tasks::Task {
-            id,
+            id: Some(id.into()),
             // TODO: more kinds of tasks...
             kind: proto::tasks::task::Kind::Spawn as i32,
             metadata: Some(self.metadata.into()),
--- a/console-subscriber/src/aggregator.rs
+++ b/console-subscriber/src/aggregator.rs
@@ -6,7 +6,7 @@ use tokio::sync::{mpsc, Notify};
 
 use futures::FutureExt;
 use std::{
-    collections::HashMap,
+    collections::{hash_map::Entry, HashMap},
     convert::TryInto,
     ops::{Deref, DerefMut},
     sync::{
@@ -21,6 +21,8 @@ use hdrhistogram::{
     serialization::{Serializer, V2SerializeError, V2Serializer},
     Histogram,
 };
+
+pub type TaskId = u64;
 
 pub(crate) struct Aggregator {
     /// Channel of incoming events emitted by `TaskLayer`s.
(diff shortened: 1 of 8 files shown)


console-subscriber/src/aggregator.rs
```diff
@@ -431,6 +445,18 @@ impl Aggregator {
...
+    fn get_or_insert_task_id(&mut self, span_id: span::Id) -> TaskId {
+        match self.task_id_mappings.entry(span_id) {
+            Entry::Occupied(entry) => *entry.get(),
+            Entry::Vacant(entry) => {
+                let task_id = self.task_id_counter;
+                entry.insert(task_id);
+                self.task_id_counter = self.task_id_counter.wrapping_add(1);
+                task_id
+            }
+        }
```
hawkw, line 457: nit/TIOLI: I think this `match` could be replaced with [`Entry::or_insert_with`](https://doc.rust-lang.org/stable/std/collections/hash_map/enum.Entry.html#method.or_insert_with)
```suggestion
        self.task_id_mappings
            .entry(span_id)
            .or_insert_with(|| {
                let task_id = self.task_id_counter;
                entry.insert(task_id);
                self.task_id_counter = self.task_id_counter.wrapping_add(1);
                task_id
            })
```

console/src/view/tasks.rs
```diff
@@ -160,7 +160,7 @@ impl List {
             .header(header)
             .block(block)
             .widths(&[
-                layout::Constraint::Min(20),
+                layout::Constraint::Min(4),
```
hawkw: since the task IDs are assigned sequentially, i think a long-running application could conceivably have more than 4 digits of task ID fairly easily. 

i think the _ideal_ solution is probably to track the highest task ID value we've ever seen, calculate the number of decimal places necessary to display it, and use that number here, rather than hard-coding it to 4. this way, if we suddenly need more characters when drawing a frame, the layout will just change to fit it. but, it would be fine to do that in a separate PR.

proto/tasks.proto
```diff
@@ -52,7 +52,7 @@ message TaskUpdate {
 // A task details update
 message TaskDetails {
     // The task's ID which the details belong to.
-    common.SpanId task_id = 1;
+    uint64 task_id = 1;
```
hawkw: i have a slight preference for continuing to have a semantically-named newtype for these, although i agree they shouldn't be `SpanId`s any longer. I would consider adding a `TaskId` message type in this file.

not a big deal though.

console-subscriber/src/aggregator.rs
```diff
@@ -59,6 +61,12 @@ pub(crate) struct Aggregator {
 
     /// Map of task IDs to task stats.
     stats: TaskData<Stats>,
+
+    /// A counter for the pretty task IDs.
+    task_id_counter: TaskId,
+
+    /// A table that contains the span ID to pretty task ID mappings.
+    task_id_mappings: HashMap<span::Id, TaskId>,
```
hawkw, line 69: a possible optimization for this is adding a custom `Hasher` implementation that always just calls `span::Id::into_u64` and returns that value as the hash, rather than actually  hashing them. but, we can do that in a follow-up...especially because I don't really think the performance impact is particularly significant.

hawkw (commented):
Overall, this mostly looks good to me. I had a couple small suggestions.

Thanks for working on it!

console/src/view/tasks.rs
hawkw: opened #79 for intelligently resizing columns.

console-subscriber/src/aggregator.rs
oguzbilgener: This was actually the first thing I tried, but unfortunately the borrow checker is not happy about the mutation inside the closure 😕 I'm going to do a bit more reading to see if there are any other idiomatic ways.

```
error[E0500]: closure requires unique access to `self` but it is already borrowed
   --> console-subscriber/src/aggregator.rs:449:62
    |
449 |         *self.task_id_mappings.entry(span_id).or_insert_with(|| {
    |          ---------------------                -------------- ^^ closure construction occurs here
    |          |                                    |
    |          |                                    first borrow later used by call
    |          borrow occurs here
450 |             let task_id = self.task_id_counter;
451 |             self.task_id_counter = self.task_id_counter.wrapping_add(1);
    |             -------------------- second borrow occurs due to use of `self` in closure

```

hawkw: oh, huh, never mind — the current approach is fine, then, if the borrow checker doesn't like this.

I think we _could_ "split" the borrows, like this:
```suggestion
        let task_id_mappings = &mut self.task_id_mappings;
        let task_id_counter = &mut self.task_id_counter;
        task_id_mappings
            .entry(span_id)
            .or_insert_with(|| {
                let task_id = *task_id_counter;
                entry.insert(task_id);
                *task_id_counter = task_id_counter.wrapping_add(1);
                task_id
            })
```
but this is the same number of lines as the `match` version, and i'm not sure if it's meaningfully clearer. so, it may not be worth bothering...

hawkw (approved):
okay, this looks good to me, thanks for switching to a newtyped `TaskId`. most of my other suggestions can be addressed as follow-ups.

console/src/view/tasks.rs
```diff
@@ -108,7 +108,7 @@ impl List {
                 let task = task.borrow();
 
                 let mut row = Row::new(vec![
-                    Cell::from(id_width.update_str(task.id_hex()).to_string()),
+                    Cell::from(id_width.update_str(task.id()).to_string()),
```
hawkw: err, whoops, i messed this up while merging:
```suggestion
                    Cell::from(id_width.update_str(task.id().to_string())),
```

Issue #70, opened 2021-07-15
Meta: Record, replay, rewind


seanmonstar:
The current console allows a nice *live* view of the application that it is monitoring. But sometimes things happen quickly, and a user may not have been able to react to inspect the data they wanted. We want to be able to *go back in time*, and have the Console show those events again, even being able to pause.

We've written a proper design document exploring this feature set here: https://hackmd.io/6xPTpWK5RGO_eGygI9QEPA?view

Actions based on the design document:

- [x] #85
- [x] #84
- [ ] #96
- [ ] Interactive rewind (step back)
- [ ] Watch a value for a change or to equal an expression

carllerche:
I think we need to work through the UI of this. My gut is, what we *probably* want is a per-task event log (kind of like tracing log output, but filtered per task), and then, one could pick a line and use the console frozen at that point in time maybe?

WesleyAC:
What ended up happening with this? This feature would be very useful for me, I was happy to see that recording to a file is possible, but it seems that there is no software that can actually do anything particularly useful with that file at the moment?

hds:
There hasn't been any movement on this topic in Tokio Console since 2021 I'm afraid.

I've had a look at some alternatives myself, but haven't got anywhere yet.

PR #78, opened 2021-07-30
Add pause and resume

### LINKED ISSUE #85: Pausing in the console

seanmonstar:
A first step to address #70. You can now press the spacebar to "pause" the console. The subscriber will continue to process events, but as along as the client is "paused", it won't send any task updates. Resuming will bring the client back to "live", getting all updates again.

paused

I've tried a few different UI things, from making the whole background of the top red, to just making the word red... It's fine if others feel there's a better or prettier way to show this, but to keep moving forward, I just picked something that seemed good enough.

Closes #85

### DIFF
--- a/proto/tasks.proto
+++ b/proto/tasks.proto
@@ -9,6 +9,8 @@ import "common.proto";
 service Tasks {
     rpc WatchTasks(TasksRequest) returns (stream TaskUpdate) {}
     rpc WatchTaskDetails(DetailsRequest) returns (stream TaskDetails) {}
+    rpc Pause(PauseRequest) returns (PauseResponse) {}
+    rpc Resume(ResumeRequest) returns (ResumeResponse) {}
 }
 
 message TaskId {
@@ -20,6 +22,12 @@ message TasksRequest {
 
 message DetailsRequest {
     TaskId id = 1;
+}
+
+message PauseRequest {
+}
+
+message ResumeRequest {
 }
 
 // A task state update.
@@ -62,6 +70,12 @@ message TaskDetails {
 
     // HdrHistogram.rs `Histogram` serialized to binary in the V2 format
     optional bytes poll_times_histogram = 3;
+}
+
+message PauseResponse {
+}
+
+message ResumeResponse {
 }
 
 // Data recorded when a new task is spawned.
--- a/console-subscriber/src/aggregator/mod.rs
+++ b/console-subscriber/src/aggregator/mod.rs
@@ -1,6 +1,6 @@
 use crate::{record::Recorder, WatchRequest};
 
-use super::{Event, WakeOp, Watch, WatchKind};
+use super::{Command, Event, WakeOp, Watch};
 use console_api as proto;
 use tokio::sync::{mpsc, Notify};
 
@@ -33,7 +33,7 @@ pub(crate) struct Aggregator {
     events: mpsc::Receiver<Event>,
 
     /// New incoming RPCs.
-    rpcs: mpsc::Receiver<WatchKind>,
+    rpcs: mpsc::Receiver<Command>,
 
     /// The interval at which new data updates are pushed to clients.
     publish_interval: Duration,
@@ -70,12 +70,21 @@ pub(crate) struct Aggregator {
 
     /// A sink to record all events to a file.
     recorder: Option<Recorder>,
+
+    /// The time "state" of the aggregator, such as paused or live.
+    temporality: Temporality,
 }
 
 #[derive(Debug)]
 pub(crate) struct Flush {
     pub(crate) should_flush: Notify,
     pub(crate) triggered: AtomicBool,
+}
+
+#[derive(Debug)]
+enum Temporality {
+    Live,
+    Paused,
 }
 
 struct Stats {
@@ -137,7 +146,7 @@ impl Default for Stats {
 impl Aggregator {
     pub(crate) fn new(
         events: mpsc::Receiver<Event>,
-        rpcs: mpsc::Receiver<WatchKind>,
+        rpcs: mpsc::Receiver<Command>,
         builder: &crate::Builder,
     ) -> Self {
         Self {
@@ -160,6 +169,7 @@ impl Aggregator {
                 .recording_path
                 .as_ref()
                 .map(|path| Recorder::new(path).expect("creating recorder")),
+            temporality: Temporality::Live,
         }
     }
 
@@ -173,7 +183,10 @@ impl Aggregator {
             let should_send = tokio::select! {
                 // if the flush interval elapses, flush data to the client
                 _ = publish.tick() => {
-                    true
+                    match self.temporality {
+                        Temporality::Live => true,
+                        Temporality::Paused => false,
+                    }
                 }
 
                 // triggered when the event buffer is approaching capacity
@@ -183,16 +196,22 @@ impl Aggregator {
                     false
                 }
 
-                // a new client has started watching!
-                subscription = self.rpcs.recv() => {
-                    match subscription {
-                        Some(WatchKind::Tasks(subscription)) => {
+                // a new command from a client
+                cmd = self.rpcs.recv() => {
+                    match cmd {
+                        Some(Command::WatchTasks(subscription)) => {
                             self.add_task_subscription(subscription);
                         },
-                        Some(WatchKind::TaskDetail(watch_request)) => {
+                        Some(Command::WatchTaskDetail(watch_request)) => {
                             self.add_task_detail_subscription(watch_request);
                         },
-                        _ => {
+                        Some(Command::Pause) => {
+                            self.temporality = Temporality::Paused;
+                        }
+                        Some(Command::Resume) => {
+                            self.temporality = Temporality::Live;
+                        }
+                        None => {
                             tracing::debug!("rpc channel closed, terminating");
                             return;
                         }
--- a/console-subscriber/src/lib.rs
+++ b/console-subscriber/src/lib.rs
@@ -52,7 +52,7 @@ pub struct TasksLayer {
 }
 
 pub struct Server {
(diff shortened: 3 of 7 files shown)


console-subscriber/src/aggregator.rs
```diff
@@ -67,6 +70,12 @@ pub(crate) struct Flush {
     pub(crate) triggered: AtomicBool,
 }
 
+#[derive(Debug)]
+enum Temporality {
```
hawkw: cute name :)

assuming this is an enum because we might want to add replays or other forms of time travel?

console-subscriber/src/aggregator.rs
```diff
@@ -59,6 +59,9 @@ pub(crate) struct Aggregator {
 
     /// Map of task IDs to task stats.
     stats: TaskData<Stats>,
+
+    /// The time "state" of the aggregator, such as paused or live.
+    temporality: Temporality,
```
hawkw: Having one temporality for the aggregator seems like a potential issue. Since multiple consoles can connect to the same application, if one user pauses their console, sending the pause command to the application, all consoles will be paused. Similarly, if I unpause my console, everyone else's console unpauses, too. It doesn't _seem_ like this is the behavior we want when multiple consoles are connected?

If we're going to implement pausing in the aggregator-side (rather than in the UI-side), it seems like we would want to track a separate temporality for each client connection, and decide whether we are going to send updates to each client independently?

proto/tasks.proto
```diff
@@ -9,6 +9,8 @@ import "common.proto";
 service Tasks {
     rpc WatchTasks(TasksRequest) returns (stream TaskUpdate) {}
     rpc WatchTaskDetails(DetailsRequest) returns (stream TaskDetails) {}
+    rpc Pause(PauseRequest) returns (PauseResponse) {}
+    rpc Resume(ResumeRequest) returns (ResumeResponse) {}
 }
```
hawkw, line 14: Hmm, as per my comment above, I think if we want to implement pausing in the aggregator rather than in the UI, we would want to have a separate temporality for each client connection, rather than adding new RPCs that globally mutate whether the app is paused or resumed.

IMO, the best way to do this would be to turn the `WatchTasks` and `WatchTaskDetails` RPCs into bidirectional streaming RPCs, like
```suggestion
service Tasks {
    rpc WatchTasks(stream WatchCommand) returns (stream TaskUpdate) {}
    rpc WatchTaskDetails(stream WatchCommand) returns (stream TaskDetails) {}
}

message WatchCommand {
    enum Kind {
        PAUSE = 0;
        RESUME = 1;
    } 
    Kind command_kind = 1;
}
```
or something? Then, when a `PAUSE` message is sent on a given RPC, the aggregator would pause sending updates to that client until it receives a `RESUME` command from the same client.

hawkw (commented):
Would you mind explaining the rationale for implementing this in the subscriber, rather than purely in the console? 

After reading the PR description, my naive assumption was that pausing would work by having the console UI maintain two copies of the state, one that's being updated as new data comes in, and one that's being displayed. When the console is paused, in this design, we would freeze the current state, duplicate it, and continue updating the duplicate rather than the displayed state; when unpausing, we would replace the displayed state with the updated state. Without having thought about it a great deal, this _seems_ like a simpler approach with fewer moving parts (and a conceptually cleaner separation of concerns), so I'd love to know if there's something I'm overlooking.

seanmonstar:
I did spend a bit of time thinking about the different ways to implement this:

- In the console app, but the downsides are that you can't fetch "details" (like task or resources) in the paused state, since the subscriber has kept going. It also means that new viewing applications, like perhaps a web view, need to re-implement time travel entirely.
- In the subscriber app, but each pause received by a client means the state gets cloned into a frozen version. It's a bit more complicated to support, and also more "expensive".
- As this PR does, the entire subscriber state is affected by any client commands. This means multiple clients can't have separate state.

After discussing with others and some thinking, I believe the ability to support multiple clients is just because of the simple design of the current server. It doesn't seem like an actual requirement or use case that people would need. Customer interviews didn't show they expected to have multiple members of their team all connecting to the same instance. Anyone who did want to explore a process individually could copy an eventual recording file and play it over locally.

Does it seem that would be something users would likely expect to do? We certainly *can* support it, but it's more complicated in code and and costs more in CPU and memory.

hawkw:
> * In the console app, but the downsides are that you can't fetch "details" (like task or resources) in the paused state, since the subscriber has kept going. It also means that new viewing applications, like perhaps a web view, need to re-implement time travel entirely.
> 
> * In the subscriber app, but each pause received by a client means the state gets cloned into a frozen version. It's a bit more complicated to support, and also more "expensive".

Hmm, I still think that cloning the state in the client is the "correct" thing, since you can continue to fetch details. My thought process is that it's better to make the console app do as much of the "expensive" work as possible, to support debugging in prod — there is, of course, going to be some performance hit on the instrumented application no matter what we do, but it would be nice to minimize it as much as possible, and make the console app do the expensive work.

I'm also not entirely convinced that "new viewing applications, like perhaps a web view, need to re-implement time travel entirely" is necessarily all that bad a downside, since there's already a lot of stuff that a web view or other UI would have to re-implement.

I guess in general, my bias is for putting as much intelligence in the UI as possible, and having the instrumented application just focus on doing anything that _has_ to happen in the app (e.g. anchoring timestamps) and getting the data out as efficiently as possible. But, I'm okay with going against that bias when it makes sense, situationally.

> After discussing with others and some thinking, I believe the ability to support multiple clients is just because of the simple design of the current server. It doesn't seem like an actual requirement or use case that people would need. Customer interviews didn't show they expected to have multiple members of their team all connecting to the same instance. Anyone who did want to explore a process individually could copy an eventual recording file and play it over locally.
> 
> Does it seem that would be something users would likely expect to do? We certainly _can_ support it, but it's more complicated in code and and costs more in CPU and memory.

Hmm...to me, it seemed like a reasonable use case we might want to be able to support, but if you've actually talked to people and nobody really wants it, I'm willing to be convinced otherwise. I will say, though, that the future uses of multiple clients I was imagining was less about two or more human users interactively debugging the same process, and more about a human user interactively debugging a process while data from that process is also being collected/aggregated/archived by some automated tool. but, that doesn't exist yet, so maybe it's a problem we only need to solve when those tools actually exist...

FWIW I'd be interested to hear more about the conversations you've had with potential users, if you don't mind sharing?

----

In summary: I'm fine to move forward with this approach if you're convinced it's the best thing, I just wanted to make sure we talked through it first.

seanmonstar:
>  if you've actually talked to people and nobody really wants it, I'm willing to be convinced otherwise. [..] I'd be interested to hear more about the conversations you've had with potential users, if you don't mind sharing?

We showed something very similar to [this demo](https://www.youtube.com/watch?v=JGCewPUvF70) live to a few teams using Rust, and then asked what more they'd want to see, and how they'd expect to use it. [This document](https://hackmd.io/6xPTpWK5RGO_eGygI9QEPA?view) describes the feature work that we understood they wanted. (I realized this doesn't include the original conversations...)

> the future uses of multiple clients I was imagining was less about two or more human users interactively debugging the same process, and more about a human user interactively debugging a process while data from that process is also being collected/aggregated/archived by some automated tool.

The second part of the document I just linked mentions recording the events automatically (well, with a config option), and that would continue to happen "live".

> my bias is for putting as much intelligence in the UI as possible, and having the instrumented application just focus on doing anything that has to happen in the app (e.g. anchoring timestamps) and getting the data out as efficiently as possible.

This was my initial feeling too... And then it seemed somewhere along the way, the subscriber started to hold most of the state, and the console just "viewed" it. I've tried to go "remember" why that is, and it looks like it was trying to cut down on the amount of IO from the subscriber. But in order to support more of these temporal features, such as recording and stepping backward (rewind), we'll _need_ to save the individual events somewhere anyways...

I now am wondering about that decision. Perhaps sending a write/packet per event is the bad part, but just buffering a reasonable amount, and/or maybe  enabling gRPC compression (tonic now has it), would make the cost small enough that we don't need to really keep any state in the subscriber...

hawkw:
> I now am wondering about that decision. Perhaps sending a write/packet per event is the bad part, but just buffering a reasonable amount, and/or maybe enabling gRPC compression (tonic now has it), would make the cost small enough that we don't need to really keep any state in the subscriber...

Probably we should do some actual testing to try and figure out how much data actually goes on the wire in a real program; it would be nice to have some data points for what is basically reasonable.

hawkw:
> > if you've actually talked to people and nobody really wants it, I'm willing to be convinced otherwise. [..] I'd be interested to hear more about the conversations you've had with potential users, if you don't mind sharing?
> 
> We showed something very similar to [this demo](https://www.youtube.com/watch?v=JGCewPUvF70) live to a few teams using Rust, and then asked what more they'd want to see, and how they'd expect to use it. [This document](https://hackmd.io/6xPTpWK5RGO_eGygI9QEPA?view) describes the feature work that we understood they wanted. (I realized this doesn't include the original conversations...)

This doc is really nice, btw — I realize now you had posted a link in #70, but I think I'd overlooked that at the time. It would be nice to get some of the various features proposed there written up as separate issues, eventually.

hawkw:
@seanmonstar any updates on this?

seanmonstar:
So, I think it'd be good to carve out an issue to think about state in the subscriber vs the console. I suspect we can have the subscriber spit out events similar to the way it records to a file, and have the console build up the state, as long as we do the right kind of buffering.

But I also think this approach for pausing works at the moment, and the "cost" only exists if you actually connect and pause. We'd probably need to move pausing entirely to the console, but I think to do so would need answering the bigger question in the previous paragraph. So, if that seems fine, I can fixup the conflicts and open a separate issue around general state. Whatcha think?

hawkw (approved):
I'm fine to roll forwards with merging this once the merge conflicts are fixed :)

PR #79, opened 2021-08-02
feat(console): programmatically determine column widths

hawkw:
Currently, we use hard-coded fixed character widths for all columns in the
console's task list view. This works fine for time values, which are
displayed as decimals with fixed maximum precision. However, for other
stuff (such as targets and task IDs), the length can vary widely, and
space is often wasted when all values are shorter than the hard-coded
max width.

This branch updates the console to compute the maximum number of
characters needed to display the longest entry in a column every time it
draws the table, and use exactly that many characters. Therefore, we
waste significantly fewer chars on making the table columns long enough
to display "any reasonable value".

I was a bit concerned that resizing these columns would potentially be
visually disruptive and make the table harder to read. Testing it out in
the demo app, the resizing isn't really noticeable...or at least, it
didn't bother *me*. If anyone else wants to try this branch out and let
me know if they find it really disruptive, I'd appreciate that. It would
be fine to not move forward with this change if it hurts the readability
of the table.

### DIFF
--- a/console/src/view/mod.rs
+++ b/console/src/view/mod.rs
@@ -1,5 +1,5 @@
 use crate::{input, tasks::State};
-use std::borrow::Cow;
+use std::{borrow::Cow, cmp};
 use tui::{
     layout,
     style::{self, Style},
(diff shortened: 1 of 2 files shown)


seanmonstar (approved):

zaharidichev (approved):
Tried it out. I think it looks pretty smooth

PR #80, opened 2021-08-04
Bump dependencies

vorot93:


seanmonstar (approved):

console/src/tasks.rs
```diff
@@ -373,14 +373,18 @@ impl From<proto::tasks::Stats> for Stats {
...
+            created_at: pb
+                .created_at
+                .expect("task span was never created")
+                .try_into()
+                .unwrap(),
             wakes: pb.wakes,
             waker_clones: pb.waker_clones,
             waker_drops: pb.waker_drops,
-            last_wake: pb.last_wake.map(Into::into),
+            last_wake: pb.last_wake.map(|v| v.try_into().unwrap()),
```
hawkw, line 387: hmm, not totally sure if panicking is the right approach when the timestamp conversion from protobuf fails...but i'm guessing it's what would have happened anyway in the previous version of `prost`?

we can fix this later.

hawkw (approved):
thanks!

we should probably get dependabot set up so that this can be automated, in the future...

PR #82, opened 2021-08-04
fix(console): stop truncating task fields column

hawkw:
Currently, there's a bug where the console's task list view truncates
the fields column to the first ten characters, even when there are
characters remaining in the window to display more of the fields column.
This is due to `tui`'s layout constraints system apparently choosing to
give that column only the minimum number of characters specified,
instead of giving it all the remaining characters not used by other
columns.

I couldn't figure out how to get `tui` to actually do what we want
(there doesn't seem to be a `Constrait` for saying "use any remaining
characters not needed for satisfying previous layout constraints"), so I
just kind of hacked around it by adding up all the other columns,
subtracting their widths from the total width of the window, and
specifying that as a "minimum" constraint.

This is kinda ugly, but it seems to work.

### Before:
image

### After:
image

### DIFF
--- a/console/src/view/mod.rs
+++ b/console/src/view/mod.rs
@@ -161,4 +161,8 @@ impl Width {
     pub(crate) fn constraint(&self) -> layout::Constraint {
         layout::Constraint::Length(self.curr)
     }
+
+    pub(crate) fn chars(&self) -> u16 {
+        self.curr
+    }
 }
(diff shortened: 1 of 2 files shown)


hawkw:
Apparently there is some known weird behavior with tui's constraint system: https://github.com/fdehau/tui-rs/issues/499

PR #83, opened 2021-08-04
stop rendering empty fields, refactor field conversion

hawkw:
Depends on #81 (since it touches the same code and would cause merge conflicts).

* feat(console): don't display fields with "" values

Now that Tokio supports optional user-provided task names
(in tokio-rs/tokio#3881), the task spans it generates may contain empty
fields (if the user does not provide a name for a spawned task).
However, the console always displays the `task.name` field, even when
they are empty. This wastes space, and looks kind of ugly.

This branch updates the console to skip displaying fields whose value is
the empty string.

### Before:
image

### After:
image

* refac(console): factor out field from pb conversion

This factors out the protobuf `Field` -> internal `Field` conversion
into a method, and cleans it up a bit. No functional change, except that
we now take the proto `Field` type by value, so we don't have to clone
strings out of it like we did previously (due to taking it by
reference). I also added the metadata ID to the `Metadata` type, so that
we can continue asserting that the metadata's ID matches the expected
one for the field, without having to pass that in separately.

### DIFF
--- a/console/Cargo.toml
+++ b/console/Cargo.toml
@@ -19,3 +19,5 @@ crossterm = { version = "0.20", features = ["event-stream"] }
 color-eyre = "0.5"
 hdrhistogram = { version = "7.3.0", default-features = false, features = ["serialization"] }
 h2 = "0.3"
+regex = "1.5"
+once_cell = "1.8"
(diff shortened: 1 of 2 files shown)


seanmonstar (approved):

zaharidichev (approved):
LGTM

PR #86, opened 2021-08-10
subscriber: add ability to record events to a file

### LINKED ISSUE #84: Recording to a file

seanmonstar:
This adds an option (default is off for now) to record all relevant events to a file, as described in #70, with the goal of being able to share the file and/or replay the run in the Console after the fact.

- To try to prevent blocking file system writes from interfering, the events are serialized to a group of buffers, and a worker thread is spawned that juggles them and tries to write them to disk.
- The data format of JSON was somewhat arbitrarily chosen. We could chose a different format if we wanted to try to reduce the size on the disk.

Closes #84

### DIFF
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -1,6 +1,7 @@
 use super::{Server, TasksLayer};
 use std::{
     net::{SocketAddr, ToSocketAddrs},
+    path::PathBuf,
     time::Duration,
 };
 
@@ -23,6 +24,9 @@ pub struct Builder {
 
     /// The address on which to serve the RPC server.
     pub(super) server_addr: SocketAddr,
+
+    /// If and where to save a recording of the events.
+    pub(super) recording_path: Option<PathBuf>,
 }
 
 impl Default for Builder {
@@ -33,6 +37,7 @@ impl Default for Builder {
             publish_interval: TasksLayer::DEFAULT_PUBLISH_INTERVAL,
             retention: TasksLayer::DEFAULT_RETENTION,
             server_addr: SocketAddr::new(Server::DEFAULT_IP, Server::DEFAULT_PORT),
+            recording_path: None,
         }
     }
 }
@@ -100,6 +105,14 @@ impl Builder {
         }
     }
 
+    /// Sets the path to record the events to the file system.
+    pub fn recording_path(self, path: impl Into<PathBuf>) -> Self {
+        Self {
+            recording_path: Some(path.into()),
+            ..self
+        }
+    }
+
     /// Completes the builder, returning a [`TasksLayer`] and [`Server`] task.
     pub fn build(self) -> (TasksLayer, Server) {
         TasksLayer::build(self)
@@ -112,6 +125,7 @@ impl Builder {
     /// | `TOKIO_CONSOLE_RETENTION_SECS`      | The number of seconds to accumulate completed tracing data                | 3600s (1h)        |
     /// | `TOKIO_CONSOLE_BIND`                | a HOST:PORT description, such as `localhost:1234`                         | `127.0.0.1:6669`  |
     /// | `TOKIO_CONSOLE_PUBLISH_INTERVAL_MS` | The number of milliseconds to wait between sending updates to the console | 1000ms (1s)       |
+    /// | `TOKIO_CONSOLE_RECORD_PATH`         | The file path to save a recording                                         | None              |
     pub fn with_default_env(mut self) -> Self {
         if let Ok(retention) = std::env::var("TOKIO_CONSOLE_RETENTION_SECS") {
             self.retention = Duration::from_secs(
@@ -135,6 +149,10 @@ impl Builder {
                     .parse()
                     .expect("TOKIO_CONSOLE_PUBLISH_INTERVAL_MS must be an integer"),
             );
+        }
+
+        if let Ok(path) = std::env::var("TOKIO_CONSOLE_RECORD_PATH") {
+            self.recording_path = Some(path.into());
         }
 
         self
--- a/console-subscriber/src/lib.rs
+++ b/console-subscriber/src/lib.rs
@@ -1,4 +1,5 @@
 use console_api as proto;
+use serde::Serialize;
 use tokio::sync::{mpsc, oneshot};
 
 use std::{
@@ -18,10 +19,12 @@ mod aggregator;
 mod builder;
 mod callsites;
 mod init;
+mod record;
 
 use aggregator::Aggregator;
 pub use builder::Builder;
 use callsites::Callsites;
+use record::Recorder;
 
 pub use init::{build, init};
 
@@ -30,6 +33,7 @@ use crate::aggregator::TaskId;
 pub struct TasksLayer {
     tx: mpsc::Sender<Event>,
     flush: Arc<aggregator::Flush>,
+    recorder: Option<Recorder>,
     spawn_callsites: Callsites,
     waker_callsites: Callsites,
 }
@@ -91,6 +95,7 @@ enum Event {
     },
 }
 
+#[derive(Clone, Copy, Serialize)]
 enum WakeOp {
     Wake,
     WakeByRef,
@@ -122,6 +127,7 @@ impl TasksLayer {
             ?config.publish_interval,
             ?config.retention,
             ?config.server_addr,
+            ?config.recording_path,
             "configured console subscriber"
         );
 
@@ -130,6 +136,12 @@ impl TasksLayer {
 
         let aggregator = Aggregator::new(events, rpcs, &config);
         let flush = aggregator.flush().clone();
+
+        let recorder = config
+            .recording_path
+            .as_ref()
+            .map(|path| Recorder::new(path).expect("creating recorder"));
+
         let server = Server {
             aggregator: Some(aggregator),
             addr: config.server_addr,
@@ -139,6 +151,7 @@ impl TasksLayer {
         let layer = Self {
             tx,
             flush,
+            recorder,
             spawn_callsites: Callsites::default(),
             waker_callsites: Callsites::default(),
         };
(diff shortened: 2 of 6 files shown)


hawkw:
>     * The data format of JSON was somewhat arbitrarily chosen. We could chose a different format if we wanted to try to reduce the size on the disk.

I'm a bit curious about serializing these in JSON.

Is the intent that the files will be consumed by tools other than the console command-line application (and other potential console tools like a web UI)? If so, it does make sense to use JSON so that you can use `jq` or whatever. However, if the intent was that these files would specifically be used with the console CLI app (and hypothetically other console tools), I would probably have serialized them using protobuf --- that way, the same event types could be streamed over the wire without having to do any conversion between different representations (and, we don't introduce an additional serialization format, which seems conceptually simpler). 

I'm definitely open to recording these in JSON, but I wanted to ask more about the eventual intent for how this will be consumed.

seanmonstar:
I had no further goals of utilizing it as JSON, it was mostly because it was easy to verify it was working by viewing the file afterwards and seeing what was written. I briefly thought of protobuf, since it might help reduce the file size. I also thought some smarter strategy of reducing repeated strings may help 🤷‍♂️

Another thing I liked about JSON (well, serde) was being able to serialize with slices. I haven't used prost in a while, but I assumed (perhaps wrongly, we'll see) that `Message` didn't work as well with borrows. I didn't want to have to make yet another temporary copy.

hawkw:
Okay, cool...I'm flexible on this, but I wanted to talk through it. I do think that potentially sharing types with the wire format, and the file size reduction, are both potentially big points in favor of protobuf...

> Another thing I liked about JSON (well, serde) was being able to serialize with slices. I haven't used prost in a while, but I assumed (perhaps wrongly, we'll see) that `Message` didn't work as well with borrows. I didn't want to have to make yet another temporary copy.

Testing it out, I think you're right that `prost-derive` doesn't work with slices; this code
```rust
use prost::Message;

#[derive(Clone, PartialEq, Message)]
pub struct ProstTest {
    #[prost(string, tag = "1")]
    cool_string: &'static str,
}
```
errors like this
```
   Compiling prost-test v0.1.0 (/home/eliza/Code/prost-test)
error[E0308]: mismatched types
 --> src/main.rs:3:28
  |
3 | #[derive(Clone, PartialEq, Message)]
  |                            ^^^^^^^ expected struct `String`, found `&str`
  |
  = note: expected reference `&String`
             found reference `&&'static str`
  = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0308]: mismatched types
 --> src/main.rs:3:28
  |
3 | #[derive(Clone, PartialEq, Message)]
  |                            ^^^^^^^ expected struct `String`, found `&str`
  |
  = note: expected mutable reference `&mut String`
             found mutable reference `&mut &'static str`
  = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0599]: no method named `clear` found for reference `&'static str` in the current scope
 --> src/main.rs:3:28
  |
3 | #[derive(Clone, PartialEq, Message)]
  |                            ^^^^^^^ method not found in `&'static str`
  |
  = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0308]: mismatched types
 --> src/main.rs:3:28
  |
3 | #[derive(Clone, PartialEq, Message)]
  |                            ^^^^^^^
  |                            |
  |                            expected `&str`, found struct `String`
  |                            help: consider borrowing here: `&Message`
  |
  = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to 4 previous errors
```
which is a bummer!

It's probably possible to work around this with a manual `Message` impl, but I'm not sure if that's really worth the effort...

console-subscriber/src/builder.rs
```diff
@@ -100,6 +105,14 @@ impl Builder {
         }
     }
 
+    /// Sets the path to record the events to the file system.
```
hawkw, line 108: nit: maybe this should note that the recording is in json?

console-subscriber/src/builder.rs
```diff
@@ -112,6 +125,7 @@ impl Builder {
     /// | `TOKIO_CONSOLE_RETENTION_SECS`      | The number of seconds to accumulate completed tracing data                | 3600s (1h)        |
     /// | `TOKIO_CONSOLE_BIND`                | a HOST:PORT description, such as `localhost:1234`                         | `127.0.0.1:6669`  |
     /// | `TOKIO_CONSOLE_PUBLISH_INTERVAL_MS` | The number of milliseconds to wait between sending updates to the console | 1000ms (1s)       |
+    /// | `TOKIO_CONSOLE_RECORD_PATH`         | The file path to save a recording                                         | None              |
```
hawkw, line 128: nit: maybe should note that the recording is in JSON?

console-subscriber/src/lib.rs
```diff
@@ -130,6 +136,12 @@ impl TasksLayer {
 
         let aggregator = Aggregator::new(events, rpcs, &config);
         let flush = aggregator.flush().clone();
+
+        let recorder = config
+            .recording_path
+            .as_ref()
+            .map(|path| Recorder::new(path).expect("creating recorder"));
```
hawkw: panicking here is probably fine for now...

console-subscriber/src/record.rs
```diff
@@ -0,0 +1,224 @@
...
+struct SerializeField<'a>(&'a proto::Field);
+
+impl Recorder {
+    pub(crate) fn new(path: &Path) -> io::Result<Self> {
+        let buf = Arc::new(Mutex::new(RecordBuf::new()));
+        let buf2 = buf.clone();
+        let file = std::fs::File::create(path)?;
+
+        let worker = std::thread::Builder::new()
+            .name("console/subscriber/recorder/io".into())
```
hawkw, line 85: very silly nitpick, feel free to ignore: is using slashes for thread names a widespread convention? would it be "rustier" to use `::`?
```suggestion
            .name("console_subscriber::recorder::io".into())
```

console-subscriber/src/lib.rs
```diff
@@ -174,6 +187,12 @@ impl TasksLayer {
 
     fn send(&self, event: Event) {
         use mpsc::error::TrySendError;
+
+        // always be recording...
+        if let Some(ref recorder) = self.recorder {
+            recorder.record(&event);
+        }
```
hawkw: I wonder if this is really the right place to do the recording, or if it would make more sense to put it inside the aggregator task on receiving an event? The buffer management strategy for the recorder seems like it will be most efficient when the writes are all happening on one thread, since writers will all have to contend the lock on the single write buffer. 

I wonder if it would be better to perform the `record` call on the aggregator task after it receives an event, instead. That way, a single thread will be performing all the writes, and the application thread that actually observed the event doesn't have to wait to contend the lock?

Also, performing the write inside the aggregator task would mean that the file writes are done in the same order that the aggregator observes the event, which may not be the case otherwise?

console-subscriber/src/record.rs
```diff
@@ -0,0 +1,224 @@
...
+        let bytes = dst.buf.lock().unwrap().take();
+        match dst.file.write_all(&bytes) {
+            Ok(()) => {
+                dst.buf.lock().unwrap().put(bytes);
+            }
+            Err(_e) => {
+                // TODO: what to do if file error?
+            }
+        }
+    }
```
hawkw, line 178: we probably want some way to ensure that the file is flushed when the program is shut down (e.g. if it panics, we don't want to miss the last event).

but, we can probably do that in a follow-up...

console-subscriber/src/record.rs
```diff
@@ -0,0 +1,224 @@
...
+            _ => return,
+        };
+
+        self.write(&event);
+    }
+
+    fn write<T: Serialize>(&self, val: &T) {
+        let mut buf = self.buf.lock().unwrap();
+        serde_json::to_writer(&mut buf.bytes, val).expect("json");
+        buf.bytes.push(b'\n');
```
hawkw, line 132: technically, this means that the output format is not actually JSON, but _newline-delimited_ JSON, where each line is a separate JSON object. this means that (e.g.) `jq` doesn't like the format without some pre-processing so that it's invoked separately on each line.

i wonder if we want to change the output format so that the whole file is a valid JSON object. we could emit something like this instead:
```
{
    "header": <header>,
    "events": [
        <event 1>,
        <event 2>,
        ...
        <event n>
    ]
}
```
that way, the whole file would parse as JSON and could be more easily used with tools like `jq`.

the downside to this is just that we would have to detect shutdowns and ensure that we properly emit the close delimiters for the JSON object...which would introduce more complexity for sure. wdyt?

hawkw (approved):
okay, i think i'm onboard with this approach. i had a few small suggestions but most of them are addressable in follow-up PRs.

as a datapoint re: file size, i checked out this PR and ran the example app for 2 minutes. this produces a 166 KB JSON file, which is not that bad...but, I think the example app may not be the most representative example, because it spends most of its time sleeping for durations >1 second, so there are probably significantly fewer waker ops than a real program doing IO under load... may be worth testing with a more representative workload if we're concerned about file sizes.

console-subscriber/src/builder.rs
seanmonstar: I personally find the data _format_ an implementation detail, and wouldn't want people to depend on it. I know people _will_, but I'd like to believe we could change it anyways...

console-subscriber/src/record.rs
seanmonstar: I truly have no idea if there's much of a convention... I think the only thing I ever see is separating all words with dashes, but I don't love that. The thread names is just for debugging purposes, so 🤷

console-subscriber/src/lib.rs
seanmonstar: Yea, putting it in the aggregator thread could be fine too. 👍

console-subscriber/src/record.rs
seanmonstar: I purposefully made it separate objects per newline to mimic _streaming_ JSON, so that later on parsing events can be done in slices, instead of needing to load the entire file before understanding any of it. (I know we could make a smarter parser, but this pattern is also common.) I also didn't want an abrupt shutdown or file error to make it impossible to parse the rest of the file.

hawkw: Yup, makes sense!

hawkw: it would maybe be nice for `jq` to just work, but i think since the _primary_ intended use case for these files is to parse them with the console app, optimizing for parsing in slices is a good call.

PR #88, opened 2021-08-14
feat(console): comprehensive color configuration

hawkw:
This branch updates the console application to:
- Detect what colors are supported by the current terminal and choose
  them accordingly
- Allow user settings to override the color palette
- Allow user settings to override whether particular UI elements are
  colored

This change is rather complicated. It introduces `clap` for CLI
arguments, to allow the user to override the color palette.

When the color palette is _not_ overridden, the console will determine
what colors the terminal supports via the following process:
- If the `COLORTERM` environment variable is set to "24bit" or
  "truecolor", this indicates that the terminal supports 24-bit RGB
  color. In that case, all colors are used.
- Otherwise, the console attempts to use `tput colors` to query the
  `terminfo` database to see what colors the current terminal supports.
- If the call to `tput colors` fails, we assume no colors are supported.

Additionally, new CLI args allow the user to toggle whether durations
are color-coded, and whether terminated tasks are dimmed. In the future,
we will probably also want to add support for parsing these from a
config file --- in general, users probably won't want to set these
preferences from the command line every time they run the console.
However, I decided to punt on adding a config file in this PR --- the
addition of new CLI arg parsing logic is a big enough change. We can add
a simple TOML file or something in a follow-up change.

### DIFF
--- a/console/src/config.rs
+++ b/console/src/config.rs
@@ -78,7 +78,7 @@ pub struct ColorToggles {
 
 impl Config {
     pub fn trace_init(&mut self) -> color_eyre::Result<()> {
-        let filter = std::mem::replace(&mut self.env_filter, Default::default());
+        let filter = std::mem::take(&mut self.env_filter);
         use tracing_subscriber::prelude::*;
 
         // If we're on a Linux distro with journald, try logging to the system
--- a/console/src/config.rs
+++ b/console/src/config.rs
@@ -1,0 +1,162 @@
+use crate::view::Palette;
+use clap::{ArgGroup, Clap, ValueHint};
+use std::process::Command;
+use tonic::transport::Uri;
+
+#[derive(Clap, Debug)]
+#[clap(
+    name = clap::crate_name!(),
+    author,
+    about,
+    version,
+)]
+#[deny(missing_docs)]
+pub struct Config {
+    /// The address of a console-enabled process to connect to.
+    ///
+    /// This may be an IP address and port, or a DNS name.
+    #[clap(default_value = "http://127.0.0.1:6669", value_hint = ValueHint::Url)]
+    pub(crate) target_addr: Uri,
+
+    /// Log level filter for the console's internal diagnostics.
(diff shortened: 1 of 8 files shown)


console/src/config.rs
```diff
@@ -0,0 +1,162 @@
...
+
+        // Does the terminal advertise truecolor support via the COLORTERM env var?
+        if self.truecolor {
+            tracing::debug!("millions of colors enabled via `COLORTERM=truecolor`");
+            return Palette::All;
+        }
+
+        // Okay, try to use `tput` to ask the terminfo database how many colors
+        // are supported...
+        let tput = Command::new("tput").arg("colors").output();
```
seanmonstar, line 140: Calling out to a new process worried me in what looked like an accessor method (`&self`), but I see it's only ever called once at start-up time, so it should be fine. Perhaps this method could consume `self`, but I don't have a strong opinion on it, just a musing.

seanmonstar, line 140: Separate point, it doesn't seem like `tputs` is available in Window CMD. I haven't checked if the console fully works in the console, but this might change that?

hawkw: Hmm, if the `tput` command fails, we'll just assume colors aren't supported. But, we could put the attempt to call out to `tput` behind `#[cfg(unix)]`, and not try to do it at all, if this won't work on Windows.

Do you know if there's a way to see if the console supports ANSI color escape codes on Windows? *Does* the windows console even support these escape codes?

seanmonstar: Oh I didn't notice that it handles `tput` failing. I guess that's fine. I know Windows terminals do support colors, though I believe through a different means. I remember `ripgrep` came up with a crate to make colors work on well on both... 🤷

seanmonstar (approved):

console/src/config.rs
hawkw: Okay, we should figure that out in a follow-up, I guess.

PR #93, opened 2021-08-20
feat(subscriber): accept durations with units

hawkw:
This branch updates the `console-subscriber` env var parsing code to accept
durations with units. Durations may now be given in any of `ns`, `us`,
`ms`, `s`, `sec`, `seconds`, `m`, `min`, `minutes`, `h`, or `hours`.
Durations specified in seconds or a larger unit may also be given as fractional values,
such as `1.5min`.

As a follow-up, we may also want to allow specifying durations as a
combination of multiple units, such as `3h15m`.

### DIFF
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -195,6 +195,7 @@ fn parse_duration(s: &str) -> Result<Duration, Box<dyn std::error::Error>> {
         .or_else(|| s.strip_suffix("minutes"))
     {
         return Ok(s
+            .trim()
             .parse::<u64>()
             .map(|mins| Duration::from_secs(mins * 60))
             .or_else(|_| {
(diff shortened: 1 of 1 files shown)


console-subscriber/src/builder.rs
```diff
@@ -158,3 +150,68 @@ impl Builder {
...
+        if let Some(s) = s.strip_suffix("us") {
+            return Ok(Duration::from_micros(s.trim().parse()?));
+        }
+
+        if let Some(s) = s.strip_suffix("ms") {
+            return Ok(Duration::from_millis(s.trim().parse()?));
+        }
+
+        if let Some(s) = s
+            .strip_suffix('s')
```
seanmonstar: Will this also match any units that the user ends with an `s`? Like `minutes`, `mins`, etc? Also, looking up, that'd mean `ns` would catch `2mins`?

hawkw: whoops, good catch --- this would then fail, because the rest of the string is not a valid integer/float. we should probably try to parse `s` last!

seanmonstar (approved):

PR #95, opened 2021-08-23
subscriber: remove trace event calls from the subscriber

### LINKED ISSUE #27: Potential deadlock when using `console_subscriber`
### Platform
x86_64 GNU/Linux

### Description
Spawning multiple instrumented futures along with the `console_subscriber` consistently results in the program hanging, with 0% CPU usage – looks like a deadlock. It could of course be that I'm doing something completely wrong!

We're spawning multiple tasks in the same location, with the following snippet:
```rust
#[track_caller]
fn spawn<W, G, F>(&mut self, g: G)
where
    W: Worker<Self>,
    G: FnOnce(oneshot::Receiver<()>) -> F,
    F: Future<Output = ()> + Send + 'static,
{
    let (tx, rx) = oneshot::channel();
    let future = g(rx);

    let caller = Location::caller();
    let span = tracing::info_span!(
        target: "tokio::task", 
        "task", 
        file = caller.file(), 
        line = caller.line(),
    );

    let task = tokio::spawn(future.instrument(span))

    self.tasks
        .entry(TypeId::of::<W>())
        .or_default()
        .push((tx, Box::new(task)));
}
```

The following code is used to create the subscriber. If the `TasksLayer` is never constructed, and `layer` never passed to the `tracing_subscriber`, the code runs as expected:

```rust
let (layer, server) = console_subscriber::TasksLayer::new();

let filter = tracing_subscriber::EnvFilter::from_default_env()
    .add_directive(tracing::Level::INFO.into())
    .add_directive("tokio=info".parse().unwrap());

tracing_subscriber::registry()
    .with(tracing_subscriber::fmt::layer())
    .with(filter)
    .with(layer)
    .init();

let serve = tokio::spawn(async move { server.serve().await });

// ----------------

let _ = tokio::try_join!(tokio__spawn(node.run()), serve);
```

seanmonstar:
It was noted that emitting a tracing event inside the subscriber could
trigger a deadlock, since `std::sync::Once` is not re-entrant. This
removes the tracing from the subscriber `Layer` methods.

If the capacity was full, sending the event was spawned into tokio, but
the event might not have been inside a runtime in the first place. And
it essentially turns it into an unbounded queue. So that is also
removed.

It also changes from using a constant `FLUSH_AT_CAPACITY` to a struct
field that is based on the configured event buffer capacity. A value of
half the buffer capacity was chosen, since hitting a full channel will
now mean the event is completely lost.

Fixes #27

### DIFF
--- a/console-subscriber/src/lib.rs
+++ b/console-subscriber/src/lib.rs
@@ -33,6 +33,9 @@ use crate::aggregator::TaskId;
 pub struct TasksLayer {
     tx: mpsc::Sender<Event>,
     flush: Arc<aggregator::Flush>,
+    /// When the channel capacity goes under this number, a flush in the aggregator
+    /// will be triggered.
+    flush_under_capacity: usize,
 
     /// Set of callsites for spans representing spawned tasks.
     ///
@@ -147,6 +150,10 @@ impl TasksLayer {
         let aggregator = Aggregator::new(events, rpcs, &config);
         let flush = aggregator.flush().clone();
 
+        // Conservatively, start to trigger a flush when half the channel is full.
+        // This tries to reduce the chance of losing events to a full channel.
+        let flush_under_capacity = config.event_buffer_capacity / 2;
+
         let server = Server {
             aggregator: Some(aggregator),
             addr: config.server_addr,
@@ -156,6 +163,7 @@ impl TasksLayer {
         let layer = Self {
             tx,
             flush,
+            flush_under_capacity,
             spawn_callsites: Callsites::default(),
             waker_callsites: Callsites::default(),
         };
@@ -170,11 +178,6 @@ impl TasksLayer {
 
     /// By default, completed spans are retained for one hour.
     pub const DEFAULT_RETENTION: Duration = Duration::from_secs(60 * 60);
-    // how much capacity should remain in the buffer before triggering a
-    // flush on capacity?
-    //
-    // chosen by fair die roll, guaranteed to be random :)
-    const FLUSH_AT_CAPACITY: usize = 100;
 
     fn is_spawn(&self, meta: &'static Metadata<'static>) -> bool {
         self.spawn_callsites.contains(meta)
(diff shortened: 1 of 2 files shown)


console-subscriber/src/lib.rs
```diff
@@ -201,27 +204,12 @@ impl TasksLayer {
...
-                tracing::warn!(
-                    "console buffer is full; some task stats may be delayed. \
-                     preemptive flush interval should be adjusted..."
-                );
-                let tx = self.tx.clone();
-                tokio::spawn(async move {
-                    if tx.send(event).await.is_err() {
-                        tracing::debug!("task event channel closed after lag");
-                    }
-                });
```
hawkw, line 214: Since we are now dropping events here, we should probably also ensure that the aggregator task will gracefully handle cases where it sees an event for a task with an ID it hasn't seen before. I think right now, it may record stats that are inaccurate, because it uses `update_or_default` on `TaskData` in a few places. Instead, we should probably ensure that if, for example, we never observed the creation of a task ID, we don't record any future timestamps for it (since they'll be wrong).

hawkw: also, we now don't really have a way to record cases where we missed events, which is probably worth doing (since it could indicate a console configuration error). 

perhaps we might want to have an `AtomicUsize` or something here that's incremented whenever we drop an event, and then in the aggregator task, we could log a warning when that counter is > 0? this wouldn't be a problem since the logging would occur in the aggregator task, outside of the subscriber. but, we can do that in a follow-up change!

hawkw (approved):
This looks good to me. I think we probably also want to track dropped events, but we can do that in a subsequent change. 

I think we should probably consider removing the use of `std::sync::Once` upstream in `tracing`, due to the re-entrancy issue. But it's good that this fixes the deadlock in the subscriber, so :+1:

hawkw:
Oh, I think rustfmt needs to be run?

console-subscriber/src/lib.rs
```diff
@@ -275,17 +263,11 @@ where
...
-                WakerVisitor {
-                    id: Some(id),
-                    op: Some(op),
-                } => {
-                    self.send(Event::Waker { id, op, at });
-                }
-                _ => {
-                    tracing::warn!("unknown waker event: {:?}", event);
-                }
+            if let WakerVisitor { id: Some(id), op: Some(op) } = visitor {
```
pnkfelix:
```suggestion
            if let WakerVisitor {
                id: Some(id),
                op: Some(op),
            } = visitor
            {
```

pnkfelix: this is the rustfmt change

seanmonstar: Wow, that's hideous 🙈🙃

hawkw:
Looks like there are also conflicts with `main` that need to be resolved?

PR #98, opened 2021-08-25
fix(console): don't crash if `COLORTERM`/`LANG` is not set

hawkw:
Currently, the console CLI/env var handling code _requires_ the
`COLORTERM` env variable. This isn't correct, as some terminals may not
set it. In that case, we should assume the terminal isn't advertising
24-bit color support, rather than crashing.

Currently, this happens if I unset the env variable:

```console
$ unset COLORTERM

$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.32s
     Running `target/debug/console`
error: The following required arguments were not provided:
    --colorterm <truecolor>

USAGE:
    console [FLAGS] [OPTIONS] --lang <LANG> --colorterm <truecolor> [TARGET_ADDR]

```

After this branch, if I unset `COLORTERM`, the console just runs without
colors.

Similarly, the `LANG` environment variable is no longer required, either.
If it's unset, we'll default to `en_us.UTF-8`. Possibly, we should *really*
be falling back to ASCII if `$LANG` is not set, but _most_ terminals really
ought to support UTF-8...

Finally, this fixes a bug in the way we parse `tput colors` output. The output
ends in a newline, but our parsing code doesn't handle whitespace...so we 
will currently always assume no colors are supported if we fall back to `tput`.
This branch adds a `trim` call.

### DIFF
--- a/console/src/config.rs
+++ b/console/src/config.rs
@@ -157,7 +157,7 @@ impl ViewOptions {
             return stdout
                 .map_err(|err| tracing::warn!(%err, "`tput colors` stdout was not utf-8 (this shouldn't happen)"))
                 .and_then(|s| {
-                    let palette = s.trim().parse::<Palette>();
+                    let palette = s.parse::<Palette>();
                     tracing::debug!(?palette, "parsed `tput colors`");
                     palette.map_err(|_| tracing::warn!(palette = ?s, "invalid color palette from `tput colors`"))
                 })
--- a/console/src/config.rs
+++ b/console/src/config.rs
@@ -55,7 +55,7 @@ pub struct ViewOptions {
         parse(from_str = parse_true_color),
         possible_values = &["24bit", "truecolor"],
     )]
-    truecolor: bool,
+    truecolor: Option<bool>,
 
     /// Explicitly set which color palette to use.
     #[clap(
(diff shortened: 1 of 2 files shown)


seanmonstar:
Just an anecdote from my environment, even though the variable isn't set, manually passing `--colorterm truecolor` works just fine, and `tput colors` returns 256. If the env var is missing, could we default to some sort of "auto"?

hawkw:
>  If the env var is missing, could we default to some sort of "auto"?

If the env var is missing, we will check `tput colors` next (which is essentially "auto"). There may be more ways to detect color support that we should also try --- we could potentially also look at `$TERM` and see if the terminal program is in a list of terminals we know support colors regardless of what `tput` says.

I'm a little hesitant to default to emitting color escapes if we don't have _some_ way of positively determining that they'll work --- I don't want to give users whose terminals don't support colors, or don't support 24-bit millions of colors, a big pile of gibberish by default.

hawkw:
Also, I've updated this branch to also handle the `$LANG` variable not being set.

seanmonstar:
Oh yea, by "auto" I meant check `tput` or something. Sounds good!

seanmonstar (approved):

PR #99, opened 2021-08-27
subscriber: support runtime:: target for tracing events

zaharidichev:
Currently the subscriber has an `EnvFilter` that filters out anything that is not
part of the `tokio::` target. However the convention for targets is `runtime::`. This
PR adds the needed directive to the `EnvFilter`. 


hawkw (approved):
thanks, looks good to me!

PR #102, opened 2021-08-27
console: prevent huge take names from panicking

seanmonstar:
If a task name were too long, calculating the remaining width for the
fields column would panic on a subtraction overflow. This removes the
need to calculate the remainder, by just setting a percentage constraint
instead.

This also caps the `view::Width` so that an excessively long string
doesn't overflow the stored `u16`.

### DIFF
--- a/console/src/view/tasks.rs
+++ b/console/src/view/tasks.rs
@@ -204,6 +204,7 @@ impl List {
             Span::from(format!(" Idle ({})", num_idle)),
         ]);
 
+        /* TODO: use this to adjust the max size of name and target columns...
         // How many characters wide are the fixed-length non-field columns?
         let fixed_col_width = id_width.chars()
             + STATE_LEN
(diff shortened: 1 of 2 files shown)


console/src/view/tasks.rs
```diff
@@ -204,6 +204,7 @@ impl List {
             Span::from(format!(" Idle ({})", num_idle)),
         ]);
 
+        /* TODO: use this to adjust the max size of name and target columns...
```
seanmonstar, line 207: I want to use this to calculate a max relative size for the name and target columns, but I'm not sure how exactly I'd want to work when there isn't even room for the full fixed-width columns. So for now, I figured I'd just fix these things.

console/src/view/tasks.rs
```diff
@@ -213,11 +214,15 @@ impl List {
...
-        // TODO(eliza): there's gotta be a nicer way to do this in `tui`...what
-        // we want is really just a constraint that says "always use all the
-        // characters remaining".
-        let fields_width = frame.size().width - fixed_col_width;
+        //
+        // Ideally we'd use Min(0), and it would fill the rest of the space. But that is broken
+        // in tui 0.16. We can use Percentage to fill the space for now.
+        //
+        // See https://github.com/fdehau/tui-rs/issues/525
+        let fields_width = layout::Constraint::Percentage(100);
```
hawkw, line 225: oh, that's much easier than what i was doing, props for figuring that out...

console/src/view/tasks.rs
hawkw: Hmm, what did you have in mind, exactly? Might be worth a follow-up issue for the TODO?

hawkw (approved):
looks good to me, thanks for fixing that!

console/src/view/tasks.rs
seanmonstar: Well for instance, in this image, it'd make sense to cap the length of the name, maybe with an ellipsis, so you can see the other columns too. 🤷 

console-long

PR #112, opened 2021-09-05
feat: emit and show self-wake counts for tasks

hawkw:
This counts all wakes of a task that happen while the task is "currently
polled". This could be from `task::yield_now()`, from `tokio::coop`, or
any other pattern that triggers the task to wake even though it's
currently being polled.

This implementation builds upon @seanmonstar's branch #55. However,
rather than detecting self-wakes by simply looking at _whether_ the task
is being polled at all, we instead detect them by checking whether the
task's span is on the current thread's stack of entered spans. This
approach should _not_ produce self-positives when a task is woken by
another thread while it is currently entered (as discussed in
https://github.com/tokio-rs/console/pull/55#discussion_r659223228).

Testing this with the "burn" subcommand for the example app, which
spawns one task that does a lot of self-wakes via `task::yield`,
indicates that it is, in fact, working properly. Here are the stats for the
`burn` task:
image
Note the very large number of self-wakes.

In comparison, here are the task stats for a "normal" task, which does
not wake itself (and is instead woken by timers):
image
Note that no self-wakes stat is displayed.

Closes #55.

### DIFF
--- a/console-subscriber/src/aggregator/mod.rs
+++ b/console-subscriber/src/aggregator/mod.rs
@@ -184,6 +184,7 @@ struct TaskStats {
     wakes: u64,
     waker_clones: u64,
     waker_drops: u64,
+    self_wakes: u64,
     last_wake: Option<SystemTime>,
 
     poll_times_histogram: Histogram<u64>,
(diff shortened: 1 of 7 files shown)


zaharidichev (approved):
LGTM

seanmonstar (approved):

PR #113, opened 2021-09-06
feat(console): add self-wake percentage and warning

hawkw:
Depends on #112

This branch builds on #112 by calculating the *percentage* of a task's
wakeups that were self-wakes, and displaying them in the tasks view. It
also adds the beginnings of a rudimentary "Warnings" system, and
displays warnings for any tasks who have woken themselves for more than
50% of their wakeups. There is a new `Warning` trait for generating
warnings that can be used to add new warnings in the future.

The warnings functionality can almost certainly be expanded --- for
example, it would be nice to be able to quicky jump to the details view
for a task that has warnings. In the future, I could imagine generating web
documentation with details about a particular warning and how to fix it,
like [clippy has][1], and linking to them from the console. But, this is a start,
at least!

[1]: https://rust-lang.github.io/rust-clippy/v0.0.174/
image
image

### DIFF
--- a/console/src/util.rs
+++ b/console/src/util.rs
@@ -2,18 +2,18 @@ pub(crate) trait Percentage {
     // Using an extension trait for this is maybe a bit excessive, but making it
     // a method has the nice advantage of making it *really* obvious which is
     // the total and which is the amount.
-    fn percent_of(self, total: Self) -> f64;
+    fn percent_of(self, total: Self) -> Self;
 }
 
 impl Percentage for usize {
-    fn percent_of(self, total: Self) -> f64 {
-        percentage(total as f64, self as f64)
+    fn percent_of(self, total: Self) -> Self {
+        percentage(total as f64, self as f64) as Self
     }
 }
 
 impl Percentage for u64 {
-    fn percent_of(self, total: Self) -> f64 {
-        percentage(total as f64, self as f64)
+    fn percent_of(self, total: Self) -> Self {
+        percentage(total as f64, self as f64) as Self
     }
 }
 
--- a/console/src/util.rs
+++ b/console/src/util.rs
@@ -1,0 +1,28 @@
+pub(crate) trait Percentage {
+    // Using an extension trait for this is maybe a bit excessive, but making it
+    // a method has the nice advantage of making it *really* obvious which is
+    // the total and which is the amount.
+    fn percent_of(self, total: Self) -> f64;
+}
+
+impl Percentage for usize {
+    fn percent_of(self, total: Self) -> f64 {
+        percentage(total as f64, self as f64)
+    }
+}
+
+impl Percentage for u64 {
+    fn percent_of(self, total: Self) -> f64 {
+        percentage(total as f64, self as f64)
+    }
+}
+
+pub(crate) fn percentage(total: f64, amount: f64) -> f64 {
+    debug_assert!(
+        total >= amount,
+        "assertion failed: total >= amount; total={}, amount={}",
+        total,
+        amount
+    );
+    (amount / total) * 100.0
+}
--- a/console/src/view/tasks.rs
+++ b/console/src/view/tasks.rs
@@ -14,7 +14,9 @@ use tui::{
 
 #[derive(Debug)]
 pub(crate) struct List {
-    warnings: Vec<Box<dyn warnings::Warning<Task>>>,
+    /// A list of linters (implementing the [`Warn`] trait) used to generate
+    /// warnings.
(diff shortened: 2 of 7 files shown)


console/src/util.rs
```diff
@@ -0,0 +1,28 @@
+pub(crate) trait Percentage {
+    // Using an extension trait for this is maybe a bit excessive, but making it
+    // a method has the nice advantage of making it *really* obvious which is
+    // the total and which is the amount.
+    fn percent_of(self, total: Self) -> f64;
```
seanmonstar: Does this return the percent out of 100.0, or out of 1.0? Many things assume when it's a float, that it's out of 1.0. It looks like you're making it out of 100 down below. Is there reason to hold onto the float then? Maybe 99.99 percent is nicer to see than 99?

console/src/view/tasks.rs
```diff
@@ -260,6 +295,14 @@ impl List {
 
         frame.render_widget(Paragraph::new(controls), controls_area);
 
+        if let Some(area) = warnings_area {
+            let block = styles.border_block().title(Spans::from(vec![
+                bold("Warnings"),
```
seanmonstar, line 302: I quite like how this highlights warnings at a quick glance. This let's a user know what the warning is more quickly. I do wonder if a lot of tasks trigger a warning, if this extra table will be too big...

I had originally thought of adding another column to the regular table, that would show a ⚠️ in a row if the task had any. Then you could hit enter and see the task detail view to read about what warnings there were. Maybe that will help when there are a lot of warnings.

console/src/warnings.rs
```diff
@@ -0,0 +1,37 @@
+use crate::tasks::Task;
+
+pub trait Warning<T>: std::fmt::Debug {
```
seanmonstar: This seems fine to start with. Maybe we discover we need something different when trying to add new warnings, but no point in trying to imagine.

console/src/view/tasks.rs
```diff
@@ -10,8 +11,10 @@ use tui::{
     text::{self, Span, Spans},
     widgets::{Cell, Paragraph, Row, Table, TableState},
 };
-#[derive(Clone, Debug, Default)]
+
+#[derive(Debug)]
 pub(crate) struct List {
+    warnings: Vec<Box<dyn warnings::Warning<Task>>>,
```
seanmonstar: Ah, after reading through the file, I realized I misunderstood this field. I thought this was all the warnings that had been found in the tasks, but it's the list of warnings to check for. Maybe a field name of `lints`, or a doc comment?

seanmonstar (approved):
This is fantastic! I'm so excited! I left some thoughts inline, but really just some rambling.

console/src/util.rs
hawkw: we could cast the float back to a usize before returning it. i doubt fractional percentages are terribly valuable here...

console/src/warnings.rs
hawkw: yeah, i figured the trait can be pretty minimal to start and we can add stuff as we add more warnings. one thing i'd really like to have in the eventual future is some way of generating structured web-based documentation for the different categories of warnings/lints, explaining why they might be a problem, how to fix them, etc, like [what clippy has](https://rust-lang.github.io/rust-clippy/master/index.html#). then, we can have a URL and/or an lint number users can use to get more information. but, i think adding all of that infrastructure is a fairly big project...perhaps we can crib some of it from clippy?

similarly, it could eventually be nice to have categories of warnings/lints, and allow users to turn some of them on or off. but that's not important unless we have a _big_ list of warnings.

console/src/view/tasks.rs
hawkw: hmm, that's a good point that there potentially might be a large number of warnings. we should probably also add the :warning: column to the tasks table (this would also let you sort by that column so that tasks with warnings are at the top of the list!). we could _also_ show warnings at the top like this, though...we could maybe have a maximum number of lines to use for it (say, 5), and if there's more than that, we could just show a message that says ":warning: 385 tasks have warnings! See the task details page for those tasks for more information." or something.

hawkw: eventually, when there are different classes of warnings, it would be cool to show a summary of each warning kind that was found and a count, if there are a lot of warnings. like
```
⚠ 201 tasks have woken themselves more than 50% of the time!
⚠ 3 tasks have never yielded for over 300 ms! Are those tasks blocking?
⚠ 519 tasks have been idle for more than 90% of their lifetime. Is the runtime over-saturated?
```

hawkw: (i made these up off the top of my head)

PR #114, opened 2021-09-06
subscriber: look at event parents to determine resource id

zaharidichev:
This PR changes the way poll and state update ops are associated
with resource by incorporating the feedback in https://github.com/tokio-rs/tokio/pull/4072#discussion_r699479788
Instead of only looking at entered spans, we now use `Context::event_span`


hawkw (approved):
looks good to me, thanks!

Issue #109, opened 2021-09-02
Name or classify tasks spawned by the subscriber runtime


seanmonstar:
The subscriber starts its own runtime and server, and those tasks are instrumented and shown in the Console too. That is actually useful in some situations. But it also can be distracting. It would help if we could name or classify any tasks in that runtime to help a user viewing the Console, so they understand "oh, those aren't my tasks".

hawkw:
Since using the console already requires enabling `tokio_unstable`, we can probably just use `task::Builder` to add a name, anywhere we spawn a task.

In the future, it would be nice to be able to associate tasks with runtimes, and name runtimes as well. That way, we could say things like "hide all tasks belonging to the console runtime". But that would require new instrumentation in Tokio, so it's a bigger project.

hawkw:
#117 added names to the console's `aggregate` and `serve` tasks. However, we can't easily name the per-connection tasks spawned by Tonic without switching to tonic's lower-level API and writing our own accept loop to spawn those tasks on our own. We'll do that in a subsequent branch (or consider changes upstream?)

g2p:
Could the many tonic / hyper tasks be filtered out using spans and layers?
I don't know enough about tracing to know if logic like metadata.is_within_span is implementable.
Edit: something like a dynamic_filter_fn?
When debugging a hyper-based application tonic adds a lot of noise, ways to hide or at least distinguish would be welcome.

hawkw:
@g2p it's definitely possible to use `dynamic_filter_fn` to filter out those spans at the subscriber-level, but I think the best long-term solution is to add filtering to the console UI: https://github.com/tokio-rs/console/issues/131

PR #117, opened 2021-09-06
feat(subscriber): name tasks spawned by the console subscriber

hawkw:
This branch updates the `console-subscriber` crate so that the
"aggregate" and "serve" tasks spawned by the console are spawned with
names. This allows the user to distinguish between tasks spawned by the
console subscriber and tasks spawned by other parts of the application.

Note that this does *not* include the tasks spawned by Tonic to handle
each accepted client connection. Those tasks are spawned inside of
Tonic, rather than by the console-subscriber crate, so we can't easily
add our own names to them. We could probably fix that by using Tonic's
lower level APIs to spawn our own per-connection tasks, and give them
names...but that seems like a follow-up PR.

This is a first pass on #109 (although it isn't a *complete* solution
due to the above issue with Tonic).

Screenshot: 
image

### DIFF
--- a/console-subscriber/src/lib.rs
+++ b/console-subscriber/src/lib.rs
@@ -541,16 +541,16 @@ impl Server {
             .aggregator
             .take()
             .expect("cannot start server multiple times");
-        let aggregate = tokio::spawn(aggregate.run());
+        let aggregate = spawn_named(aggregate.run(), "console::aggregate");
         let addr = self.addr;
-        let res = builder
+        let serve = builder
             .add_service(proto::instrument::instrument_server::InstrumentServer::new(
                 self,
             ))
-            .serve(addr)
-            .await;
+            .serve(addr);
+        let res = spawn_named(serve, "console::serve").await;
         aggregate.abort();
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):

console-subscriber/src/lib.rs
```diff
@@ -643,3 +643,18 @@ impl WakeOp {
...
+
+#[track_caller]
+pub(crate) fn spawn_named<T>(
+    task: impl std::future::Future<Output = T> + Send + 'static,
+    _name: &str,
+) -> tokio::task::JoinHandle<T>
+where
+    T: Send + 'static,
+{
+    #[cfg(tokio_unstable)]
```
seanmonstar, line 662: It looks like the builder already asserts this is set, but this doesn't hurt.

hawkw: yeah...it's possible people are using the crate without the builder. We may want to remove the assertion from the builder, in the case where the console is used with non-tokio runtimes like Rayon...

PR #119, opened 2021-09-07
console: add `retain-for` cmd line arg

### LINKED ISSUE #108: Consider option to disable expiring of stopped tasks
In applications with short-lived tasks, it may be desirable to always see all finished tasks. Or at least be able to specify as such when running the `console` app. (Does this seem like a feature anyone else would want?)

To do this, we would add another command line option to the console in `console/src/config.rs`, set the value in `console/src/tasks.rs`, and check it before calling `retain`.

zaharidichev:
This PR adds a `retain-for` duration argument to the console. Additionally, 
I have modified the proto to not include the total_time for tasks,resources and
async ops and instead contain `created_at` and `dropped_at` fields. A separate 
`console-util` crate has been added that contains the logic for parsing a duration
string. 

close #108


console/src/tasks.rs
```diff
@@ -249,14 +252,20 @@ impl State {
...
 
-        self.tasks.retain(|_, task| {
-            let mut task = task.borrow_mut();
-            if task.completed_for == 0 {
-                return true;
-            }
-            task.completed_for += 1;
-            task.completed_for <= Self::RETAIN_COMPLETED_FOR
-        })
+        if let Some(now) = self.last_updated_at {
```
zaharidichev: Not sure whether this is a problem but this is not entirely accurate. The problem is that we calculate the `dopped_for` number based on the time of the last update.  So if no update is coming for a while we might see tasks sticking around longer than the configured `retain_for`

console-util/Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[package]
+name = "console-util"
+version = "0.1.0"
+authors = ["Eliza Weisman <eliza@buoyant.io>"]
+edition = "2018"
```
hawkw: hmm, i'm not the biggest fan of creating a separate `console-util` crate just for the duration-parsing code...but I agree that it doesn't make sense in `console-api`. i think it might be better if it went in a crate specifically for duration parsing that we could publish to crates.io eventually.

alternatively, we might just want to replace our homemade duration parser with an existing library, especially if an existing library would accept more durations we don't handle nicely at the moment. i would probably go with [the `humantime` crate](https://lib.rs/crates/humantime) for this. i believe [`chrono`](https://lib.rs/crates/chrono) can also do this, but i think `humantime` would be a better choice for us, since `chrono` includes a lot of code for working with _dates_, which we don't need. `humantime` (45KB / 917 SLoC) is a significantly smaller dependency than `chrono` (760KB / 11k SLoC), and _only_ provides code for parsing/formatting durations. i think maybe we should switch to `humantime`.

console-subscriber/src/aggregator/mod.rs
```diff
@@ -105,12 +105,12 @@ pub(crate) struct Flush {
...
-// in use anymore
-pub(crate) trait Closable {
-    fn closed_at(&self) -> Option<SystemTime>;
+// An entity (e.g Task, Resource) that at some point in
+// time can be dropped. This generally refers to spans that
+// have been closed indicating that a task, async op or a
+// resource is not in use anymore
+pub(crate) trait Droppable {
+    fn dropped_at(&self) -> Option<SystemTime>;
 }
```
hawkw: What's the motivation for renaming this? I'm not a _big_ fan of having a trait called `Droppable`, since it seems like it could be easy to confuse with `Drop`. `Droppable` doesn't really indicate that something _can_ be dropped, since _everything_ can be dropped...instead, it indicates that something _has a dropped timestamp_. Maybe we should call this something like `DroppedTime` or `DroppedAt`?

console/src/config.rs
```diff
@@ -28,6 +29,11 @@ pub struct Config {
 
     #[clap(flatten)]
     pub(crate) view_options: ViewOptions,
+
+    /// The amount of time resource that have been dropped
+    /// will be displayed in the console.
```
hawkw: how about something like this?
```suggestion
    /// How long to continue displaying completed tasks and dropped resources
    /// after they have been closed.
```

hawkw: we may want to also add a note about how durations can be specified so the user knows what kinds of input will be accepted. this, of course, will depend on whether we continue using our homemade duration parsing code or switch to the `humantime` crate...

console/src/tasks.rs
hawkw: I think that's fine, actually. We should expect updates to come pretty frequently (about once every second by default). In the previous implementation, we were only removing completed tasks after a particular number of updates had passed, so tasks would only be removed if we are continually receiving updates, and I think that's actually good --- if the connection temporarily goes down and we have to reconnect, I think it's nicer to keep showing whatever state we have until we are able to get back in touch with the app...

console/src/config.rs
```diff
@@ -28,6 +29,11 @@ pub struct Config {
...
+    #[clap(long = "retain-for", default_value = "5h", parse(try_from_str = parse_retain_for))]
```
hawkw: i think the default value here basically  means that completed tasks will never be removed (for all practical purposes), which is a significant change from the current behavior. i think in order to get something closer to what the console does on `main`, we should probably set a much smaller default value, like
```suggestion
    #[clap(long = "retain-for", default_value = "6s", parse(try_from_str = parse_retain_for))]
```
(since we previously retained completed tasks for 6 updates, and updates default to coming every second, this should be _about_ the same as what we were doing before).

hawkw: another thought, we may want to have some way to _totally disable_ expiring completed tasks? in most cases you could just set a very long duration, but if i want to like, leave the console running for several hours to monitor some program while i'm doing other work, i might be surprised if tasks suddenly start disappearing.

we could either add another flag to disable expiring completed completely (which would conflict with this flag), or we could accept `--retain-for=none` or something? i'm not sure which would be a better user experience for the CLI...

hawkw (commented):
thanks for working on this. i had some thoughts.

i think at this point, we might be better off replacing the homemade duration-parsing code with the [`humantime`](https://lib.rs/crates/humantime) crate, which is a little more flexible (it can accept mixed durations like "2mins 30secs") and also has [much more helpful error messages](https://github.com/tailhook/humantime/blob/c6a23911677ebaf4f14dab8bca1120b7c0af2be1/src/duration.rs#L53-L77). it's a fairly lightweight dependency, so i think it might be a better choice than making a new crate for what's essentially a worse version of `humantime` that i wrote in an afternoon. :)

besides that, this mostly looks good --- i had some naming nits, and some potential additional CLI changes we might want to make. let me know what you think?

console-subscriber/src/aggregator/mod.rs
zaharidichev: The main motivation was that I renamed the proto field (where it already existed) from `closed_at` to `dropped_at`. But yes, that all make sense. I will rename it.

console/src/config.rs
zaharidichev: Yes, I left that by mistake. Was testing it with 5 hours to ensure tasks are not cleared. having a `--retain-for=none` works for me

console/src/config.rs
```diff
@@ -28,7 +29,17 @@ pub struct Config {
...
     #[clap(flatten)]
     pub(crate) view_options: ViewOptions,
+
+    /// How long to continue displaying completed tasks and dropped resources
+    /// after they have been closed. Accepted values are:
+    ///
+    /// - durations in the form of 5days 2min 2s.
+    /// - `none` to disable this functionality
+    #[clap(long = "retain-for", default_value = "6s", parse(try_from_str = parse_retain_for))]
+    retain_for: RetainFor,
```
zaharidichev: FWIW, having an `Option<T>` conflicts with `default_value`, therefore the wrapper.

Cargo.toml
```diff
@@ -8,3 +8,4 @@ resolver = "2"
 
 [patch.crates-io]
 tokio = { git = 'https://github.com/zaharidichev/tokio', branch = 'zd/instrument-sleep' }
+
```
hawkw: extra line of whitespace here?
```suggestion
tokio = { git = 'https://github.com/zaharidichev/tokio', branch = 'zd/instrument-sleep' }
```

console-subscriber/src/aggregator/mod.rs
```diff
@@ -105,12 +105,12 @@ pub(crate) struct Flush {
...
-// indicating that a task, async op or a resource is not
-// in use anymore
-pub(crate) trait Closable {
-    fn closed_at(&self) -> Option<SystemTime>;
+// An entity (e.g Task, Resource) that at some point in
+// time can be dropped. This generally refers to spans that
+// have been closed indicating that a task, async op or a
+// resource is not in use anymore
+pub(crate) trait DroppedAt {
+    fn dropped_at(&self) -> Option<SystemTime>;
```
hawkw, line 113: Thanks, this is better than `Droppable` IMO.

I'm still not convinced that renaming this fields from `closed_at` to `dropped_at` is really worth such a large diff, but I'm fine with it if you think "dropped" is clearer or more accurate than "closed"!

console/src/config.rs
```diff
@@ -28,7 +29,17 @@ pub struct Config {
...
 }
+#[derive(Debug)]
+struct RetainFor(Option<Duration>);
```
hawkw: nit: add a line of whitespace between struct definitions please :)
```suggestion
}

#[derive(Debug)]
struct RetainFor(Option<Duration>);
```

console/src/config.rs
```diff
@@ -28,7 +29,17 @@ pub struct Config {
 
     #[clap(flatten)]
     pub(crate) view_options: ViewOptions,
+
+    /// How long to continue displaying completed tasks and dropped resources
+    /// after they have been closed. Accepted values are:
+    ///
+    /// - durations in the form of 5days 2min 2s.
+    /// - `none` to disable this functionality
```
hawkw:
```suggestion
i wonder if it's worth explicitly saying something like "
```suggestion
    /// - Durations in the form of `5days 2min 2s` and similar. Durations are parsed using
    ///    the [`humantime` crate's `parse_duration` function][1]. See that crate's documentation
    ///    for details.
    /// - `none` to disable this functionality
    /// [1]: https://docs.rs/humantime/2.1.0/humantime/fn.parse_duration.html
```
or, alternatively, we could just coy from their docs:
```suggestion
    /// - Durations, parsed as a combination of time spans (such as `5days 2min 2s`). 
    ///    
    ///   Each time span is an integer number followed by a suffix. Supported suffixes are:
    ///
    ///   * `nsec`, `ns` -- nanoseconds
    ///   * `usec`, `us` -- microseconds
    ///   * `msec`, `ms` -- milliseconds
    ///   * `seconds`, `second`, `sec`, `s`
    ///   * `minutes`, `minute`, `min`, `m`
    ///   * `hours`, `hour`, `hr`, `h`
    ///   * `days`, `day`, `d`
    ///   * `weeks`, `week`, `w`
    ///   * `months`, `month`, `M` -- defined as 30.44 days
    ///   * `years`, `year`, `y` -- defined as 365.25 days
    /// - `none` to disable removing completed task spans
```

(i don't know how `clap`'s derive macro handles markdown links in RustDoc, so if we want to link to their docs, we may be better off just including the actual URL)

console/src/config.rs
```diff
@@ -176,3 +191,13 @@ fn parse_true_color(s: &str) -> bool {
...
+
+fn parse_retain_for(s: &str) -> Result<RetainFor, String> {
+    match s {
+        "none" => Ok(RetainFor(None)),
+        _ => match s.parse::<humantime::Duration>() {
+            Ok(duration) => Ok(RetainFor(Some(duration.into()))),
+            Err(error) => Err(error.to_string()),
+        },
+    }
+}
```
hawkw: nit/TIOLI: this could be an implementation of the `FromStr` trait for `RetainFor`, since `RetainFor` is a type we define in this crate. then, we wouldn't need the `parse(try_from_str = parse_retain_for))` in the `clap` attribute.

not a blocker

console/src/config.rs
```diff
@@ -176,3 +191,13 @@ fn parse_true_color(s: &str) -> bool {
     let s = s.trim();
     s.eq_ignore_ascii_case("truecolor") || s.eq_ignore_ascii_case("24bit")
 }
+
+fn parse_retain_for(s: &str) -> Result<RetainFor, String> {
+    match s {
+        "none" => Ok(RetainFor(None)),
+        _ => match s.parse::<humantime::Duration>() {
+            Ok(duration) => Ok(RetainFor(Some(duration.into()))),
+            Err(error) => Err(error.to_string()),
```
hawkw: nit: i would probably change this to just return the `humantime` error:
```suggestion
fn parse_retain_for(s: &str) -> Result<RetainFor, humantime::Error> {
    match s {
        "none" => Ok(RetainFor(None)),
        _ => s
            .parse::<humantime::Duration>()
            .map(|duration| RetainFor(Some(duration.into()))),
```
or something?

hawkw (approved):
looks good to me overall! couple last minor nits, but take it or leave it.

console/src/config.rs
```diff
@@ -176,3 +206,16 @@ fn parse_true_color(s: &str) -> bool {
     let s = s.trim();
     s.eq_ignore_ascii_case("truecolor") || s.eq_ignore_ascii_case("24bit")
 }
+
+impl FromStr for RetainFor {
+    type Err = humantime::DurationError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "none" => Ok(RetainFor(None)),
```
hawkw: i think we may want to do this instead:
```suggestion
            s if s.eq_ignore_ascii_case("none") => Ok(RetainFor(None)),
```
so that the user can also write "None"...

hawkw (approved):
looks good to me! thanks for working on this!

PR #123, opened 2021-09-08
feat(console): display warnings in task details *and* list

hawkw:
The current method of displaying warnings is as list of all individual
tasks with warnings on the task list screen. As discussed in the
comments on PR #113, this may not be the ideal approach when there are a
very large number of tasks with warnings.

This branch changes this behavior so that the details of a particular
warning for a given task is shown only on the task details screen. On
the task list screen, we instead list the different _categories_ of
warnings that were detected, along with the number of tasks with that
warning. This means that when a large number of tasks have warnings, we
will only use a number of lines equal to the number of different
categories of warning that were detected, rather than the number of
individual tasks that have that warning.

Each individual task that has warnings shows a warning icon and count in
a new column in the task list table. This makes it easy for the user to
find the tasks that have warnings and get details on them, including
sorting by the number of warnings.

Implementing this required some refactoring of how warnings are
implemented. This includes:

* Changing the `Warn` trait to have separate methods for detecting a
  warning, formatting the warning for an individual task instance, and
  summarizing the warning for the list of detected warning types
* Introducing a new `Linter` struct that wraps a `dyn Warning` in an
  `Rc` and clones it into tasks that have lints. This allows the task
  details screen to determine how to format the lint when it is needed.
  It also allows tracking the total number of tasks that have a given
  warning, by using the `Rc`'s reference count.
## Screenshots

To demonstrate how this design saves screen real estate when there are
many tasks with warnings, I modified the `burn` example to spawn
several burn tasks rather than just one.

Before, we spent several lines on warnings (one for each task):
warn_old

After, we only need one line:
warnings1

The detailed warning text for the individual tasks are displayed on the
task details view:
warnings1_details

And, it still looks okay in ASCII-only mode:
warnings1_ascii-only
warnings1_details_ascii-only

### DIFF
--- a/console/src/main.rs
+++ b/console/src/main.rs
@@ -44,7 +44,12 @@ async fn main() -> color_eyre::Result<()> {
     // A channel to send the task details update stream (no need to keep outdated details in the memory)
     let (details_tx, mut details_rx) = mpsc::channel::<TaskDetails>(2);
 
-    let mut tasks = State::default();
+    let mut tasks = State::default()
+        // TODO(eliza): allow configuring the list of linters via the
+        // CLI/possibly a config file?
+        .with_linters(vec![warnings::Linter::new(
+            warnings::SelfWakePercent::default(),
+        )]);
     let mut input = input::EventStream::new();
     let mut view = view::View::new(styles);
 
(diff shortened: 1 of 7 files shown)


seanmonstar (approved):
Looks beautiful!

PR #125, opened 2021-09-08
console: spell out ASCII state with BUSY/IDLE/DONE

seanmonstar:


hawkw:
looks like this needs rustfmt?

hawkw (approved):
cool yeah this is definitely better than my randomly-chosen incomprehensible ascii characters :+1:

PR #126, opened 2021-09-09
console: lint new tasks, not just when there is a stats update

seanmonstar:
I was working on a lost waker lint, and noticed that the "coma" task wasn't being linted, because it never sends another stats update. This fixes it to lint tasks when we first get them, too.


hawkw (approved):
whoops, good catch!

PR #127, opened 2021-09-09
console: add LostWaker lint

seanmonstar:
This lint will warn when a task that isn't completed has no more wakers,
implying the task will never wake up again.

Note that this depends on #126 in most cases.

image


hawkw (approved):
Testing this out locally with the `coma` example, I can confirm that this does indeed detect the task that is never woken:
image

Very cool!

PR #128, opened 2021-09-09
Make proto/ vendor-able 

guswynn:
I work with a large monorepo that `cargo vendor'`s all third-party code. We use tokio quite a bit, and want to experiment with `tokio/console`

When vendoring `console-api` the `../proto` directory is lost, and the crate is unbuildable. I understand that keeping `proto/` as a top-level directory is probably desirable, so this change symlinks that directory into that sub-crate itself.

My understanding is that this change may also be required once `console` is put on crates.io?

`cargo vendor` before this change:
```
Cargo.toml build.rs   src
```
after:
```
Cargo.toml
build.rs
proto
src
```

This is basically copying what https://github.com/dtolnay/cxx does with symlinks


hawkw:
Thanks, it hadn't occurred to me that this would be a potential issue for users who are vendoring.

Re:
>  I understand that keeping `proto/` as a top-level directory is probably desirable, so this change symlinks that directory into that sub-crate itself.

FWIW, I think it would probably be fine if we _did_ move `proto/` into `console-api`, since I would generally expect most (or all?) Rust code to consume the protobufs via that crate rather than directly, and I am not sure if it's likely that they'll be used outside of Rust any time in the near future. But, the symlink is fine, too, just wanted to put it out there that I don't think not moving the proto definitions is a hard blocker.

hawkw (approved):

guswynn:
@hawkw did some random commits accidentally get added to this PR?

hawkw:
> @hawkw did some random commits accidentally get added to this PR?

We have GitHub configured to require that branches are up to date before merging them (to ensure CI runs with the branch integrated with the latest changes). Therefore, I clicked the "Update this branch with the latest changes" button to prepare to merge it.

spacemeowx2:
Hi, I found that this PR makes the console not compile in Windows and [`cross`](https://github.com/rust-embedded/cross). I think it's because of the symbolic links.

Error message is like:
```
Caused by:
  process didn't exit successfully: `/target/release/build/console-api-a01a908c13fc7610/build-script-build` (exit status: 1)
  --- stderr
  Error: Custom { kind: Other, error: "protoc failed: Could not make proto path relative: proto/trace.proto: No such file or directory\n" }
```

hawkw:
> Hi, I found that this PR makes the console not compile in Windows and [`cross`](https://github.com/rust-embedded/cross). I think it's because of the symbolic links.
> 
> Error message is like:
> 
> ```
> Caused by:
>   process didn't exit successfully: `/target/release/build/console-api-a01a908c13fc7610/build-script-build` (exit status: 1)
>   --- stderr
>   Error: Custom { kind: Other, error: "protoc failed: Could not make proto path relative: proto/trace.proto: No such file or directory\n" }
> ```

Yeah, other users have reported that as well. PR #141 _should_ fix this...but unfortunately, I don't have easy access to a Windows machine to test that on. If you're interested in checking out that branch and confirming that it does fix the build, that would be lovely!

guswynn:
@spacemeowx2 I don't have access to a windows machine, can you try building doing this? https://github.com/dtolnay/cxx/blob/master/.github/workflows/ci.yml#L36

in the meantime, I think @hawkw is right, #141 will simplify this

Sorry about the noise!

spacemeowx2:
@guswynn Unfortunately it doesn't work for me:

```
error: unable to create symlink console-api/proto: Permission denied
```

@hawkw This PR fixed my problem.

Thanks!

PR #135, opened 2021-09-11
chore: don't ignore Cargo.lock

90-008:
This PR removes the .gitignore entry for `Cargo.lock`. This is useful for external tools such as Nix to build console without generating a lock file locally and having to patch the source.


hawkw (approved):
thanks, this looks good to me.

one question: since this workspace contains both a binary application _and_ libraries, do you know if this is the recommended approach for such mixed workspaces? is it possible to configure the CI build so that the library crates are built without the lockfile (to ensure that they get the latest dependency versions that meet their constraints)?  if there was a way to check in a lockfile only for the binary crate, that seems like it would be the ideal solution...

regardless, i'm happy to merge this as-is and keep iterating on those questions in follow-up branches, but if you happen to know off the top of your head what the recommended approach is, that would be nice...

hawkw:
hmm, this actually breaks CI. it looks like we have a `tracing-subscriber` version constraint that is too old to ensure that APIs we use are present...let me fix that!

90-008:
> thanks, this looks good to me.
> 
> one question: since this workspace contains both a binary application _and_ libraries, do you know if this is the recommended approach for such mixed workspaces? is it possible to configure the CI build so that the library crates are built without the lockfile (to ensure that they get the latest dependency versions that meet their constraints)? if there was a way to check in a lockfile only for the binary crate, that seems like it would be the ideal solution...
> 
> regardless, i'm happy to merge this as-is and keep iterating on those questions in follow-up branches, but if you happen to know off the top of your head what the recommended approach is, that would be nice...

I'm not sure, but since that is only for CI, I think a simple way would be to just move the lockfile out of source for checking / building libraries, and then move it back for binaries.

hawkw:
> hmm, this actually breaks CI. it looks like we have a `tracing-subscriber` version constraint that is too old to ensure that APIs we use are present...let me fix that!

i just merged 62f5e576645934c08ba2505a794ce563cd6ab504, which should fix the incorrect minimum dependency version for `tracing-subscriber`. You'll want to merge `main` and re-build the lockfile in order to get CI to pass.

PR #138, opened 2021-09-11
Fix polls column width

### LINKED ISSUE #134: console: `Polls` are stripped when a value is greater than 99999
image

console doesn't show `Polls` properly when the value is greater than 99999.

hatoo:
Fixes #134 

This PR is from https://github.com/tokio-rs/console/issues/134#issuecomment-917436937

`Polls` columns in now left-justified. Since right-justify needs to know the maximum `Polls` width before the iteration of `sortedtask`.
I like right-justify numbers, but I chose a more simple implementation.
 
image

### DIFF
--- a/console/src/view/tasks.rs
+++ b/console/src/view/tasks.rs
@@ -83,7 +83,6 @@ impl List {
         // sense to have a lot of precision in timestamps (and this makes sure
         // there's room for the unit!)
         const DUR_PRECISION: usize = 4;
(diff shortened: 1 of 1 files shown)


hawkw (approved):
Thanks, this looks good to me. I agree that the right-justified numbers look nicer, but that it's more important to not truncate the poll count. :+1: 

We can figure out a nice way to right-justify here in the future?

Issue #50, opened 2021-06-15
subscriber: record a histogram of wake-to-poll times


seanmonstar:
A way for someone to better understand how the runtime itself is behaving is to record a history of how long it takes a task to be polled after it was been woken. We could either record this information per task, or on a whole. I think as a whole is probably sufficient, and would mean less data needs to be sent per task (see https://github.com/tokio-rs/console/pull/47#discussion_r651976189). But if there's a very helpful use case for knowing this data per task, we could do that.

sieut:
Hi @seanmonstar, I've been reading up on https://github.com/tokio-rs/console/pull/47 to see how to do this. Do you mind if I take a stab at this issue?

hawkw:
It's possible that this data is also collected by tokio's runtime metrics feature (https://github.com/tokio-rs/tokio/issues/4073), which I would like to start integrating into the console. It might be worth looking at the upstream API to see if that's the case.

Otherwise, we can definitely record this with the data we're already aggregating.

sieut:
Thank you @hawkw for the note. I read the RFC and it doesn't look like there's something we can use for this issue specifically. It looks like that work is more concerned with collecting perf-counter type metrics than aggregating latency.

hds:
This task is done now per task. As we don't currently have any "as a whole" or "per runtime" histograms, I think that we can consider it complete after #406 was merged.

Any histograms per runtime would be added once runtimes themselves have been instrumented in Tokio.

PR #139, opened 2021-09-12
feat(console): add wake-to-poll histogram

sieut:
- refactored MiniHistogram so wake to poll histogram can reuse some
  code from poll duration histogram
- added Percentiles widget so wake to poll can reuse code from poll
  duration histogram

for https://github.com/tokio-rs/console/issues/50

### DIFF
--- a/console-subscriber/src/aggregator/mod.rs
+++ b/console-subscriber/src/aggregator/mod.rs
@@ -97,6 +97,8 @@ pub(crate) struct Aggregator {
 
     /// The time "state" of the aggregator, such as paused or live.
     temporality: Temporality,
+
+    wake_to_poll_times_histogram: Histogram<u64>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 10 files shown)


sieut:
This PR is quite long with the refactoring, I can split it into two if that's better for reviewing

seanmonstar:
This is excellent work, thanks for getting started here! I think one thing to work out that may affect this PR is whether we want a histogram per task, or if we just want the history of the whole runtime.

I'm tempted to say both sound interesting... The full runtime histogram can help show the overall "health" or busy level. If things are starting to take longer and longer to be polled, it could imply the system has too much load. But knowing at an individual task level could also help understand, when load got high, or another few tasks ate up precious CPU time, that it delayed waking some specific task.

By keeping a log of all the events and timestamps in the "recording", we could always reconstruct either of these. So probably we also want to consider the cost of building these up per tasks and sending them over the wire...

All of my rambling didn't have any decisions, just musings, but hopefully with some joint thinking we'll quickly arrive at a proper balance. :)

sieut:
@seanmonstar thank you for the thoughts! I agree with you that recording this data for both runtime and per task is useful here.

> So probably we also want to consider the cost of building these up per tasks and sending them over the wire...

This is probably my main concern in keeping logs of events. Given that HdrHistograms are pretty small, another option would be to have another histogram in each `TaskStats`. What do you think about that idea?

hds:
As this PR was not progressing, a new PR was opened (#409) which implemented the new histogram. That PR has been merged, so I'll close this one.

PR #140, opened 2021-09-13
feat(subscriber): use per-layer filtering

### LINKED ISSUE #76: Need some way to enable "tokio=trace" filter for the console Layer only
First off...this is super cool, thanks for creating it!

The requirement to enable trace-level events for tokio to use this console means that other layers also receive those events, which makes a fmt Layer pretty unusable without out-of-band filtering.

Per https://github.com/tokio-rs/tracing/issues/302, it looks like per-layer filtering does not yet exist. Before I'd discovered that I'd tried combining separate `EnvFilter`s with a fmt Layer and a console Layer using `Layer::and_then`, but this clearly doesn't work as expected.

Maybe there's another workaround I haven't thought of?

hawkw:
Currently, the `console_subscriber::build()` and
`console_subscriber::init()` functions configure an `EnvFilter` that
enables the `tokio` and `runtime` targets, plus any targets configured
by `RUST_LOG`. This means that the `tokio` spans used by the console
will always be enabled at the `TRACE` level for all layers. This
includes the `fmt` layer we previously added in `init`.

Since this resulted in the `fmt` layer logs being cluttered with a bunch
of `tokio=trace` spans and events even when only high-level application
logs are enabled by `RUST_LOG`, we removed the `fmt` layer in PR #64.

However, tokio-rs/tracing#1523 was recently merged and [released][1],
adding support for [per-layer filtering][2] to `tracing-subscriber`. We
can now use the per-layer filtering API to selectively enable the
`tokio`/`runtime` spans and events at the `TRACE` level _only_ for the
`TasksLayer`, and add a filter to the `fmt` layer based on `RUST_LOG`.
This allows us to put back the `fmt` layer in
`console_subscriber::init`.

Now, if we run the example app (which uses `init`) with a `RUST_LOG`
value that enables only the app's logs, we get nice output:

image

However, we can also enable the `console-subscriber` crate's internal
logging:

image

And, we can still enable `tokio=trace` ourselves, if we actually _want_
to see all the task spans and waker events:

image

I also added some `trace!` and `debug!` macros in `examples/app.rs` to
demo that `console_subscriber::init()` also enables logging.

Closes #76

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1501,9 +1501,9 @@ dependencies = [
 
 [[package]]
 name = "tracing-core"
-version = "0.1.18"
+version = "0.1.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a9ff14f98b1a4b289c6248a023c1c2fa1491062964e9fed67ab29c4e4da4a052"
+checksum = "46125608c26121c81b0c6d693eab5a420e416da7e43c426d2e8f7df8da8a3acf"
 dependencies = [
  "lazy_static",
 ]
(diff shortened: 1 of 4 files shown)


seanmonstar (approved):
🔥

PR #141, opened 2021-09-13
chore(api): move protobuf definitions into `console-api` dir

hawkw:
Currently, the protobuf definitions for the console API are symlinked
into the `console-api` directory. This is intended to support vendoring
the `console-api` crate (see #128).

However, some users have reported that the symlink isn't the ideal
solution. Windows users have had issues following the Unix symlink, so
#128 broke the build for people trying to develop the console on Windows
(I suspect it may still work in WSL?).

There's no _real_ reason that the `proto/` directory needs to be in the
repo root, rather than the `console-api` crate; the symlink was only
added to avoid moving it around. Therefore, it's fine to just move it
instead. This branch moves the `proto/` dir into `console-api/proto`,
hopefully fixing the build issues for Windows users.


hawkw:
cc @Milo123459, please let me know if this fixes the issues you were having building the console?

Milo123459:
Fixed my issue, thank you!

hawkw:
> Fixed my issue, thank you!

awesome, sorry for any inconvenience!

PR #145, opened 2021-09-14
feat(console): resources UI

zaharidichev:
This introduces another view that shows all the current resources. 

Screenshot 2021-09-14 at 21 31 52

### DIFF
--- a/console/src/main.rs
+++ b/console/src/main.rs
@@ -109,7 +109,14 @@ async fn main() -> color_eyre::Result<()> {
             let chunks = Layout::default()
                 .direction(Direction::Vertical)
                 .margin(0)
-                .constraints([Constraint::Length(1), Constraint::Percentage(95)].as_ref())
+                .constraints(
+                    [
+                        Constraint::Length(1),
+                        Constraint::Length(1),
+                        Constraint::Percentage(95),
+                    ]
+                    .as_ref(),
+                )
                 .split(f.size());
 
             let mut header_text = conn.render(&view.styles);
@@ -119,8 +126,12 @@ async fn main() -> color_eyre::Result<()> {
                     .push(Span::styled(" PAUSED", view.styles.fg(Color::Red)));
             }
             let header = Paragraph::new(header_text).wrap(Wrap { trim: true });
+            let view_controls =
+                Paragraph::new("views: Tasks (F1), Resources (F2)").wrap(Wrap { trim: true });
+
             f.render_widget(header, chunks[0]);
-            view.render(f, chunks[1], &mut state);
+            f.render_widget(view_controls, chunks[1]);
+            view.render(f, chunks[2], &mut state);
         })?;
     }
 }
--- a/console/src/main.rs
+++ b/console/src/main.rs
@@ -8,11 +8,11 @@ use tokio::sync::{mpsc, watch};
 use tui::{
     layout::{Constraint, Direction, Layout},
     style::Color,
-    text::Span,
+    text::{Span, Spans},
     widgets::{Paragraph, Wrap},
 };
 
-use crate::view::UpdateKind;
+use crate::view::{bold, UpdateKind};
 
 mod config;
 mod conn;
@@ -126,8 +126,14 @@ async fn main() -> color_eyre::Result<()> {
                     .push(Span::styled(" PAUSED", view.styles.fg(Color::Red)));
             }
             let header = Paragraph::new(header_text).wrap(Wrap { trim: true });
-            let view_controls =
-                Paragraph::new("views: Tasks (F1), Resources (F2)").wrap(Wrap { trim: true });
+            let view_controls = Paragraph::new(Spans::from(vec![
+                Span::raw("views: "),
+                bold("t"),
+                Span::raw(" = tasks, "),
+                bold("r"),
+                Span::raw(" = resources"),
+            ]))
+            .wrap(Wrap { trim: true });
 
             f.render_widget(header, chunks[0]);
             f.render_widget(view_controls, chunks[1]);
--- a/console/src/state/mod.rs
+++ b/console/src/state/mod.rs
@@ -1,3 +1,4 @@
+use self::resources::ResourcesState;
 use crate::view;
 use crate::warnings::Linter;
 use console_api as proto;
@@ -17,6 +18,7 @@ use tui::{
     text::Span,
 };
 
+pub mod resources;
 pub mod tasks;
 
 pub(crate) type DetailsRef = Rc<RefCell<Option<Details>>>;
@@ -27,6 +29,7 @@ pub(crate) struct State {
     last_updated_at: Option<SystemTime>,
     temporality: Temporality,
     tasks_state: TasksState,
+    resources_state: ResourcesState,
     current_task_details: DetailsRef,
     retain_for: Option<Duration>,
 }
@@ -105,6 +108,15 @@ impl State {
                 current_view.is_tasks_view(),
             )
         }
+
+        if let Some(resources_update) = update.resource_update {
+            self.resources_state.update_resources(
+                styles,
+                resources_update,
+                &self.metas,
+                current_view.is_resources_view(),
+            )
+        }
     }
 
     pub(crate) fn retain_active(&mut self) {
@@ -114,6 +126,7 @@ impl State {
 
         if let (Some(now), Some(retain_for)) = (self.last_updated_at(), self.retain_for) {
             self.tasks_state.retain_active(now, retain_for);
+            self.resources_state.retain_active(now, retain_for);
         }
     }
 
@@ -127,6 +140,10 @@ impl State {
 
     pub(crate) fn tasks_state_mut(&mut self) -> &mut TasksState {
         &mut self.tasks_state
+    }
+
+    pub(crate) fn resources_state_mut(&mut self) -> &mut ResourcesState {
+        &mut self.resources_state
     }
 
     pub(crate) fn update_task_details(&mut self, update: proto::tasks::TaskDetails) {
--- a/console/src/state/mod.rs
+++ b/console/src/state/mod.rs
@@ -33,11 +33,9 @@ pub(crate) struct State {
     current_task_details: DetailsRef,
     retain_for: Option<Duration>,
 }
-
-#[derive(Debug)]
-enum Temporality {
-    Live,
-    Paused,
+pub(crate) enum Visibility {
+    Show,
+    Hide,
 }
 
 #[derive(Debug)]
@@ -61,6 +59,12 @@ pub(crate) enum FieldValue {
     U64(u64),
     I64(i64),
     Debug(String),
+}
+
+#[derive(Debug)]
+enum Temporality {
+    Live,
+    Paused,
 }
 
 impl State {
@@ -101,21 +105,23 @@ impl State {
         }
 
         if let Some(tasks_update) = update.task_update {
-            self.tasks_state.update_tasks(
-                styles,
-                tasks_update,
-                &self.metas,
-                current_view.is_tasks_view(),
-            )
+            let visibility = if matches!(current_view, view::ViewState::TasksList) {
+                Visibility::Show
+            } else {
+                Visibility::Hide
+            };
+            self.tasks_state
+                .update_tasks(styles, tasks_update, &self.metas, visibility)
         }
 
         if let Some(resources_update) = update.resource_update {
-            self.resources_state.update_resources(
-                styles,
-                resources_update,
-                &self.metas,
-                current_view.is_resources_view(),
-            )
+            let visibility = if matches!(current_view, view::ViewState::ResourcesList) {
+                Visibility::Show
+            } else {
+                Visibility::Hide
+            };
+            self.resources_state
+                .update_resources(styles, resources_update, &self.metas, visibility)
         }
     }
 
--- a/console/src/state/resources.rs
+++ b/console/src/state/resources.rs
@@ -45,6 +45,7 @@ pub(crate) struct Resource {
     stats: ResourceStats,
     target: Arc<str>,
     concrete_type: String,
+    location: Option<proto::Location>,
 }
 
 pub(crate) type ResourceRef = Weak<RefCell<Resource>>;
@@ -159,6 +160,7 @@ impl ResourcesState {
                 target: meta.target.clone(),
                 concrete_type: resource.concrete_type,
                 meta_id,
+                location: resource.location,
             };
             let resource = Rc::new(RefCell::new(resource));
             new_list.push(Rc::downgrade(&resource));
@@ -224,6 +226,13 @@ impl Resource {
 
     pub(crate) fn dropped(&self) -> bool {
         self.stats.total.is_some()
+    }
+
+    pub(crate) fn location(&self) -> String {
+        if let Some(location) = self.location.as_ref() {
+            return format!("{} ", location);
+        }
+        "unknown location".to_string()
     }
 }
 
--- a/console/src/state/resources.rs
+++ b/console/src/state/resources.rs
@@ -1,4 +1,4 @@
-use crate::state::{truncate_registry_path, Field, Metadata};
+use crate::state::{truncate_registry_path, Field, Metadata, Visibility};
 use crate::view;
 use console_api as proto;
 use std::{
@@ -50,18 +50,18 @@ pub(crate) struct Resource {
 pub(crate) type ResourceRef = Weak<RefCell<Resource>>;
 
 #[derive(Debug)]
+pub(crate) struct Attribute {
+    field: Field,
+    unit: Option<String>,
+}
+
+#[derive(Debug)]
 struct ResourceStats {
     created_at: SystemTime,
     dropped_at: Option<SystemTime>,
     total: Option<Duration>,
     attributes: Vec<Attribute>,
     formatted_attributes: Vec<Vec<Span<'static>>>,
-}
-
-#[derive(Debug)]
-pub(crate) struct Attribute {
-    field: Field,
-    unit: Option<String>,
 }
 
 impl Default for SortBy {
@@ -121,11 +121,11 @@ impl ResourcesState {
         styles: &view::Styles,
         update: proto::resources::ResourceUpdate,
         metas: &HashMap<u64, Metadata>,
-        clear_new_list: bool,
+        visibility: Visibility,
     ) {
         let mut stats_update = update.stats_update;
         let new_list = &mut self.new_resources;
-        if clear_new_list {
+        if matches!(visibility, Visibility::Show) {
             new_list.clear();
         }
 
--- a/console/src/state/resources.rs
+++ b/console/src/state/resources.rs
@@ -154,7 +154,7 @@ impl ResourcesState {
             let kind = resource.kind?.into();
 
             // remove cargo part of the file path
-            resource.location = resource.location.map(|mut l| {
+            let location = resource.location.take().map(|mut l| {
                 if let Some(file) = l.file.take() {
                     let truncated = truncate_registry_path(file);
                     l.file = Some(truncated);
@@ -168,7 +168,7 @@ impl ResourcesState {
                 target: meta.target.clone(),
                 concrete_type: resource.concrete_type,
                 meta_id,
-                location: resource.location,
+                location,
             };
             let resource = Rc::new(RefCell::new(resource));
             new_list.push(Rc::downgrade(&resource));
--- a/console/src/state/resources.rs
+++ b/console/src/state/resources.rs
@@ -44,7 +44,7 @@ pub(crate) struct Resource {
     stats: ResourceStats,
     target: Arc<str>,
     concrete_type: String,
-    location: Option<proto::Location>,
+    location: String,
 }
 
 pub(crate) type ResourceRef = Weak<RefCell<Resource>>;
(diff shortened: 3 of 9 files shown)


zaharidichev:
Will update shortly to include the location of where the resource was created

hawkw:
@zaharidichev thanks so much for working on this! this is a rather large change --- in particular, it looks like there was some refactoring to set up for actually adding the new code, by adding the `State` type and module and renaming `tasks::State` to `TasksState`, which is now owned by `State`, as well as factoring out the `view::Table` type. I think these were all good decisions. However, it makes the diff a bit harder to review, since _all_ code in that module now appears to have been added in this branch. It would be much easier to review that branch if just the new code for actually implementing the resources UI was part of the diff.

Would you be willing to extract the module and type renaming changes to a separate branch, and rebase this branch on top of that branch? That way, this branch can contain only the diff for actually adding the resources UI, which should make the code much easier to review. We could land the renaming change very easily, and then this could be rebased back onto `main`. Thanks!

hawkw:
If you like, I'd be happy to make that change myself.

console/src/state/mod.rs
```diff
@@ -0,0 +1,406 @@
...
+    current_task_details: DetailsRef,
+    retain_for: Option<Duration>,
+}
+
+#[derive(Debug, Default)]
+pub(crate) struct Details {
+    task_id: u64,
+    poll_times_histogram: Option<Histogram<u64>>,
+    last_updated_at: Option<SystemTime>,
+}
```
hawkw: nit: i feel like this should probably go in the `tasks` module, since it's specifically _task_ details. if we add a details view for resources, i suspect that resource details will probably have different data...

console/src/state/mod.rs
```diff
@@ -0,0 +1,406 @@
...
+            return;
+        }
+
+        if let (Some(now), Some(retain_for)) = (self.last_updated_at(), self.retain_for) {
+            self.tasks_state.retain_active(now, retain_for);
+            self.resources_state.retain_active(now, retain_for);
+        }
+    }
+
+    pub(crate) fn details_ref(&self) -> DetailsRef {
```
hawkw: nit: should probably be `task_details_ref` or something now that the `State` type also has resource state?

console/src/state/resources.rs
```diff
@@ -0,0 +1,322 @@
...
+    pub(crate) fn dropped(&self) -> bool {
+        self.stats.total.is_some()
+    }
+
+    pub(crate) fn location(&self) -> String {
+        if let Some(location) = self.location.as_ref() {
+            return format!("{} ", location);
+        }
+        "unknown location".to_string()
+    }
```
hawkw: nit, take it or leave it: it seems like it would be nice to store the formatted location on the `Resource` struct, so that we don't allocate a new string to format it into every time we call `Resource::location`? then, we could format the location when the `Resource` struct is constructed, and have a `fn location(&self) -> &str` that borrows it, instead.

console/src/state/resources.rs
```diff
@@ -0,0 +1,322 @@
...
+                l
+            });
+            let resource = Resource {
+                id,
+                kind,
+                stats,
+                target: meta.target.clone(),
+                concrete_type: resource.concrete_type,
+                meta_id,
+                location: resource.location,
```
hawkw: nit: it's a little bit strange to me that we mutate the protobuf `resource`'s location field in place, and then add it to the new `Resource` struct, instead of just doing this:
```suggestion
            let location = resource.location.take().map(|mut l| {
                if let Some(file) = l.file.take() {
                    let truncated = truncate_registry_path(file);
                    l.file = Some(truncated);
                }
                l
            });
            let resource = Resource {
                id,
                kind,
                stats,
                target: meta.target.clone(),
                concrete_type: resource.concrete_type,
                meta_id,
                location,
```
but, not a major issue!

console/src/view/mod.rs
```diff
@@ -20,18 +31,28 @@ pub struct View {
...
     /// Inspecting a single task instance.
     TaskInstance(self::task::TaskView),
 }
 
+#[derive(Eq, PartialEq)]
+pub(crate) enum CurrentView {
+    Tasks,
+    Resources,
+    Task,
+}
```
hawkw: I'm a bit on the fence about adding a `CurrentView` enum in addition to `ViewState`.Why not just change `ViewState` like this:
```suggestion
#[derive(Eq, PartialEq)]
pub(crate) enum ViewState {
    /// The table list of all tasks.
    TasksList,
    /// The table list of all resources.
    ResourcesList,
    /// Inspecting a single task instance.
    TaskInstance(self::task::TaskView),
}
```
and have the `current_view` method be
```rust
    pub(crate) current_view(&self) -> &ViewState
        &self.state
    }
```

Not a big deal, but it might be simpler...

console/src/view/resources.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    fn render<B: tui::backend::Backend>(
+        &mut self,
+        styles: &view::Styles,
+        frame: &mut tui::terminal::Frame<B>,
+        area: layout::Rect,
+        state: &mut state::State,
+    ) {
+        const DUR_LEN: usize = 10;
+        const DUR_PRECISION: usize = 4;
```
hawkw: since these have the same values as the `DUR_LEN` and `DUR_PRECISION` constants in `view::tasks`
https://github.com/tokio-rs/console/blob/27717a2a76f9f5021a556199beedeaa346f6d794/console/src/view/tasks.rs#L81-L85

we should probably factor them out into consts in the `view` module, so that we can use the same value in both the resources _and_ tasks modules?

console/src/view/resources.rs
```diff
@@ -0,0 +1,185 @@
...
+            type_width.constraint(),
+            location_width.constraint(),
+            attributes_width,
+        ];
+
+        let table = table
+            .header(header)
+            .block(block)
+            .widths(widths)
+            .highlight_symbol(">> ")
```
hawkw: nit/TIOLI: maybe worth making this a const in the `view` module, since we use the same symbol here as in the tasks list? that way, both modules could just say
```suggestion
            .highlight_symbol(view::TABLE_HIGHLIGHT_SYMBOL)
```
or something? that way, if we change it, we would only have to change it in one place.

alternatively, this could be a method on `Styles`, which might be nice if we wanted to use a unicode character instead of `>> ` when UTF-8 is enabled...but that should be a separate change.

console/src/view/table.rs
```diff
@@ -0,0 +1,183 @@
...
+    layout,
+    text::{self, Span, Spans, Text},
+    widgets::TableState,
+};
+
+use std::cell::RefCell;
+use std::rc::Weak;
+
+pub(crate) trait TableList<R, S: TryFrom<usize>> {
+    fn header(&self) -> &'static [&'static str];
```
hawkw: it seems like this could probably be an associated constant, rather than a method, since the headers are not going to change dynamically:
```suggestion
    const HEADER: &'static [&'static str];
```

console/src/view/table.rs
```diff
@@ -0,0 +1,183 @@
...
+            if i == 0 {
+                // If the first item is currently selected, wrap around to the
+                // last item.
+                resources.len() - 1
+            } else {
+                // Otherwise, decrease the selected item by 1.
+                i - 1
+            }
+        })
+    }
```
hawkw: i feel like these could just be methods on `TableListState` rather than on the `TableList` trait? we could remove the `R` type parameter from `scroll_with` if we changed the closure to just pass in the `len` rather than the vec...

console/src/view/table.rs
```diff
@@ -0,0 +1,183 @@
...
+            text::Span::raw(" = scroll, "),
+            bold(styles.if_utf8("\u{21B5}", "enter")),
+            text::Span::raw(" = view details, "),
+            bold("i"),
+            text::Span::raw(" = invert sort (highest/lowest), "),
+            bold("q"),
+            text::Span::raw(" = quit"),
+        ]))
+    }
+}
```
hawkw: nit: does this really need to be a method on `TableList`? it seems like it could just be a free function in this module, instead, since it never actually uses any data from `self`...

```suggestion
}

pub(in crate::view) fn controls(tyles: &view::Styles) -> Text {
    tui::text::Text::from(Spans::from(vec![
        Span::raw("controls: "),
        bold(styles.if_utf8("\u{2190}\u{2192}", "left, right")),
        text::Span::raw(" = select column (sort), "),
        bold(styles.if_utf8("\u{2191}\u{2193}", "up, down")),
        text::Span::raw(" = scroll, "),
        bold(styles.if_utf8("\u{21B5}", "enter")),
        text::Span::raw(" = view details, "),
        bold("i"),
        text::Span::raw(" = invert sort (highest/lowest), "),
        bold("q"),
        text::Span::raw(" = quit"),
    ]))
}
```

console/src/view/table.rs
```diff
@@ -0,0 +1,183 @@
...
+
+pub(crate) trait SortBy {
+    fn as_usize(&self) -> usize;
```
hawkw: nit: i would probably give this a name that explains what the usize _means_ in this context, rather than just saying what type conversion is performed. In this case, it's converting the enum to a column index in the table...so i might call this
```suggestion
    fn as_column(&self) -> usize;
```

what do you think?

console/src/view/table.rs
```diff
@@ -0,0 +1,183 @@
...
+
+use std::cell::RefCell;
+use std::rc::Weak;
+
+pub(crate) trait TableList<R, S: TryFrom<usize>> {
+    fn header(&self) -> &'static [&'static str];
+
+    fn state_mut(&mut self) -> &mut TableListState<R, S>;
+
+    fn state(&self) -> &TableListState<R, S>;
```
hawkw: this...feels kind of strange to me. the fact that a `TableList` is expected to _own_ a `TableListState`, but then it's only ever implemented _for_ `TableListState` seems like a kind of weird factoring. what do you think about changing the trait so that more of its methods are actually methods _on_ `TableListState`, like this:

```rust
pub(crate) trait TableList {
    type Row;
    type Sort: SortBy + TryFrom<usize>;

    fn header() -> &'static [&'static str];

    fn render<B: tui::backend::Backend>(
        state: &mut TableListState<Self>,
        styles: &view::Styles,
        frame: &mut tui::terminal::Frame<B>,
        area: layout::Rect,
        state: &mut state::State,
    );
}

pub(crate) struct TableListState<T: TableList> {
    pub(crate) sorted_items: Vec<Weak<RefCell<T::Row>>>,
    pub(crate) sort_by: T::Sort,
    pub(crate) selected_column: usize,
    pub(crate) sort_descending: bool,
    pub(crate) table_state: TableState,
}

impl<T: TableList> TableListState<T> {
    pub(in crate::view) fn len(&self) -> usize {
        self.sorted_items.len()
    }

    pub(in crate::view) fn update_input(&mut self, event: input::Event) {
         // ...
    }

    pub(in crate::view) fn key_input(&mut self, input::KeyEvent { code, .. }: input::KeyEvent) {
        use input::KeyCode::*;
        let header_len = T::header().len();
        match code {
            Left => {
                if self.selected_column == 0 {
                    self.selected_column = header_len - 1;
                } else {
                    self.selected_column -= 1;
                }
            }
            Right => {
                if self.selected_column == header_len - 1 {
                    self.selected_column = 0;
                } else {
                    self.selected_column += 1;
                }
            }
            Char('i') => {
                self.sort_descending = !self.sort_descending
            }
            Down => self.scroll_next(),
            Up => self.scroll_prev(),
            _ => {} // do nothing for now...
        }
        if let Ok(sort_by) = S::try_from(self.selected_column) {
            self.sort_by = sort_by;
        }
    }

    pub(in crate::view) fn scroll_with(&mut self, f: impl Fn(&Vec<Weak<RefCell<T::Row>>>, usize) -> usize) {
        // If the list of sorted items is empty, don't try to scroll...
        if self.sorted_items.is_empty() {
            self.table_state.select(None);
            return;
        }

        // Increment the currently selected row, or if no row is selected, start
        // at the first row.
        let i = self.table_state.selected().unwrap_or(0);
        let i = f(&self.sorted_items, i);
        self.table_state.select(Some(i));
    }

    pub(in crate::view) fn scroll_next(&mut self) {
         // ...
    }

    pub(in crate::view) fn scroll_prev(&mut self) {
        // ...
    }

    pub(in crate::view) fn controls(&self, styles: &view::Styles) -> Text {
        // ...
    }
    
    pub(in crate::view) fn render<B: tui::backend::Backend>(
        &mut self,
        styles: &view::Styles,
        frame: &mut tui::terminal::Frame<B>,
        area: layout::Rect,
    ) {
        T::render(self, styles, frame, layout, self.state)
    }
}
```
Then, we could implement `TableList` for zero-sized structs that just define the header and `render` function, like this:

```rust
#[derive(Debug, Default)]
struct ResourcesState {}

impl TableList for ResourcesState {
    type Row = Resource;
    type Sort = resources::SortBy;

    fn header(&self) -> &'static [&'static str] {
        &[
            "ID",
            "Kind",
            "Total",
            "Target",
            "Type",
            "Location",
            "Attributes",
        ]
    }

    fn render<B: tui::backend::Backend>(
        state: &mut TableListState<Self>,
        styles: &view::Styles,
        frame: &mut tui::terminal::Frame<B>,
        area: layout::Rect,
        state: &mut state::State,
    ) {
       // ... actually render the table ...
    }
}
```

this would also give us a place to eventually factor out some of the common parts of the `render` function.

what do you think? this might end up being just as complex as the current implementation...

console-subscriber/src/visitors.rs
```diff
@@ -118,9 +121,26 @@ impl ResourceVisitor {
     const RES_CONCRETE_TYPE_FIELD_NAME: &'static str = "concrete_type";
     const RES_KIND_FIELD_NAME: &'static str = "kind";
     const RES_KIND_TIMER: &'static str = "timer";
+    const LOCATION_FILE: &'static str = "loc.file";
+    const LOCATION_LINE: &'static str = "loc.line";
+    const LOCATION_COLUMN: &'static str = "loc.col";
```
hawkw: we should probably do something similar for task spans in a separate branch...

console-api/proto/resources.proto
```diff
@@ -37,8 +37,10 @@ message Resource {
     common.MetaId metadata = 2;
     // The resources's concrete rust type.
     string concrete_type = 3;
+    // The location in code where the resource was created.
+    common.Location location = 4;
     // The kind of resource (e.g timer, mutex)
-    Kind kind = 4;
+    Kind kind = 5;
```
hawkw: i'd really prefer to make the protobuf and console-subscriber changes in a separate branch from the console change that adds the resources UI, so that the git history is easier to read. if it's not too much work, would you mind moving the proto/subscriber changes to a separate PR and having this branch depend on that? thanks!

hawkw (commented):
overall, this looks like basically the right approach and i'm really excited to merge it in!

i had some suggestions, but they're not really hard blockers. let me know if you have any questions?

it would be nice if some of these changes could be moved to separate pull requests, if that's not *too* much extra work...but if it's really impractical to do that, I'm fine with merging this as one commit.

zaharidichev:
> Would you be willing to extract the module and type renaming changes to a separate branch, and rebase this branch on top of that branch? That way, this branch can contain only the diff for actually adding the resources UI, which should make the code much easier to review. We could land the renaming change very easily, and then this could be rebased back onto main. Thanks!

@hawkw That sounds like a good idea. Will do that. And thanks a lot for looking at this. I know its a pretty large diff

console/src/state/resources.rs
zaharidichev: Yes... that was a bit strange for sure

console/src/main.rs
```diff
@@ -119,8 +126,12 @@ async fn main() -> color_eyre::Result<()> {
                     .push(Span::styled(" PAUSED", view.styles.fg(Color::Red)));
             }
             let header = Paragraph::new(header_text).wrap(Wrap { trim: true });
+            let view_controls =
+                Paragraph::new("views: Tasks (F1), Resources (F2)").wrap(Wrap { trim: true });
+
             f.render_widget(header, chunks[0]);
-            view.render(f, chunks[1], &mut state);
+            f.render_widget(view_controls, chunks[1]);
+            view.render(f, chunks[2], &mut state);
```
hawkw: TIOLI: we could alternatively just push this to a new line in the `header_text` paragraph (and make that chunk 2 lines long). either way is probably fine though.

console/src/main.rs
```diff
@@ -119,8 +126,12 @@ async fn main() -> color_eyre::Result<()> {
                     .push(Span::styled(" PAUSED", view.styles.fg(Color::Red)));
             }
             let header = Paragraph::new(header_text).wrap(Wrap { trim: true });
+            let view_controls =
+                Paragraph::new("views: Tasks (F1), Resources (F2)").wrap(Wrap { trim: true });
```
hawkw: nit: would be nice if this was formatted more like the other controls text (with the keys in bold, delimited by equals signs), for consistency.

console/src/state/resources.rs
```diff
@@ -0,0 +1,322 @@
...
+
+impl From<proto::resources::resource::Kind> for Kind {
+    fn from(pb: proto::resources::resource::Kind) -> Self {
+        use proto::resources::resource::kind::Kind::Known as PbKnown;
+        use proto::resources::resource::kind::Kind::Other as PBOther;
+        use proto::resources::resource::kind::Known::Timer as PbTimer;
+
+        match pb.kind.expect("a resource should have a kind field") {
+            PbKnown(known) if known == (PbTimer as i32) => Kind::Timer,
+            PbKnown(known) => panic!("failed to parseknown kind from {}", known),
```
hawkw:
```suggestion
            PbKnown(known) => panic!("failed to parse known kind from {}", known),
```

console/src/state/resources.rs
```diff
@@ -0,0 +1,322 @@
...
+            PBOther(other) => Kind::Other(other),
+        }
+    }
+}
```
hawkw: nit, take it or leave it: should this maybe be a `TryFrom` implementation, rather than just panicking? it might be better to just log and ignore resources with messed up kinds?

console/src/state/resources.rs
```diff
@@ -0,0 +1,322 @@
...
+                    field,
+                    unit: pb.unit,
+                })
+            })
+            .collect::<Vec<_>>();
+
+        let formatted_attributes = Attribute::make_formatted(styles, &mut attributes);
+        let created_at = pb
+            .created_at
+            .expect("task span was never created")
```
hawkw:
```suggestion
            .expect("resource span was never created")
```

console/src/view/mod.rs
```diff
@@ -93,12 +104,28 @@ impl View {
...
+                        self.state = ResourcesList;
+                    }
                     _ => {
                         // otherwise pass on to view
                         self.tasks_list.update_input(event);
                     }
                 }
             }
+            ResourcesList => {
+                // The enter key changes views, so handle here since we can
```
hawkw: enter key?

console/src/view/mod.rs
```diff
@@ -93,12 +104,28 @@ impl View {
                             ));
                         }
                     }
+                    key!(F(2)) => {
```
hawkw: somewhat on the fence about using function keys here; on some keyboard layouts they might require a second keypress, be mapped to other commands by the user, or be on the touch bar on some macbooks...part of me wants to suggest `r` for resources and `t` for tasks, instead. but, i could be convinced otherwise...

we could always add support for custom keymaps later!

hawkw (approved):
overall, this looks great. i had some small suggestions.

also, it looks like a couple of my previous suggestions are still valid, if you're interested in addressing those as well...

console/src/state/resources.rs
```diff
@@ -0,0 +1,332 @@
...
+    total: Option<Duration>,
+    attributes: Vec<Attribute>,
+    formatted_attributes: Vec<Vec<Span<'static>>>,
+}
+
+#[derive(Debug)]
+pub(crate) struct Attribute {
+    field: Field,
+    unit: Option<String>,
+}
```
hawkw: style nit: mind putting the `pub(crate)` struct before the module-private one?
```suggestion
#[derive(Debug)]
pub(crate) struct Attribute {
    field: Field,
    unit: Option<String>,
}

#[derive(Debug)]
struct ResourceStats {
    created_at: SystemTime,
    dropped_at: Option<SystemTime>,
    total: Option<Duration>,
    attributes: Vec<Attribute>,
    formatted_attributes: Vec<Vec<Span<'static>>>,
}
```

console/src/view/resources.rs
```diff
@@ -0,0 +1,189 @@
...
+            .sort(now, &mut table_list_state.sorted_items);
+
+        let dur_cell = |dur: std::time::Duration| -> Cell<'static> {
+            Cell::from(styles.time_units(format!(
+                "{:>width$.prec$?}",
+                dur,
+                width = DUR_LEN,
+                prec = DUR_PRECISION,
+            )))
+        };
```
hawkw: nit/TIOLI: this is copied from the tasks list view, right? it _could_ be worth making it a function in `view`...but, then we would have to pass in `Styles` every time we call it instead of capturing it as a closure. so, maybe this is actually nicer...

hawkw: ...actually, i see that we only have one duration for resources currently. maybe this doesn't really need to be a function here, we could just do it once to format the total time. i think it makes sense for it to be a function in the tasks list since it's repeated a bunch.

console/src/state/resources.rs
```diff
@@ -0,0 +1,332 @@
...
+    pub fn sort(&self, now: SystemTime, resources: &mut Vec<Weak<RefCell<Resource>>>) {
+        match self {
+            Self::Rid => {
+                resources.sort_unstable_by_key(|resource| resource.upgrade().map(|r| r.borrow().id))
+            }
+            Self::Kind => resources.sort_unstable_by_key(|resource| {
+                resource.upgrade().map(|r| r.borrow().kind.clone())
+            }),
+            Self::ConcreteType => resources.sort_unstable_by_key(|resource| {
+                resource.upgrade().map(|r| r.borrow().concrete_type.clone())
```
hawkw, line 83: it's a bummer that we have to clone all the concrete type strings _every_ time we sort the resources. we might want to consider `Arc`ing them to make this lighter-weight, since we can't borrow them as `&str`s because they would outlive the `RefCell`'s `Ref` lifetime...

i also have some future ideas about possible string interning but that's best saved for a follow-up.

hawkw (approved):
overall, looks good to me! i pinged @seanmonstar to take a quick look as well, but hopefully we can get this merged today!

console/src/state/resources.rs
```diff
@@ -0,0 +1,332 @@
...
+    pub(crate) fn take_new_resources(&mut self) -> impl Iterator<Item = ResourceRef> + '_ {
+        self.new_resources.drain(..)
+    }
+
+    pub(crate) fn update_resources(
+        &mut self,
+        styles: &view::Styles,
+        update: proto::resources::ResourceUpdate,
+        metas: &HashMap<u64, Metadata>,
+        clear_new_list: bool,
```
seanmonstar: I find that functions with boolean arguments are less clear, especially when the function becomes more complex (Simple yes-no functions are fine). I'd personally advocate for a more specific enum argument, even if it's just this "yes" and "no" variant names. I think in this case we can make a nicer one:

```rust
enum Visibility {
    Show,
    Hide,
}
```

seanmonstar (approved):
Looks great! I did leave a note about a boolean vs an enum argument, but take that as you will. Thank you for the exciting work! 😍

hawkw (approved):

hawkw:
:tada:

PR #146, opened 2021-09-14
make the console binary vendorable

guswynn:
I am back with "pls help me make this crate vendorable"

2 main changes here:

1. Rename the crate to something not taken on `crates.io` (`cargo vendor` appears to be interested in clashing names, even if I import this crate as `tokio-console = { package = console ... }`. This will be required when its launched crates.io anyways
2. add a dummy `lib.rs` to add a library target to the binary crate in addition to the binary target. Currently, the way our system works is it uses cargo metadata to find all the metadata for all the crates we have vendored. Cargo does NOT add binary-only crates to this, but DOES add info about binary targets if the crate also has a lib target. See https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/cargo.20metadata.20with.20binary.20git.20dep for more info. This is a far simpler fix than doing something like copying and renaming the crate locally.

`cargo run` still works in the workspace

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -238,31 +238,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "console"
-version = "0.1.0"
-dependencies = [
- "atty",
- "chrono",
- "clap",
- "color-eyre",
- "console-api",
- "crossterm",
- "futures",
- "h2",
- "hdrhistogram",
- "humantime",
- "once_cell",
- "prost-types",
- "regex",
- "tokio",
- "tonic",
- "tracing",
- "tracing-journald",
- "tracing-subscriber",
- "tui",
-]
-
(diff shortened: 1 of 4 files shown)


hawkw:
Just to make sure I understand what's going on here, @guswynn: what's the motivation behind vendoring the console command-line application? Is it a build dependency of your application? Or, are you generally in the practice of vendoring the source code for other development tools that are used while working on your app?

guswynn:
@hawkw yeah, other dev tools (like the protoc we use to build console-api!) are vendored source code as well!

hawkw:
Okay, thanks for clarifying --- if you were actually including the console binary in application builds, I would have been curious about the use case for that...

hawkw (approved):

PR #151, opened 2021-09-17
fix(console): make LostWaker lint stricter to reduce false positives

### LINKED ISSUE #149: warnings: False positives with `LostWaker` warning
Currently, the `LostWaker` warning seems to often display false positives. These seem to occur when a task has a single waker clone, is woken by value (dropping the waker), and yields _after_ having been woken. This seems to occur the most frequently when a task self-wakes (and that _may_ be the only instance where this occurs?).

It would be nice to fix this --- the warning _should_ only be shown when a task definitely won't be woken again. However, I'm not immediately sure how we would determine this.

One potential solution is that the false-positives seem to be pretty transient. They flicker in and out of existence rapidly. We could potentially just only display the lint if the task has _remained_ in the 0-waker state for a long time.

seanmonstar:
Closes #149

### DIFF
--- a/console/src/tasks.rs
+++ b/console/src/tasks.rs
@@ -407,6 +407,13 @@ impl Task {
         self.self_wakes().percent_of(self.wakes())
     }
 
+    /// Returns whether this task has signaled via its waker to run again.
+    ///
+    /// Once the task has been polled, this is changed back to false.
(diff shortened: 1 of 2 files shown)


hawkw (approved):
<3

PR #153, opened 2021-09-20
refac(console): refactor table view

zaharidichev:
This change factors out the table logic from the tasks view in order
to make it reusable for future features (e.g. resources view).

### DIFF
--- a/console/src/view/mod.rs
+++ b/console/src/view/mod.rs
@@ -146,10 +146,7 @@ impl View {
 
 impl ViewState {
     pub(crate) fn is_tasks_view(&self) -> bool {
-        if let Self::TasksList = self {
-            return true;
-        }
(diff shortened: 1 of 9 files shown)


console/src/view/mod.rs
```diff
@@ -109,21 +121,35 @@ impl View {
...
+        &self.state
+    }
+}
+
+impl ViewState {
+    pub(crate) fn is_tasks_view(&self) -> bool {
+        if let Self::TasksList = self {
+            return true;
+        }
+        false
```
hawkw: nit: we could simplify this to just be
```suggestion
        matches!(self, Self::TasksList)
```

hawkw (approved):
looks good to me, thanks!

PR #154, opened 2021-09-20
chore(console): use `Location` for tasks and resources

zaharidichev:
This pr modifies the proto schema and the subscriber crate to
allow us to capture the spawn location of a task and the instantiation
location of a resource into a `Location` proto message rather than
a field. Note that the special treatment given to the `spawn.location`
filed, found on tasks has not been touched in order not to break 
apps that use tokio versions, not emitting this data as structured
fields.

### DIFF
--- a/console-api/proto/resources.proto
+++ b/console-api/proto/resources.proto
@@ -40,7 +40,7 @@ message Resource {
     // The kind of resource (e.g timer, mutex)
     Kind kind = 4;
     // The location in code where the resource was created.
-    common.Location location = 4;
+    common.Location location = 5;
 
     message Kind {
         oneof kind {
--- a/console-api/proto/resources.proto
+++ b/console-api/proto/resources.proto
@@ -37,8 +37,10 @@ message Resource {
     common.MetaId metadata = 2;
     // The resources's concrete rust type.
     string concrete_type = 3;
+    // The location in code where the resource was created.
+    common.Location location = 4;
     // The kind of resource (e.g timer, mutex)
-    Kind kind = 4;
+    Kind kind = 5;
 
     message Kind {
         oneof kind {
--- a/console-api/proto/resources.proto
+++ b/console-api/proto/resources.proto
@@ -37,10 +37,10 @@ message Resource {
     common.MetaId metadata = 2;
     // The resources's concrete rust type.
     string concrete_type = 3;
+    // The kind of resource (e.g timer, mutex)
+    Kind kind = 4;
     // The location in code where the resource was created.
     common.Location location = 4;
-    // The kind of resource (e.g timer, mutex)
-    Kind kind = 5;
 
     message Kind {
         oneof kind {
--- a/console-subscriber/src/aggregator/mod.rs
+++ b/console-subscriber/src/aggregator/mod.rs
@@ -150,14 +150,15 @@ struct Resource {
     metadata: &'static Metadata<'static>,
     concrete_type: String,
     kind: resource::Kind,
+    location: Option<proto::Location>,
 }
 
 /// Represents a key for a `proto::field::Name`. Because the
 /// proto::field::Name might not be unique we also include the
-/// metadata id in this key
+/// resource id in this key
 #[derive(Hash, PartialEq, Eq)]
 struct FieldKey {
-    meta_id: u64,
+    resource_id: u64,
     field_name: proto::field::Name,
 }
 
@@ -173,6 +174,7 @@ struct Task {
     id: Id,
     metadata: &'static Metadata<'static>,
     fields: Vec<proto::Field>,
+    location: Option<proto::Location>,
 }
 
 struct TaskStats {
@@ -590,7 +592,7 @@ impl Aggregator {
                 metadata,
                 at,
                 fields,
-                ..
+                location,
             } => {
                 let id = self.ids.id_for(id);
                 self.tasks.insert(
@@ -599,6 +601,7 @@ impl Aggregator {
                         id,
                         metadata,
                         fields,
+                        location,
                         // TODO: parents
                     },
                 );
@@ -703,6 +706,7 @@ impl Aggregator {
                 metadata,
                 kind,
                 concrete_type,
+                location,
                 ..
             } => {
                 let id = self.ids.id_for(id);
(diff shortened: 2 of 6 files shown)


hawkw:
> . Note that the special treatment given to the `spawn.location`
> filed, found on tasks has not been touched in order not to break
> apps that use tokio versions, not emitting this data as structured
> fields.

We should probably do a change upstream to replace `spawn.location` with structured fields in Tokio, but we should continue accepting both `spawn.location` _and_ structured location fields in order to maintain compatibility with older versions. So, :thumbsup:

console-api/proto/resources.proto
```diff
@@ -37,8 +37,10 @@ message Resource {
     common.MetaId metadata = 2;
     // The resources's concrete rust type.
     string concrete_type = 3;
+    // The location in code where the resource was created.
+    common.Location location = 4;
     // The kind of resource (e.g timer, mutex)
-    Kind kind = 4;
+    Kind kind = 5;
```
hawkw: nit: is there any reason we put the `location` field _before_ `Kind`? why not just add it afterwards, so we don't need to change the field number for the `Kind` field?
```protobuf
    // The kind of resource (e.g timer, mutex)
    Kind kind = 4;
    // The location in code where the resource was created.
    common.Location location = 5;
```

Changing the field number of an existing protobuf version will break protobuf backwards compatibility with earlier versions of the protobufs. Although we don't currently guarantee proto compatibility between two commits to this repo, we'll want to care about this once we start publishing releases of the console API, so it seems better to get in the practice of not re-numbering fields if we don't _need_ to...

console-subscriber/src/aggregator/mod.rs
```diff
@@ -767,7 +772,11 @@ impl Aggregator {
             } => {
                 let resource_id = self.ids.id_for(resource_id);
                 if let Some(mut stats) = self.resource_stats.update(&resource_id) {
-                    let upd_key = (&update.field).into();
+                    let field_name = update.field.name.clone().expect("field misses name");
```
hawkw: hmm, i'm not sure if we really want to panic here...this should maybe just log a warning and then bail?
```suggestion
                    let field_name = match update.field.name.clone() {
                        Some(name) => name,
                        None => {
                            tracing::warn!(?update.field, "field missing name, skipping...");
                            return;
                        }
                    };
```

console-subscriber/src/visitors.rs
```diff
@@ -36,6 +43,30 @@ pub(crate) struct FieldVisitor {
...
+///     kind = "local",
+///     task.name = "some_name",
+///     loc.file = "some_file.rs"
+///     loc.line = 555,
+///     loc.col = 5,
+/// );
+///
+/// Fields:
+/// concrete_type - indicates the concrete rust type for this resource
+/// kind - indicates the type of resource (i.e. timer, sync, io )
```
hawkw: nit: let's add code formatting to this doc comment.

also, I don't think the list of fields are correct; it looks like it was copied from the docs for `ResourceVisitor`. i went ahead and updated that too:

```suggestion
/// Used to extract the fields needed to construct
/// an `Event::Spawn` from the metadata of a tracing span
/// that has the following shape:
///
/// ```
/// tracing::trace_span!(
///     target: "tokio::task",
///     "runtime.spawn",
///     kind = "local",
///     task.name = "some_name",
///     loc.file = "some_file.rs"
///     loc.line = 555,
///     loc.col = 5,
/// );
/// ```
///
/// # Fields
///
/// This visitor has special behavior for `loc.line`, `loc.file`, and `loc.col`
/// fields, which are interpreted as a Rust source code location where the task
/// was spawned, if they are present. Other fields are recorded as arbitrary
/// key-value pairs.
```

console-subscriber/src/visitors.rs
```diff
@@ -119,8 +150,22 @@ impl ResourceVisitor {
     const RES_KIND_FIELD_NAME: &'static str = "kind";
     const RES_KIND_TIMER: &'static str = "timer";
 
-    pub(crate) fn result(self) -> Option<(String, resource::Kind)> {
-        self.concrete_type.zip(self.kind)
+    pub(crate) fn result(self) -> Option<(String, resource::Kind, Option<proto::Location>)> {
+        let concrete_type = self.concrete_type?;
+        let kind = self.kind?;
+
+        let location = if self.file.is_some() && self.line.is_some() && self.column.is_some() {
```
hawkw, line 162: i'm on the fence about this, but i think maybe we want to record locations that are "incomplete" (e.g. they are missing some of these fields). we can do this since all the fields on `proto::Location` are optional. I think we probably _don't_ want to record a location if we didn't find a file, since a column/line doesn't make sense without a file, but I think we might want to allow locations that have only file names and no line/column?

what do you think?

console-subscriber/src/visitors.rs
```diff
@@ -119,8 +150,22 @@ impl ResourceVisitor {
...
+            Some(proto::Location {
+                file: self.file,
+                line: self.line,
+                column: self.column,
+                ..Default::default()
+            })
+        } else {
+            None
+        };
```
hawkw, line 171: i wonder if it would make sense to factor out some of the common location-recording code between the resource and task visitors? not a blocker for this PR, but potentially a nice follow-up...

hawkw (approved):
looks good to me overall! i had a couple minor suggestions.

hawkw:
@zaharidichev i went ahead and made a couple small tweaks; beyond that, this LGTM and i'm going to merge it as soon as CI passes!

console-subscriber/src/visitors.rs
```diff
@@ -36,6 +43,35 @@ pub(crate) struct FieldVisitor {
...
+/// an `Event::Spawn` from the metadata of a tracing span
+/// that has the following shape:
+///
+/// ```
+/// tracing::trace_span!(
+///     target: "tokio::task",
+///     "runtime.spawn",
+///     kind = "local",
+///     task.name = "some_name",
+///     loc.file = "some_file.rs"
```
hawkw: missing comma here
```suggestion
///     loc.file = "some_file.rs",
```

zaharidichev:
@hawkw,  was going to address these now, but thanks for doing that!

hawkw:
@zaharidichev of course! i'm just very excited to get your main resources PR merged, so i figured i'd just make the minor changes i suggested :)

PR #157, opened 2021-09-20
docs: improve documentation on enabling required targets

hawkw:
This should hopefully make it a bit clearer which tracing targets need
to be enabled and how to do that.


blt:
This wording is clear to me.

PR #158, opened 2021-09-21
perf(console): quick and dirty string interning

hawkw:
While reviewing PR #145, I [noticed] that when we sort the task and
resource lists by the `name` field, we clone the actual strings for each
task/resource's name while sorting. This seems...sad, but it's
unfortunately necessary because the task/resource data is behind a
`RefCell`, and the refcell guard can't return a borrowed string that
outlives it. We should make these cheap to clone, such as by `Arc`ing
them.

The other thing I noticed was that there are likely to be a large number
of repeated strings that are recieved in data from the application, such
as targets and field names. Since these string values are coming from
the app rather than defined in the console, we can't easily replace them
with string _constants_...but we can intern them, so that we don't have
to store a ton of copies of the string "tokio::task" (for example), or a
given task name that's spawned a bunch of times. This branch adds a
_very_ simplistic string interning system based on a `HashSet` of `Rc`s,
which should reduce the console's memory footprint a bit (and possibly
make things like sorting and rendering the UI a bit faster?).

This is, of course, not the most optimal string interner implementation,
but we can always replace it with [something faster][1] later --- this
seemed good for now, since it's simple, easy to understand, and 100%
safe.

Running the console with the demo app for about ten seconds results in
us interning the following strings (note the counts, which show how many
times that string would have been duplicated):

<details>

<summary>Debug output</summary>

```rust
Strings {
    strings: {
        InternedStr(
            "duration.op",
            3,
        ),
        InternedStr(
            "tokio::time::driver::sleep",
            4,
        ),
        InternedStr(
            "runtime::resource::state_update",
            3,
        ),
        InternedStr(
            "spawn.location",
            13,
        ),
        InternedStr(
            "task1",
            2,
        ),
        InternedStr(
            "concrete_type",
            2,
        ),
        InternedStr(
            "task2",
            2,
        ),
        InternedStr(
            "op_name",
            4,
        ),
        InternedStr(
            "loc.line",
            2,
        ),
        InternedStr(
            "runtime::resource::poll_op",
            4,
        ),
        InternedStr(
            "is_ready",
            4,
        ),
        InternedStr(
            "wait",
            6,
        ),
        InternedStr(
            "console::serve",
            2,
        ),
        InternedStr(
            "source",
            3,
        ),
        InternedStr(
            "kind",
            14,
        ),
        InternedStr(
            "tokio::task",
            13,
        ),
        InternedStr(
            "loc.file",
            2,
        ),
        InternedStr(
            "duration.unit",
            3,
        ),
        InternedStr(
            "console::aggregate",
            2,
        ),
        InternedStr(
            "task.name",
            2,
        ),
        InternedStr(
            "loc.col",
            2,
        ),
        InternedStr(
            "duration",
            3,
        ),
    },
}
```

</details>

I imagine in a real program, there would be significantly more duplicate
strings eliminated, especially if we run the console for longer than 10
seconds. :)

[noticed]: https://github.com/tokio-rs/console/pull/145/files#r713207224
[1]: https://matklad.github.io/2020/03/22/fast-simple-rust-interner.html

### DIFF
--- a/console/src/state/mod.rs
+++ b/console/src/state/mod.rs
@@ -1,5 +1,8 @@
-use crate::view;
-use crate::warnings::Linter;
+use crate::{
+    intern::{self, InternedStr},
+    view,
+    warnings::Linter,
+};
 use console_api as proto;
 use std::{
     cell::RefCell,
@@ -8,7 +11,6 @@ use std::{
     fmt,
     io::Cursor,
     rc::Rc,
-    sync::Arc,
     time::{Duration, SystemTime},
 };
 use tasks::{Details, Task, TasksState};
@@ -29,6 +31,7 @@ pub(crate) struct State {
     tasks_state: TasksState,
     current_task_details: DetailsRef,
     retain_for: Option<Duration>,
+    strings: intern::Strings,
 }
 
 #[derive(Debug)]
@@ -39,15 +42,15 @@ enum Temporality {
 
 #[derive(Debug)]
 pub(crate) struct Metadata {
-    field_names: Vec<Arc<str>>,
-    target: Arc<str>,
+    field_names: Vec<InternedStr>,
+    target: InternedStr,
     id: u64,
     //TODO: add more metadata as needed
 }
 
 #[derive(Debug)]
 pub(crate) struct Field {
-    pub(crate) name: Arc<str>,
+    pub(crate) name: InternedStr,
     pub(crate) value: FieldValue,
 }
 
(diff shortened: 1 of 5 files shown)


console/src/state/mod.rs
```diff
@@ -39,15 +42,15 @@ enum Temporality {
 
 #[derive(Debug)]
 pub(crate) struct Metadata {
-    field_names: Vec<Arc<str>>,
-    target: Arc<str>,
+    field_names: Vec<InternedStr>,
```
seanmonstar, line 46: Since the interned strings are never pruned, this does mean that if a program had an event or span with a dynamically changing field name (is that a thing in `tracing`?), we could "leak" memory here... :shrug:

seanmonstar (approved):

zaharidichev (approved):
LGTM

console/src/state/mod.rs
hawkw: `tracing` field names will never change, although the wire proto might allow changing them.

I considered pruning interned strings by storing `Weak` refs in the hashmap, but I decided we should do that in a follow-up branch. Could add it now if you prefer.

seanmonstar: If that can't happen in `tracing`, that's fine. We just need to be careful not to use this for a field "value" or similar strings. :)

seanmonstar (approved):

PR #160, opened 2021-09-21
fix(console): assume waiting for first poll is not a lost waker

seanmonstar:
Noted in https://github.com/tokio-rs/console/issues/149#issuecomment-924390639


hawkw (approved):

hawkw:
Nice catch!

PR #162, opened 2021-09-22
pref(console): drop un-referenced interned strings

hawkw:
Follow-up from #158.

This should keep us from gradually using more and more memory if we
intern a bunch of strings that are only used by one or two tasks.

It will also periodically shrink the hashmap itself, so that we don't
allocate a bunch of hashmap capacity we never use again.

My initial thought was that we would do this using `Weak` references,
but it turns out it's actually much simpler if we just `retain` only
the `Rc`s with `strong_count` > 1.

We will only try to drop strings if we dropped any task or resource
data, to avoid iterating over the hash map in `retain` when we know
that no ref counts will have changed. This required changing the
`retain_active` methods for tasks and resources to return `bool`s
indicating whether anything was dropped.

### DIFF
--- a/console/src/intern.rs
+++ b/console/src/intern.rs
@@ -49,6 +49,35 @@ impl Strings {
         self.strings.insert(string.clone());
         string
     }
+
+    /// Drop any interned strings that are not currently referenced.
+    pub(crate) fn retain_referenced(&mut self) {
+        const FOUR_KILOBYTES: usize = 4 * 1024;
+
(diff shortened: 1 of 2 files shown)


seanmonstar (approved):

hawkw:
We may also want to consider using the `ShrinkMap` and `ShrinkVec` types from the `console-subscriber` crate in the CLI as well, but this would probably require factoring them out into their own crate...

PR #166, opened 2021-09-22
console: show spawn location as a column

zaharidichev:
This PR extracts the spawn location of a task into a separate
column, in the case where structured location information is
present. This depends on: https://github.com/tokio-rs/tokio/pull/4128

Screenshot 2021-09-22 at 22 56 50

### DIFF
--- a/console/src/state/tasks.rs
+++ b/console/src/state/tasks.rs
@@ -1,6 +1,6 @@
 use crate::{
     intern::{self, InternedStr},
-    state::{Field, Metadata, Visibility},
+    state::{format_location, Field, Metadata, Visibility},
     util::Percentage,
     view,
     warnings::Linter,
@@ -62,6 +62,7 @@ pub(crate) struct Task {
     name: Option<InternedStr>,
     /// Currently active warnings for this task.
     warnings: Vec<Linter<Task>>,
+    location: String,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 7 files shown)


console/src/view/task.rs
```diff
@@ -148,22 +148,27 @@ impl TaskView {
         ]);
 
         // Just preallocate capacity for ID, name, target, total, busy, and idle.
-        let mut metrics = Vec::with_capacity(6);
-        metrics.push(Spans::from(vec![
+        let mut overview = Vec::with_capacity(6);
```
hawkw: +1 for renaming this.

now that it also has a line for the location, we should probably update the `with_capacity` call:
```rust
        // Just preallocate capacity for ID, name, target, location, total, busy, and idle.
        let mut overview = Vec::with_capacity(7);
```

hawkw (approved):
overall, this looks good! i think we should probably also add support for sorting tasks by location, but it's up to you if you want to do that as part of this PR or in a follow-up?

console/src/state/tasks.rs
```diff
@@ -416,6 +418,11 @@ impl SortBy {
             Self::Polls => {
                 tasks.sort_unstable_by_key(|task| task.upgrade().map(|t| t.borrow().stats.polls))
             }
+            Self::Target => {
+                tasks.sort_unstable_by_key(|task| task.upgrade().map(|t| t.borrow().target.clone()))
+            }
```
hawkw, line 423: :+1: for adding this as well!

hawkw (approved):
wonderful!

PR #168, opened 2021-09-29
Remove `authors` fields from `Cargo.toml`s

lunacookies:
The `authors` field has been [effectively soft-deprecated](https://github.com/rust-lang/rust/issues/83227), as it is no longer required, shown on crates.io or docs.rs, and crates created with `cargo new` no longer automatically create the `authors` field. Keeping or removing the field is of course a matter of taste, but I thought that, since it doesn’t have any effect anymore, it might be easier to just remove it instead of updating it over time as contributors come and go.

### DIFF
--- a/console-api/Cargo.toml
+++ b/console-api/Cargo.toml
@@ -1,7 +1,6 @@
 [package]
 name = "console-api"
 version = "0.1.0"
-authors = ["Eliza Weisman <eliza@buoyant.io>"]
 edition = "2018"
 [features]
 # Generate code that is compatible with Tonic's `transport` module.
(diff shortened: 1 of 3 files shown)


hawkw (approved):
In the case of the `console` CLI app, the `authors` field _will_ auto-populate the help text generated by `clap`, which _might_ be worth keeping, especially in the case that the CLI is ever packaged by distros. But, we can add authors to the help text [using an attribute, as well](https://docs.rs/clap/3.0.0-beta.4/clap/index.html#using-derive-macros), so it's not critical that we keep it in the cargo metadata.

This is fine with me!

PR #169, opened 2021-09-30
fix(subscriber): unset subscriber in aggregator

hawkw:
This changes the spawned aggregator task in `console_subscriber` to
override the default `tracing` subscriber with `NoSubscriber`, which
 does nothing. This means that any events that occur inside the aggregator
task will _not_ be recorded. When instrumentation that the console
aggregates is triggered inside the aggregator, it should therefore not
cause more data to be recorded.


hawkw:
I think this work was obsoleted by #314

PR #170, opened 2021-09-30
fix(subscriber): unset default dispatcher in span callbacks

hawkw:
Tracing [automatically unsets the default dispatcher][1] when accessed
through `dispatcher::get_default`. This prevents infinite loops caused
by events inside of dispatcher code when recording new events.

However, when recording span callbacks (`enter`, `exit`, `record`, and
`close`), the dispatcher is *not* accessed through `get_default`, but
through the span's own `Dispatch` ref. In that case, the default
dispatcher is never unset. Therefore, events emitted inside of the
`TasksLayer` type's `on_enter`, `on_exit`, and `on_close` callbacks are
not disabled. This means that when instrumentation is added to the tokio
`Semaphore` (and thus also the MPSC), calls to `send` inside the
`TasksLayer` will trigger those events when recording a span callback.
This can cause a deadlock.

This PR changes the `TasksLayer` to set a no-op dispatcher as the
default inside of the `on_enter`, `on_exit`, and `on_close` callbacks.
This should fix the deadlock @zaharidichev noticed when instrumenting
the semaphore.

[1]: https://github.com/tokio-rs/tracing/blob/v0.1.x/tracing-core/src/dispatcher.rs#L791-L827

### DIFF
--- a/console-subscriber/src/lib.rs
+++ b/console-subscriber/src/lib.rs
@@ -11,8 +11,9 @@ use std::{
 use thread_local::ThreadLocal;
 use tokio::sync::{mpsc, oneshot};
 use tracing_core::{
+    dispatcher::{self, Dispatch},
     span,
-    subscriber::{self, Subscriber},
+    subscriber::{self, NoSubscriber, Subscriber},
     Metadata,
 };
 use tracing_subscriber::{layer::Context, registry::LookupSpan, Layer};
@@ -78,6 +79,9 @@ pub struct TasksLayer {
     ///
     /// TODO: Take some time to determine more reasonable numbers
     state_update_callsites: Callsites<32>,
+
+    /// Used for unsetting the default dispatcher inside of span callbacks.
+    no_dispatch: Dispatch,
 }
 
 pub struct Server {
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):

hawkw:
(waiting to hear from @zaharidichev as to whether this actually fixes the deadlock he saw)

zaharidichev:
@hawkw I am still seeing events coming through. [Here is a (very WIP) branch](https://github.com/zaharidichev/tokio/tree/zd/semaphore-instrument) with the instrumentation. If you run the example app and patch tokio with this branch, you will observe hang

hawkw:
> @hawkw I am still seeing events coming through. Here is a (very WIP) branch with the instrumentation. If run the example app and patch tokio with this branch, you will observe hang

Thanks, I think it'll be much easier to fix with the branch that produces the hang!

zaharidichev:
Ah, sorry. Forgot to link to it: https://github.com/zaharidichev/tokio/tree/zd/semaphore-instrument

hawkw:
> Ah, sorry. Forgot to link to it: https://github.com/zaharidichev/tokio/tree/zd/semaphore-instrument

@zaharidichev it looks like this branch doesn't have any commits that aren't on `master`...did you forget to push your changes?

PR #174, opened 2021-10-25
Fix import for clap beta.5

### LINKED ISSUE #173: console imports `Clap` trait that has been removed in clap 3.0.0-beta.5
console currently depends on clap v3 beta release.

The most immediate problem with this is that the clap v3 beta release has an unstable API. In the specific case I am currently observing, `console` is trying to `use clap::Clap;` trait, but that has been removed at least as of clap 3.0.0-beta.5.

My suggestion is that we try to revise the code to only depend on clap v2, since that is the officially recommended stable release from the clap team. (That, or we try to follow the moving target; note that the beta numbering system they are employing does not obey semver, AFAICT...)

pnkfelix:
Fix #173

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -180,9 +180,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.0.0-beta.4"
+version = "3.0.0-beta.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fcd70aa5597dbc42f7217a543f9ef2768b2ef823ba29036072d30e1d88e98406"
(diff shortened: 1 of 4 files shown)


pnkfelix:
This resolves breakage I saw locally where the import was silently upgraded to a more recent version than beta.2 that had breaking changes compared to beta.2.

hawkw:
Thanks for the fix! I'm surprised Cargo will do that, I thought `.beta.X` versions were considered non-compatible?

pnkfelix:
> Thanks for the fix! I'm surprised Cargo will do that, I thought `.beta.X` versions were considered non-compatible?

To be honest, I was surprised by what I observed as well.

I will dig a little deeper and see if I can figure out exactly what sequence of events caused me to encounter local version "upgrades" and subsequent build failures.

PR #175, opened 2021-10-27
Remove `chrono` from dependencies

Adam-Gleave:
Since RUSTSEC has been flagging up the vulnerabilities in `chrono` and the older `time` versions recently, `cargo audit` is grumbling about things quite a bit. I noticed that `tracing-subscriber` had been updated recently, so this should get rid of all the vulnerabilities.

`cargo audit` now has no complaints!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.15.1"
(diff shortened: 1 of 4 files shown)


hawkw (approved):
thanks for the update, this looks good to me!

we should maybe consider setting up dependabot for automatically updating deps in the future...

PR #178, opened 2021-11-02
fix(subscriber): fix potential spurious flush notifications

### LINKED ISSUE #177: console fails to connect when app has compute intensive task
Finally got this issue that has been bothering me for ages narrowed to small that I think demonstrates a clear problem.

See program below (which is, in part, an adaptation of the examples/app from console-subscriber).

```rust
#![allow(dead_code)]

use std::time::Duration;

static HELP: &str = r#"                                                                                                                                                                                                        
Example demo of console failing to connect when you spin up compute intensive task.                                                                                                                                            
                                                                                                                                                                                                                               
By default, the compute-intensive task *does* frequently call yield (but this                                                                                                                                                  
does not address the problem). One can also opt out of yielding (which continues                                                                                                                                               
to exhibit failure to connect), or can sleep for 10ms (which resumes the ability                                                                                                                                               
to connect).                                                                                                                                                                                                                   
                                                                                                                                                                                                                               
USAGE:                                                                                                                                                                                                                         
    toys [OPTION]                                                                                                                                                                                                              
                                                                                                                                                                                                                               
OPTIONS:                                                                                                                                                                                                                       
    -h, help    prints this message                                                                                                                                                                                            
    no-fact     skips spawning computation (so you can demonstrate successful connect)                                                                                                                                         
    yield-now   computation frequently calls yield_now (this is on by default)                                                                                                                                                 
    no-yield    skips calling yield from computation (ideally would also support connect)                                                                                                                                      
    sleep       sleeps for 10ms instead of yielding from computation (enables successful connect)                                                                                                                              
"#;

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
    console_subscriber::init();

    let mut run_fact = true;
    let mut yield_type = YieldType::YieldNow;
    for opt in std::env::args().skip(1) {
        match &*opt {
            "no_fact" | "no-fact" => {
                run_fact = false;
                println!("will not run compute intensive loop.");
                break;
            }
            "no_yield" | "no-yield"  => yield_type = YieldType::NoYield,
            "yield_now" | "yield-now" => yield_type = YieldType::YieldNow,
            "sleep" => yield_type = YieldType::SleepMillis(10),
            "-h" | "help" | "-help" | "--help" => {
                println!("{}", HELP);
                return Ok(());
            }
            opt => { panic!("unrecognized command line option: `{}`", opt); }
        }
        println!("Will now yield with yield type {:?}", yield_type);
    }

    let task1 = tokio::task::Builder::new()
        .name("task1")
        .spawn(spawn_tasks(1, 10));
    let task2 = tokio::task::Builder::new()
        .name("task2")
        .spawn(spawn_tasks(10, 30));

    let mut handles = Vec::new();
    if run_fact {
        handles.push(tokio::task::spawn(async move {
            loop { fact_yielding(10, yield_type).await; }
        }));
    }

    println!("handles.len(): {:?}", handles.len());

    println!("joining task1 and task2");
    let result = tokio::try_join! {
        task1,
        task2,
    };

    result?;

    println!("joining handles");
    for h in handles {
        let val = h.await?;
        tracing::trace!("h.await?: {:?}", val);
    }

    Ok(())
}


// Computes factorial of n, printing progress along the way.                                                                                                                                                                   
async fn fact_stiff(n: u32) -> f64 {
    let mut i = 0_u32;
    let mut accum = 1_f64;
    loop {
        tracing::trace!("i: {I} fact_{N}({I}): {A}", N=n, I=i, A=accum);
        if i == n { break; }
        i += 1;
        accum *= i as f64;
    }
    return accum;
}

#[derive(Copy, Clone, Debug)]
enum YieldType {
    NoYield,
    YieldNow,
    SleepMillis(u64),
}

// Computes factorial of n, printing progress along the way.                                                                                                                                                                   
async fn fact_yielding(n: u32, yt: YieldType) -> f64 {
    let mut i = 0_u32;
    let mut accum = 1_f64;
    loop {
        tracing::trace!("i: {I} fact_{N}({I}): {A}", N=n, I=i, A=accum);
        if i == n { break; }
        match yt {
            YieldType::NoYield => {}
            YieldType::YieldNow => {
            }
            YieldType::SleepMillis(ms) => {
                tokio::time::sleep(Duration::from_millis(ms)).await;
            }
        }
        i += 1;
        accum *= i as f64;
    }
    return accum;
}

#[tracing::instrument]
async fn spawn_tasks(min: u64, max: u64) {
    loop {
        for i in min..max {
            tracing::trace!(i, "spawning wait task");
            tokio::task::Builder::new().name("wait").spawn(wait(i));

            let sleep = Duration::from_secs(max) - Duration::from_secs(i);
            tracing::trace!(?sleep, "sleeping...");
            tokio::time::sleep(sleep).await;
        }
    }
}

#[tracing::instrument]
async fn wait(seconds: u64) {
    tracing::debug!("waiting...");
    tokio::time::sleep(Duration::from_secs(seconds)).await;
    tracing::trace!("done!");
}
```

hawkw:
Currently, there is a potential issue with the flush notification
channel. An `AtomicBool` is used to determine whether a flush
notification should be sent, so that only one thread will trigger a
flush when approaching the flush capacity. However, the `AtomicBool` is
cleared by the aggregator task _as soon as_ it's woken by the flush
notification. This seems to potentially cause a situation where several
notifications are sent while the channel capacity is below the flush
limit, because the flag is cleared as soon as the aggregator task is
_woken_, and before it has the time to actually drain the channel. This
seems to result in many notifications being queued, causing the
aggregator task to loop a bunch of times.

This branch fixes this by changing the aggregator task to only clear the
flag when it has drained all queued events, hopefully reducing needless
notifications. This appears to partially resolve #177, although there
appears to also be an issue where console *updates* seem to become
increasingly delayed over time.

### DIFF
--- a/console-subscriber/src/lib.rs
+++ b/console-subscriber/src/lib.rs
@@ -222,7 +222,7 @@ impl TasksLayer {
 
         // Conservatively, start to trigger a flush when half the channel is full.
         // This tries to reduce the chance of losing events to a full channel.
-        let flush_under_capacity = config.event_buffer_capacity / 2;
+        let flush_under_capacity = config.event_buffer_capacity / 4;
 
         let server = Server {
             aggregator: Some(aggregator),
@@ -248,7 +248,7 @@ impl TasksLayer {
 }
 
 impl TasksLayer {
-    pub const DEFAULT_EVENT_BUFFER_CAPACITY: usize = 1024 * 10;
+    pub const DEFAULT_EVENT_BUFFER_CAPACITY: usize = 1024 * 100;
     pub const DEFAULT_CLIENT_BUFFER_CAPACITY: usize = 1024 * 4;
     pub const DEFAULT_PUBLISH_INTERVAL: Duration = Duration::from_secs(1);
 
@@ -406,6 +406,7 @@ where
     fn on_event(&self, event: &tracing::Event<'_>, ctx: Context<'_, S>) {
         let metadata = event.metadata();
         if self.waker_callsites.contains(event.metadata()) {
+            tracing::debug!(?event, "waker op event");
             let at = SystemTime::now();
             let mut visitor = WakerVisitor::default();
             event.record(&mut visitor);
@@ -424,6 +425,7 @@ where
             }
             // else unknown waker event... what to do? can't trace it from here...
         } else if self.poll_op_callsites.contains(event.metadata()) {
+            tracing::debug!(?event, "poll op event");
             match ctx.event_span(event) {
                 Some(resource_span) if self.is_resource(resource_span.metadata()) => {
                     let mut poll_op_visitor = PollOpVisitor::default();
@@ -463,6 +465,7 @@ where
                 ),
             }
         } else if self.state_update_callsites.contains(event.metadata()) {
+            tracing::debug!(?event, "state update");
             match ctx.event_span(event) {
                 Some(resource_span) if self.is_resource(resource_span.metadata()) => {
                     let meta_id = event.metadata().into();
--- a/console-subscriber/src/lib.rs
+++ b/console-subscriber/src/lib.rs
@@ -222,7 +222,7 @@ impl TasksLayer {
 
         // Conservatively, start to trigger a flush when half the channel is full.
         // This tries to reduce the chance of losing events to a full channel.
-        let flush_under_capacity = config.event_buffer_capacity / 4;
+        let flush_under_capacity = config.event_buffer_capacity / 2;
 
         let server = Server {
             aggregator: Some(aggregator),
@@ -248,7 +248,7 @@ impl TasksLayer {
 }
 
 impl TasksLayer {
-    pub const DEFAULT_EVENT_BUFFER_CAPACITY: usize = 1024 * 100;
+    pub const DEFAULT_EVENT_BUFFER_CAPACITY: usize = 1024 * 10;
     pub const DEFAULT_CLIENT_BUFFER_CAPACITY: usize = 1024 * 4;
     pub const DEFAULT_PUBLISH_INTERVAL: Duration = Duration::from_secs(1);
 
@@ -406,7 +406,6 @@ where
     fn on_event(&self, event: &tracing::Event<'_>, ctx: Context<'_, S>) {
         let metadata = event.metadata();
         if self.waker_callsites.contains(event.metadata()) {
-            tracing::debug!(?event, "waker op event");
             let at = SystemTime::now();
             let mut visitor = WakerVisitor::default();
             event.record(&mut visitor);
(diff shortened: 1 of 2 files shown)


hawkw:
I think we probably also want to remove the code that _doesn't_ publish updates when flushing due to capacity, which may reduce update lag. I'm going to do that in a separate branch.

console-subscriber/src/lib.rs
```diff
@@ -222,7 +222,7 @@ impl TasksLayer {
 
         // Conservatively, start to trigger a flush when half the channel is full.
         // This tries to reduce the chance of losing events to a full channel.
-        let flush_under_capacity = config.event_buffer_capacity / 2;
+        let flush_under_capacity = config.event_buffer_capacity / 4;
```
seanmonstar: Conflicts with comment "half the channel is full".

To also help my mental model (feels like double negatives or something), this would now flush when there is only 1/4 of the capacity remaining. So the channel is 3/4 full. Right?

console-subscriber/src/lib.rs
```diff
@@ -406,6 +406,7 @@ where
     fn on_event(&self, event: &tracing::Event<'_>, ctx: Context<'_, S>) {
         let metadata = event.metadata();
         if self.waker_callsites.contains(event.metadata()) {
+            tracing::debug!(?event, "waker op event");
```
seanmonstar: Aren't these problematic for triggering an event inside an event handler?

pnkfelix:
Awesome, the change here seems to fix my issue!

console-subscriber/src/lib.rs
```diff
@@ -248,7 +248,7 @@ impl TasksLayer {
 }
 
 impl TasksLayer {
-    pub const DEFAULT_EVENT_BUFFER_CAPACITY: usize = 1024 * 10;
```
pnkfelix: was this change supposed to be part of this PR? In principle, the fix you put in should still work even without scaling up the default event-buffer capacity, right?

hawkw: i can back out those changes, yeah; i was just messing with things trying to figure out what was going on

console-subscriber/src/lib.rs
pnkfelix: (I believe Sean's reading is correct: it would now trigger a flush when channel is 3/4 full. But I also think Eliza is planning to remove this and some other unnecessary changes from this PR.)

console-subscriber/src/lib.rs
hawkw: whoops, meant to remove these!

pnkfelix (approved):
Awesome thanks!

seanmonstar (approved):

PR #182, opened 2021-11-08
Add license to Cargo.toml

daladim:
Thanks for these awesome crates!

Reading at your main LICENSE file, it looks you're using the MIT license, so I'm including them in your Cargo.toml files (using your crates would otherwise break the scanning of licenses that I'm using in my CI pipelines at home :D )


seanmonstar (approved):
Thanks :)

PR #188, opened 2021-11-23
feature: add resource detail view

zaharidichev:
This PR introduces the resoruce detail view. It also brings a couple of additional changes: 

- we now rely on context of entered spans rather than explicit parent - child relationships
- we expose internal resources (BatchSempahore)
- we show the relationships between resources (Mutex -> BatchSemaphore)
- we show the async ops that are live on a specific resource
- we also show the tasks that are awaiting on these async ops, which allows the user to draw a relationship between tasks and resources
- there are more examples added in the examples folder (semaphore and mutex)

Some screenshots: 

Screenshot 2021-11-23 at 19 12 28
Screenshot 2021-11-23 at 19 12 38

Screenshot 2021-11-23 at 19 12 50


hawkw:
This looks _amazing_. I'm going to try to give this branch a detailed review today, but the screenshots look awesome. Amazing work!

zaharidichev:
@hawkw Thanks a bunch for taking the time. Let me know after a quick glance whether you prefer me to break up the PR somehow or whether I can do anything to make it easier to review.

console-api/proto/async_ops.proto
```diff
@@ -37,6 +37,10 @@ message AsyncOp {
     // The source of this async operation. Most commonly this should be the name
     // of the method where the instantiation of this op has happened.
     string source = 3;
+    // The ID of the parent async op.
```
hawkw, line 44: nit: it might be nice if this comment explained a bit more what the "parent async op" of an async op is --- my guess is that this is for situations such as a `Sleep` inside of an `Interval`, or a `Semaphore` op inside a channel? but, it would be nice if the comment explained that.

also, if a given async op doesn't have a parent, this will not be set, right? it might be worth documenting that as well.

console-api/proto/common.proto
```diff
@@ -124,3 +124,14 @@ message PollStats {
     // has spent *waiting* to be polled.
     google.protobuf.Duration busy_time = 6;
 }
+
+// State attributes of an entity. These are dependent on the type of the entity.
+// For example, a timer resource will have a duration while a semaphore resource may
```
hawkw: nit:
```suggestion
// State attributes of an entity. These are dependent on the type of the entity.
//
// For example, a timer resource will have a duration, while a semaphore resource may
```

console-api/proto/common.proto
```diff
@@ -124,3 +124,14 @@ message PollStats {
...
+// have permits as an attribute. Likewise the async ops of a sempahore may have attributes
```
hawkw: nit/typo:
```suggestion
// have permits as an attribute. Likewise, the async ops of a semaphore may have attributes
```

console-subscriber/examples/barrier.rs
```diff
@@ -0,0 +1,40 @@
+use std::sync::Arc;
```
hawkw, line 1: thanks for adding new examples, these are great!

take it or leave it, but it might also be nice to have some examples that combine multiple resource types. that way, we could simulate what a real application that uses a large number of resources might look like. we could maybe stick similar code to these examples into `examples/app.rs` (maybe with a CLI option) so that the example can also spawn a number of tasks that use barriers, mutexen, etc.

it would be totally fine to make that change in a follow-up branch, though.

console-subscriber/examples/mutex.rs
```diff
@@ -0,0 +1,35 @@
...
+                            tokio::time::sleep(Duration::from_secs(1)).await;
+                        }
+                    });
+            }
+
+            loop {
+                if *count.lock().await >= 50 {
+                    break;
+                }
+            }
```
hawkw: nit, take it or leave it: this seems like it could be a `while` loop:
```suggestion
            while *count.lock().await < 50 {}
```

console-subscriber/src/aggregator/mod.rs
```diff
@@ -769,50 +783,105 @@ impl Aggregator {
...
-                let resource_id = self.ids.id_for(resource_id);
-                if let Some(mut stats) = self.resource_stats.update(&resource_id) {
-                    let field_name = match update.field.name.clone() {
-                        Some(name) => name,
+            Event::StateUpdate { target, update, .. } => {
+                let target_id = self.ids.id_for(target.target_id);
+                let mut targets_to_update = vec![(target_id, target.target_type.clone())];
+
+                fn update_entry(
+                    mut e: std::collections::hash_map::Entry<'_, FieldKey, Attribute>,
```
hawkw: nit: i would probably import `std::collections::hash_map` rather than spelling this out. not a blocker, though.

console-subscriber/src/aggregator/mod.rs
```diff
@@ -769,50 +783,105 @@ impl Aggregator {
...
+                    }
+                }
+
+                match target.target_type {
+                    TargetType::Resource => {
+                        if let Some(parent) = self
+                            .resources
+                            .get(&target_id)
+                            .and_then(|r| r.parent_id)
+                            .and_then(|parent_id| self.resources.get(&parent_id))
```
hawkw: nit, take it or leave it: i think we could combine this into one `and_then` like:
```suggestion
                            .and_then(|r| self.resources.get(r.parent_id.as_ref()?))
```

console-subscriber/src/aggregator/mod.rs
```diff
@@ -769,50 +783,105 @@ impl Aggregator {
...
+                        {
+                            targets_to_update.push((parent.id, TargetType::Resource));
+                        }
+                    }
+                    TargetType::AsyncOp => {
+                        if let Some(parent) = self
+                            .async_ops
+                            .get(&target_id)
+                            .and_then(|ao| ao.parent_id)
+                            .and_then(|parent_id| self.async_ops.get(&parent_id))
```
hawkw: similarly, i think this could be
```suggestion
nit, take it or leave it: i think we could combine this into one `and_then` like:
```suggestion
                            .and_then(|r| self.async_ops.get(r.parent_id.as_ref()?))
```
```

File: console-subscriber/src/aggregator/mod.rs [hunk 9]
```diff
@@ -769,50 +783,105 @@ impl Aggregator {
...
+                    upd: &AttributeUpdate,
+                ) {
+                    match e {
+                        Entry::Occupied(ref mut attr) => {
+                            update_attribute(attr.get_mut(), upd);
+                        }
+                        Entry::Vacant(attr) => {
+                            attr.insert(upd.clone().into());
+                        }
+                    }
```
hawkw: take it or leave it: this could also be written as:
```suggestion
                    e
                        .and_modify(|attr| update_attribute(attr.get_mut(), upd))
                        .or_insert_with(|| upd.clone().into());
```

console-subscriber/src/lib.rs
```diff
@@ -65,20 +65,30 @@ pub struct TasksLayer {
...
+    /// Set of callsites for events representing state attribute state updates on resources
     ///
     /// TODO: Take some time to determine more reasonable numbers
-    state_update_callsites: Callsites<32>,
+    resource_state_update_callsites: Callsites<32>,
+
+    /// Set of callsites for events representing state attribute state updates on async resource ops
+    ///
+    /// TODO: Take some time to determine more reasonable numbers
+    async_op_state_update_callsites: Callsites<32>,
```
hawkw: at this point we care about enough different sets of callsites that i wonder if it might be worth having one array of call the callsites that stores what kind of callsite it is...but, we would need to look at the current usage patterns to determine whether that would be better or worse. anyway, probably not something we want to address in this PR.

console-subscriber/src/lib.rs
```diff
@@ -135,35 +147,50 @@ enum Event {
...
     Resource {
         id: span::Id,
+        parent_id: Option<span::Id>,
         metadata: &'static Metadata<'static>,
         at: SystemTime,
         concrete_type: String,
         kind: resource::Kind,
         location: Option<proto::Location>,
+        is_internal: bool,
+        inherit_child_attrs: bool,
```
hawkw, line 157: it might be worth adding comments explaining what some of these fields mean, for future readers? not a blocker though.

console-subscriber/src/lib.rs
```diff
@@ -135,35 +147,50 @@ enum Event {
...
+enum TargetType {
+    Resource,
+    AsyncOp,
+}
+
+#[derive(Debug, Clone)]
+struct Target {
+    target_id: span::Id,
+    target_type: TargetType,
+}
```
hawkw: i'm not the _biggest_ fan of this naming..."target" has a specific meaning in `tracing` terminology, and technically the `TargetType` enum represents the event's `tracing` target. what about maybe just calling this `StateUpdate`/`StateUpdateType`?

console-subscriber/src/lib.rs
```diff
@@ -424,56 +494,77 @@ where
...
-                            metadata,
-                            at,
-                            resource_id: resource_span.id(),
-                            update,
-                        });
-                    }
+        } else if self
+            .resource_state_update_callsites
+            .contains(event.metadata())
+        {
```
hawkw: that...certainly is some formatting...

i wonder if this `if` chain could be made a little nicer using early returns? not a blocker though.

console-subscriber/src/lib.rs
hawkw: also, AFAICT, the `Target` struct seems to only be used as a field on `StateUpdate`...is it really necessary for this to be a separate struct, or can its fields just be part of `StateUpdate`?

console-subscriber/src/visitors.rs
```diff
@@ -152,10 +160,13 @@ pub(crate) struct StateUpdateVisitor {
...
     pub(crate) const RES_SPAN_NAME: &'static str = "runtime.resource";
     const RES_CONCRETE_TYPE_FIELD_NAME: &'static str = "concrete_type";
+    const RES_VIZ_FIELD_NAME: &'static str = "is_internal";
     const RES_KIND_FIELD_NAME: &'static str = "kind";
     const RES_KIND_TIMER: &'static str = "timer";
 
-    pub(crate) fn result(self) -> Option<(String, resource::Kind, Option<proto::Location>)> {
+    pub(crate) fn result(
+        self,
+    ) -> Option<(String, resource::Kind, Option<proto::Location>, bool, bool)> {
```
hawkw: this is a _big_ tuple...at this point, it really feels like maybe this should be returning a struct with named fields instead?

console-subscriber/src/lib.rs
```diff
@@ -424,56 +494,77 @@ where
...
+            if let Some(resource_id) = resource_id {
+                let mut poll_op_visitor = PollOpVisitor::default();
+                event.record(&mut poll_op_visitor);
+                if let Some((op_name, is_ready)) = poll_op_visitor.result() {
+                    let task_and_async_op_ids = self.current_spans.get().and_then(|stack| {
+                        let stack = stack.borrow();
+                        let task_id =
+                            self.first_entered(&stack, |id| self.is_id_spawned(id, &ctx))?;
+                        let async_op_id =
+                            self.first_entered(&stack, |id| self.is_id_async_op(id, &ctx))?;
```
hawkw, line 519: minor nit, something we can always do in a follow-up: it would be nice if we could find both the task ID _and_ the async op ID in one iteration over the stack, instead of iterating twice...these stacks are generally not going to be so deeply nested that this is _really_ important, but it would be better to only iterate once, i think...

console-subscriber/src/visitors.rs
```diff
@@ -22,15 +23,21 @@ const LOCATION_COLUMN: &str = "loc.col";
...
 ///     kind = "timer",
+///     is_internal = true,
+///     inherits_child_attrs = true,
 /// );
 ///
 /// Fields:
 /// concrete_type - indicates the concrete rust type for this resource
 /// kind - indicates the type of resource (i.e. timer, sync, io )
+/// is_internal - whether this is a resource type that is not exposed publicly (i.e. BatchSemaphore)
+/// inherits_child_attrs - whether this resource should inherit the state attributes of its children
```
hawkw, line 34: hmm, the fact that we have a sort of magic field that controls how the console displays data is...not exactly my favorite thing? it seems like in an ideal world, the tracing events would _just_ expose data, and the console would be able to determine how it should be displayed, and this kind of feels like a violation of that idea. but, i also get that maybe this is necessary to get the data to display nicely.

is there any other potential solution that would let us get the UI we want without having the runtime annotate its events with this field? because...that would certainly be nice.

console-subscriber/src/visitors.rs
hawkw: this tuple feels big enough that it should _probably_ be a struct...

console/src/state/async_ops.rs
```diff
@@ -0,0 +1,308 @@
+use crate::{
```
hawkw, line 1: this file looks fine overall, but also, it does occur to me that we're really starting to reach the point where there's a _lot_ of very similar boilerplate for different parts of the data model (tasks, resources, async ops...). it would maybe be worth thinking about whether some of this code, like the sorting by fields, could be made generic so we don't have to write identical boilerplate quite so many times.

but, that's not really a problem for this branch, just something we might want to think about addressing later...

console/src/state/mod.rs
```diff
@@ -386,6 +421,35 @@ impl FieldValue {
...
+        let val_style = styles.fg(Color::Yellow);
+        let unit_style = styles.fg(Color::LightBlue);
+
+        let mut formatted = Vec::with_capacity(attributes.len());
+        let attributes = attributes.iter();
+        for attr in attributes {
+            let mut elems = vec![
+                Span::styled(attr.field.name.to_string(), key_style),
+                Span::styled("=", delim_style),
+                Span::styled(format!("{}", attr.field.value), val_style),
```
hawkw, line 436: we may want to think about only re-formatting this string when the field value _changes_, rather than every time we render a screen...

console/src/view/async_ops.rs
```diff
@@ -0,0 +1,244 @@
...
+                .iter()
+                .filter_map(move |async_op| {
+                    let async_op = async_op.upgrade()?;
+                    let async_op = async_op.borrow();
+
+                    let parent_id_str = match async_op.parent_id() {
+                        Some(id) => {
+                            format!("{}", id)
+                        }
+                        None => "n/a".to_string(),
```
hawkw: it weould be nice to avoid re-formatting and allocating these strings every time...

console/src/view/async_ops.rs
```diff
@@ -0,0 +1,244 @@
...
+                    let task_id = async_op.task_id()?;
+                    let task_name = state
+                        .tasks_state()
+                        .task(task_id)
+                        .and_then(|t| t.upgrade())
+                        .and_then(|t| t.borrow().name().map(String::from));
+
+                    let task_str = match task_name {
+                        Some(name) => format!("{} ({})", task_id, name),
+                        None => format!("{}", task_id),
```
hawkw: similarly, these could probably be formatted once, i imagine?

console/src/view/resource.rs
```diff
@@ -0,0 +1,150 @@
...
+            Span::raw(match resource.parent_id() {
+                Some(id) => state
+                    .resources_state()
+                    .resource(id)
+                    .and_then(|r| r.upgrade())
+                    .map(|r| {
+                        let r = r.borrow();
+                        format!("{} ({}::{})", r.id(), r.target(), r.concrete_type())
+                    })
+                    .unwrap_or(format!("{}", id)),
```
hawkw: ids, targets, and concrete types aren't going to change, so it would be nice to be able to avoid formatting these strings every time.

console/src/view/resource.rs
```diff
@@ -0,0 +1,150 @@
...
+            bold(styles.if_utf8("\u{238B} esc", "esc")),
+            Span::raw(" = return to task list, "),
+            bold("q"),
+            Span::raw(" = quit"),
+        ]);
+
+        let mut overview = Vec::with_capacity(6);
+        overview.push(Spans::from(vec![
+            bold("ID: "),
+            Span::raw(format!("{} ", resource.id())),
```
hawkw: it would be nice not to format this on every frame...

console/src/view/resource.rs
```diff
@@ -0,0 +1,150 @@
...
+                        format!("{} ({}::{})", r.id(), r.target(), r.concrete_type())
+                    })
+                    .unwrap_or(format!("{}", id)),
+                None => "n/a".to_string(),
+            }),
+        ]));
+
+        overview.push(Spans::from(vec![
+            bold("Kind: "),
+            Span::raw(format!("{} ", resource.kind())),
```
hawkw: can we store the kind string as well?

console/src/view/resource.rs
```diff
@@ -0,0 +1,150 @@
...
+        ]));
+
+        overview.push(Spans::from(vec![
+            bold("Target: "),
+            Span::raw(resource.target()),
+        ]));
+
+        overview.push(Spans::from(vec![
+            bold("Type: "),
+            Span::raw(format!("{} ", resource.concrete_type())),
```
hawkw: ...and the concrete type string?

console/src/view/resources.rs
```diff
@@ -23,13 +23,16 @@ pub(crate) struct ResourcesTable {}
...
+    type Context = ();
 
     const HEADER: &'static [&'static str] = &[
         "ID",
+        "Parent",
         "Kind",
         "Total",
         "Target",
         "Type",
+        "Viz",
```
hawkw: i have a slight preference for spelling this "Vis", as it's short for "Visibility"...or maybe it should just be "pub"? but, i could be convinced otherwise...

console/src/view/resources.rs
```diff
@@ -75,12 +84,20 @@ impl TableList for ResourcesTable {
                     let resource = resource.upgrade()?;
                     let resource = resource.borrow();
 
+                    let parent_id_str = match resource.parent_id() {
+                        Some(id) => {
+                            format!("{}", id)
+                        }
+                        None => "n/a".to_string(),
```
hawkw: another string that shouldn't have to be `format!`ed every time we render a frame...

hawkw (commented):
overall, this seems pretty good --- and the UI screenshots you posted are awesome! i had a bunch of minor suggestions.

also, i wonder if it would be worth adding a way to hide/show all the internal resources---and maybe hide them by default? i don't know if we _really_ want to expose tokio internals like the batch semaphore by default, unless the user _asks_ for them, it seems like it could potentially just be a bunch of information that's confusing to a user who isn't familiar with tokio's internals. but, we could address that in a separate PR.

console-subscriber/src/visitors.rs
zaharidichev: @hawkw If you are referring to `inherits_child_attrs` I totally agree. The problem is that in some scenarios we want to propagate these attrs to the parent resource/async op. For example in the situation of Semaphore/BatchSempahore. Most of the instrumentation points that update state are happening in the BatchSempahore but we want the same data on the public Semaphore as it will be a bit hard to instrument that one properly. One possible solution that will most likely yield a similar result to what we have at the moment is to make these state attrs always be inherited from child resources that are flagged as internal. In that scenario however, we will have the `Mutex` resource inherit the state attrs of the internal Semaphore that it uses. Of course we can think of some way to differentiate them in the UI to make it clear that these are attrs of an internal resource but again... this can become confusing. Do you have any better ideas?

hawkw: > The problem is that in some scenarios we want to propagate these attrs to the parent resource/async op. For example in the situation of Semaphore/BatchSempahore. Most of the instrumentation points that update state are happening in the BatchSempahore but we want the same data on the public Semaphore as it will be a bit hard to instrument that one properly. 

yeah, i get the motivation for this, and it is entirely possible that it _is_ the best solution! 

one thing i might consider trying is a global toggle in the resources view that sets whether or not private resources are displayed. when private resources are turned on, we would show those resources in the resource list _and_ display any child attributes from private resources on the parent resources. when the toggle is turned off, we would hide both the private resources from the resource list, _and_ hide their attributes from parent resources. it could be worth playing with that and seeing if it's an acceptable solution to this problem?

if a toggle in the UI doesn't adequately solve the problem, i'd be okay with moving forward with this approach; it just seems worth trying?

hawkw: another thing to consider is, maybe we just always have parents inherit child attributes if they have no attributes of their own, regardless of whether or not the toggle is set?

console/src/view/resource.rs
zaharidichev: In these situations do you mean storing the string as an `InternetStr` on the `Resource/AsyncOp` objects?

hawkw: Depending on the string, we may or may not want to intern them. In the case of `resource.kind()`, I believe they already _are_ `InternedStr`s: https://github.com/tokio-rs/console/blob/042c1f78447839bed54538ed89b9277e437ed6b4/console/src/state/resources.rs#L33-L37

I think we could probably add a method to `resources::Kind` that returns an `&'static str`, so that we don't need to `format!` it here. I can open a quick PR for that change?

hawkw: ~~...actually, i just double-checked, and `resource.kind()` _already_ returns an `&str`. so i think this can just be~~
```suggestion
            Span::raw(resource.kind()),
```


**Edit**: oh, hmm, that doesn't actually work due to lifetime issues. i think we may need to make some additional changes to `InternedStr` to make this work nicely...

console-api/proto/async_ops.proto
```diff
@@ -37,6 +37,13 @@ message AsyncOp {
     // The source of this async operation. Most commonly this should be the name
     // of the method where the instantiation of this op has happened.
     string source = 3;
+    // The ID of the parent async op. This field is set when the this async op
+    // is created in the context of another one. An example of that is for example
+    // a `Mutex::lock` which internally calls `Semaphore::acquire`. This field can
+    // be empty.
```
hawkw: some minor nits here:
```suggestion
    // The ID of the parent async op. 
    //
    // This field is only set if this async op was created while inside of another
    // async op.  For example, `tokio::sync`'s `Mutex::lock` internally calls 
    // `Semaphore::acquire`. 
    // 
    // This field can be empty; if it is empty, this async op is not a child of another
    // async op.
```

console-api/proto/common.proto
```diff
@@ -124,3 +124,15 @@ message PollStats {
     // has spent *waiting* to be polled.
     google.protobuf.Duration busy_time = 6;
 }
+
+// State attributes of an entity. These are dependent on the type of the entity.
+//
+// For example, a timer resource will have a duration, while a semaphore resource may
+// have permits as an attribute. Likewise the async ops of a semaphore may have attributes
```
hawkw:
```suggestion
// have a permit count. Likewise, the async ops of a semaphore may have attributes
```

console-subscriber/src/aggregator/mod.rs
```diff
@@ -770,49 +784,98 @@ impl Aggregator {
...
+                    UpdateType::AsyncOp => {
+                        if let Some(parent) = self
+                            .async_ops
+                            .get(&update_id)
+                            .and_then(|r| self.async_ops.get(r.parent_id.as_ref()?))
+                            .filter(|parent| parent.inherit_child_attrs)
+                        {
+                            to_update.push((parent.id, UpdateType::AsyncOp));
+                        }
+                    }
```
hawkw, line 808: i think in theory there might be a way to not repeat this code quite as much, but i don't know if it's worth the effort of figuring that out. we could maybe have a function that takes a hash map and performs the lookup part? but, not a blocker.

console/src/state/async_ops.rs
```diff
@@ -34,7 +34,7 @@ pub(crate) enum SortBy {
 #[derive(Debug)]
 pub(crate) struct AsyncOp {
     id: u64,
-    parent_id: Option<u64>,
+    parent_id: InternedStr,
```
hawkw, line 37: i think the parent ID could probably just be a normal `String` rather than an `InternedStr`, unless we expect a single async op to have a _lot_  of children. but, we'll clean up unused interned strings anyway so this is fine in either case.

console/src/state/resources.rs
```diff
@@ -158,22 +166,53 @@ impl ResourcesState {
...
+                    })
+                    .unwrap_or(format!("{}", id)),
+                None => "n/a".to_string(),
+            });
+
+            let parent_id = strings.string(match parent_id {
+                Some(id) => {
+                    format!("{}", id)
+                }
+                None => "n/a".to_string(),
```
hawkw: nit, take it or leave it: this could maybe be simplified to
```suggestion
            let parent_id = strings.string(parent_id
                .map(u64::to_string)
                .unwrap_or_else(|| "n/a".to_string());
            );
```

console/src/state/resources.rs
```diff
@@ -158,22 +166,53 @@ impl ResourcesState {
...
+
+            let parent = strings.string(match parent_id {
+                Some(id) => parents
+                    .get(&id)
+                    .and_then(|r| r.upgrade())
+                    .map(|r| {
+                        let r = r.borrow();
+                        format!("{} ({}::{})", r.id(), r.target(), r.concrete_type())
+                    })
+                    .unwrap_or(format!("{}", id)),
```
hawkw: nit: this should probably be 
```suggestion
                    .unwrap_or_else(|| format!("{}", id)),
```
so that we don't format the string if we _did_ find a parent (see the documentation for https://doc.rust-lang.org/stable/core/option/enum.Option.html#method.unwrap_or)

hawkw: also, it might be _slightly_ more efficient not to go through the `format!` macro when the format arguments are just `"{}"`; we can use the `fmt::Display` implementation for `u64` directly:
```suggestion
                    .unwrap_or_else(|| id.to_string()),
```

but, that's probably not a _huge_ performance impact, regardless.

console/src/state/async_ops.rs
```diff
@@ -0,0 +1,321 @@
...
+        let total = dropped_at.map(|d| d.duration_since(created_at).unwrap());
+
+        let poll_stats = pb.poll_stats.expect("task should have poll stats");
+        let busy = poll_stats.busy_time.map(pb_duration).unwrap_or_default();
+        let idle = total.map(|total| total - busy);
+        let formatted_attributes = Attribute::make_formatted(styles, &mut attributes);
+        let task_id = pb.task_id.map(|id| id.id);
+        let task_id_str = strings.string(
+            task_id
+                .map(|tid| format!("{}", tid))
```
hawkw: nit: could be 
```suggestion
                .map(u64::to_string)
```
which might be very slightly more efficient (but not a big deal)

console/src/view/async_ops.rs
```diff
@@ -114,24 +114,16 @@ impl TableList for AsyncOpsTable {
...
-                        .and_then(|t| t.borrow().name().map(String::from));
+                        .map(|t| t.borrow().short_desc().to_owned());
 
-                    let task_str = match task_name {
-                        Some(name) => format!("{} ({})", task_id, name),
-                        None => format!("{}", task_id),
+                    let task_str = match task {
+                        Some(task_str) => task_str,
+                        None => async_op.task_id_str().to_owned(),
                     };
```
hawkw: nit/tioli: could just be
```suggestion
                    let task_str = task.unwrap_or_else(||  async_op.task_id_str().to_owned());
```

hawkw (approved):
overall, this looks pretty good to me and i think we should merge this and address the bigger potential improvements in follow-up branches. i commented on a few minor suggestions we may want to address first.

console/src/state/async_ops.rs
zaharidichev: @hawkw I actually wonder whether we can take a different approach with all the parent info. Namely, what if whenever we have a parent we simply have an `Option<AsyncOpRef> `or  `Option<ResourceRef>`  and then get all the needed strings, etc by upgrading the parent ref? That will make it much easier when we have a toggle that determines whether we should display any info about internal resources. WDYT?

console-subscriber/src/aggregator/mod.rs
pnkfelix: Alternatively, you could have a method that takes an `UpdateType` and gives you back the hashmap, and then leave the currently duplicated code inline as a single non-duplicated `if` statement. (This works in part because you can just pass along the `UpdateType` into the `to_update.push` invocation, rather than writing out the specific variant the way you are doing here.

console/src/state/async_ops.rs
hawkw: We could definitely do something like that; it might be worth doing in a follow-up branch so that we can get this PR merged sooner?

console-subscriber/src/aggregator/mod.rs
hawkw: yeah, that's more or less what i was imagining. i think the `filter` might have to be kept separate, as well, because the field access is being performed on a different type. we could introduce a trait for that, but, hmm...

console/src/state/async_ops.rs
zaharidichev: 👍

hawkw:
@zaharidichev is this ready to merge or are you planning on making an additional changes in this branch?

zaharidichev:
I can address more of the feedback in follow-ups. Its already getting quite big.  I think the corresponding tokio branch needs to be sanity checked before we merge this though:  https://github.com/tokio-rs/tokio/pull/4302

carllerche:
@hawkw @zaharidichev to confirm: this PR is ready to merge pending the Tokio PR?

carllerche:
@zaharidichev Tokio 1.15 is out.

console-api/proto/common.proto
```diff
@@ -183,3 +183,15 @@ message PollStats {
...
+//
+// For example, a timer resource will have a duration, while a semaphore resource may
+// have a permit count. Likewise, the async ops of a semaphore may have attributes
+// indicating how many permits they are trying to acquire vs how many are acquired.
+// These values may change over time. Therefore, they live in the runtime stats rather
+// than the static data describing the entity.
+message Attribute {
+    Field field = 1;
+    optional string unit = 2;
+}
```
hawkw: just adding the docs addition from @pnkfelix's PR:
```suggestion
    message Attribute {
        // The key-value pair for the attribute
        common.Field field = 1;
        // Some values carry a unit of measurement. For example, a duration
        // carries an associated unit of time, such as "ms" for milliseconds.
        optional string unit = 2;
    }
```

console-api/proto/resources.proto
```diff
@@ -41,6 +41,9 @@ message Resource {
     Kind kind = 4;
     // The location in code where the resource was created.
     common.Location location = 5;
+    // The ID of the parent resource.
+    common.Id parent_resource_id = 6;
+    bool is_internal = 7;
```
hawkw, line 50:
```suggestion
    // Is the resource an internal component of another resource?
    //
    // For example, a `tokio::time::Interval` resource might contain a 
    // `tokio::time::Sleep` resource internally.
    bool is_internal = 7;
```

hawkw (approved):
added docs

console-api/proto/common.proto
```diff
@@ -183,3 +183,18 @@ message PollStats {
...
+// indicating how many permits they are trying to acquire vs how many are acquired.
+// These values may change over time. Therefore, they live in the runtime stats rather
+// than the static data describing the entity.
+    message Attribute {
+        // The key-value pair for the attribute
+        common.Field field = 1;
+        // Some values carry a unit of measurement. For example, a duration
+        // carries an associated unit of time, such as "ms" for milliseconds.
+        optional string unit = 2;
+    }
```
hawkw: oh whoops my copy-paste indented this too much:
```suggestion
message Attribute {
    // The key-value pair for the attribute
    common.Field field = 1;
    // Some values carry a unit of measurement. For example, a duration
    // carries an associated unit of time, such as "ms" for milliseconds.
    optional string unit = 2;
}
```

PR #195, opened 2021-12-15
Attempt to unify build, init, and the Layer system

### LINKED ISSUE #183: Specify a custom port when serving
Just a simple option for serving the server on a custom port. Useful for docker containers with specific ports needing to be used.

asonix:
In reference to a conversation I observed on discord, this is an attempt to make the `build` and `init` methods play nicer with the existing `tracing_subscriber` layer system.

Changes:
- the `Builder` struct has it's own `.build_console_layer()` and `.init()` methods now to produce the layer or install a subscriber directly from a configured builder
- `build` is now a wrapper around `Builder::build_console_layer`
- `init` is now a wrapper around `Builder::init`

Closes #183 
Closes #196


seanmonstar:
Hey, it looks like we were both working on it at the same time! Thanks for the PR! For comparison, here's the direction I was thinking of (doesn't mean it's the right direction): https://github.com/tokio-rs/console/pull/196

asonix:
I brought this PR more in line with yours @seanmonstar by deleting the `init.rs` file and including everything directly in the builder file, So at this point the main difference is the `builder` free function in yours and the naming of `build_console_layer` vs `spawn`, and the return type of that method.

console-subscriber/src/builder.rs
```diff
@@ -149,6 +159,279 @@ impl Builder {
...
+    ///
+    /// ## Configuration
+    ///
+    /// Tokio console subscriber is configured with sensible defaults for most
+    /// use cases. If you need to tune these parameters, several environmental
+    /// configuration variables are available:
+    ///
+    /// | **Environment Variable**            | **Purpose**                                                               | **Default Value** |
+    /// |-------------------------------------|---------------------------------------------------------------------------|-------------------|
+    /// | `TOKIO_CONSOLE_RETENTION_SECS`      | The number of seconds to accumulate completed tracing data                | 3600s (1h)        |
```
pnkfelix: I know this isn't your mistake (since you presumably just cut-and-pasted the table from the `init` docs), but there are typos in the environment variable names from the `init` docs.

See this commit for how I addressed it in a parallel PR: https://github.com/tokio-rs/console/pull/198/commits/aad08331a8bb90c03ba9b2c3d0c893413f65151d

console-subscriber/src/builder.rs
```diff
@@ -149,6 +159,279 @@ impl Builder {
...
+///
+/// ## Configuration
+///
+/// Tokio console subscriber is configured with sensible defaults for most
+/// use cases. If you need to tune these parameters, several environmental
+/// configuration variables are available:
+///
+/// | **Environment Variable**            | **Purpose**                                                               | **Default Value** |
+/// |-------------------------------------|---------------------------------------------------------------------------|-------------------|
+/// | `TOKIO_CONSOLE_RETENTION_SECS`      | The number of seconds to accumulate completed tracing data                | 3600s (1h)        |
```
pnkfelix: Same issue applies to this table that I put into another comment: https://github.com/tokio-rs/console/pull/195/files#r769753155

console-subscriber/src/builder.rs
```diff
@@ -149,6 +159,279 @@ impl Builder {
...
+    /// tracing_subscriber::registry()
+    ///     .with(console_layer)
+    ///     .with(tracing_subscriber::fmt::layer())
+    /// //  .with(..potential additional layer..)
+    ///     .init();
+    /// ```
+    ///
+    /// [`Targets`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/filter/struct.Targets.html
+    pub fn init(self) {
+        let fmt_filter = std::env::var("RUST_LOG")
```
hawkw, line 221: what do you think about making the filter environment variable a builder parameter? so, by default, it uses `RUST_LOG`, but there's a separate builder method to change the filter env var? something like:

```rust
let console = console_subscriber::Builder::new()
    .with_filter_env_variable("MYAPP_LOG")
    .build_console_layer();
```

console-subscriber/src/builder.rs
```diff
@@ -149,6 +159,279 @@ impl Builder {
...
+        let console_layer = self.build_console_layer();
+
+        tracing_subscriber::registry()
+            .with(console_layer)
+            .with(tracing_subscriber::fmt::layer().with_filter(fmt_filter))
+            .init();
+    }
+
+    /// Returns a new `tracing` [subscriber] configured with a [`TasksLayer`]
+    /// and a [filter] that enables the spans and events required by the console.
```
hawkw: this no longer returns a subscriber:
```suggestion
    /// Returns a new `tracing` [`Layer`] consisting of a [`TasksLayer`]
    /// and a [filter] that enables the spans and events required by the console.
```

console-subscriber/src/builder.rs
```diff
@@ -149,6 +159,279 @@ impl Builder {
...
+    ///     .init();
+    /// ```
+    /// [`.init()`]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/util/trait.SubscriberInitExt.html
+    /// [default]: https://docs.rs/tracing/latest/tracing/dispatcher/index.html#setting-the-default-subscriber
+    /// [sub]: https://docs.rs/tracing/latest/tracing/trait.Subscriber.html
+    /// [`tracing_subscriber::fmt`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/fmt/index.html
+    /// [`fmt::Layer`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/fmt/struct.Layer.html
+    /// [`console_subscriber::init`]: crate::init()
+    #[must_use = "build_console_layer() without init() will not set the default tracing subscriber"]
+    pub fn build_console_layer(self) -> ConsoleLayer {
```
hawkw: i think i prefer the `spawn` naming from #196, since one of the main differences between this and the `build` method is that it spawns the console's server on a background runtime...

console-subscriber/src/builder.rs
```diff
@@ -149,6 +159,279 @@ impl Builder {
...
+    /// //  .with(...)
+    ///     .init();
+    /// ```
+    /// [`.init()`]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/util/trait.SubscriberInitExt.html
+    /// [default]: https://docs.rs/tracing/latest/tracing/dispatcher/index.html#setting-the-default-subscriber
+    /// [sub]: https://docs.rs/tracing/latest/tracing/trait.Subscriber.html
+    /// [`tracing_subscriber::fmt`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/fmt/index.html
+    /// [`fmt::Layer`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/fmt/struct.Layer.html
+    /// [`console_subscriber::init`]: crate::init()
+    #[must_use = "build_console_layer() without init() will not set the default tracing subscriber"]
```
hawkw: this no longer returns something that we can call `init` on, so perhaps we should change this:
```suggestion
    #[must_use = "a `Layer` must be added to a `tracing::Subscriber` in order to be used"]
```
or something like that?

console-subscriber/src/builder.rs
```diff
@@ -149,6 +159,279 @@ impl Builder {
...
+///     .init();
+/// ```
+/// [`.init()`]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/util/trait.SubscriberInitExt.html
+/// [default]: https://docs.rs/tracing/latest/tracing/dispatcher/index.html#setting-the-default-subscriber
+/// [sub]: https://docs.rs/tracing/latest/tracing/trait.Subscriber.html
+/// [`tracing_subscriber::fmt`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/fmt/index.html
+/// [`fmt::Layer`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/fmt/struct.Layer.html
+/// [`console_subscriber::init`]: crate::init()
+#[must_use = "build() without init() will not set the default tracing subscriber"]
+pub fn build() -> ConsoleLayer {
```
hawkw: maybe we should rename this to just `layer` or something? :woman_shrugging:

console-subscriber/src/builder.rs
```diff
@@ -149,6 +159,279 @@ impl Builder {
...
+    ///
+    /// In addition to the [`TasksLayer`], which collects instrumentation data
+    /// consumed by the console, the default [`Subscriber`][sub] initialized by this
+    /// function also includes a [`tracing_subscriber::fmt`] layer, which logs
+    /// tracing spans and events to stdout. Which spans and events are logged will
+    /// be determined by the `RUST_LOG` environment variable.
+    ///
+    /// **Note**: this function sets the [default `tracing` subscriber][default]
+    /// for your application. If you need to add additional layers to this
+    /// subscriber, see [`build`].
```
hawkw: this should probably refer to the new `build_console_layer` method now (which i think we may want to rename to `spawn` or something...)

hawkw (commented):
Comparing this PR with @seanmonstar's PR #196, there are some things I like better about both branches --- it would be nice if we could end up merging something that includes the desirable aspects of both proposals.

I like that this branch has the builder functions return a normal `tracing` layer, unlike what #196 does. In #196, the builder will instead return a `tracing` subscriber with a layer already added to it, which I'm not a fan of; returning a `Layer` makes it much easier to integrate the console with existing code for setting up a `tracing` subscriber with other layers. It also makes it much easier to make the console support optional --- the returned `Layer` can be inside an `Option`, and only be `Some` if the console is enabled. This composes much nicer with the rest of `tracing`, so I'd strongly prefer this approach over the one in #196. 

I'm not a big fan of the naming in this branch --- the relationship between the existing `build` method and the `build_console_layer` method is not particularly clear. I think I prefer the `spawn` naming from #196.

I commented on some minor docs nits. Overall, though, the approach feels right.

console-subscriber/src/builder.rs
hawkw: (it would be fine --- or perhaps preferable! --- to add that in a follow-up PR)

console-subscriber/src/builder.rs
```diff
@@ -149,6 +159,277 @@ impl Builder {
...
+    ///     .init();
+    /// ```
+    /// [`.init()`]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/util/trait.SubscriberInitExt.html
+    /// [default]: https://docs.rs/tracing/latest/tracing/dispatcher/index.html#setting-the-default-subscriber
+    /// [sub]: https://docs.rs/tracing/latest/tracing/trait.Subscriber.html
+    /// [`tracing_subscriber::fmt`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/fmt/index.html
+    /// [`fmt::Layer`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/fmt/struct.Layer.html
+    /// [`console_subscriber::init`]: crate::init()
+    #[must_use = "a `Layer` must be added to a `tracing::Subscriber` in order to be used"]
+    pub fn spawn(self) -> ConsoleLayer {
```
seanmonstar: Instead of a concrete type, maybe we could return an `impl Layer` or similar? That way we can add or change layers and not break... Just a thought, whatcha think @hawkw?

asonix: I'm returning `impl Layer<S>` now but it required a `for<'a> Trait<'a>` bound on `S`. It's probably fine but wanted to point that out

console-subscriber/src/builder.rs
```diff
@@ -149,6 +158,283 @@ impl Builder {
...
+    /// [`.init()`]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/util/trait.SubscriberInitExt.html
+    /// [default]: https://docs.rs/tracing/latest/tracing/dispatcher/index.html#setting-the-default-subscriber
+    /// [sub]: https://docs.rs/tracing/latest/tracing/trait.Subscriber.html
+    /// [`tracing_subscriber::fmt`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/fmt/index.html
+    /// [`fmt::Layer`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/fmt/struct.Layer.html
+    /// [`console_subscriber::init`]: crate::init()
+    #[must_use = "a `Layer` must be added to a `tracing::Subscriber` in order to be used"]
+    pub fn spawn<S>(self) -> impl Layer<S>
+    where
+        S: Subscriber + for<'a> LookupSpan<'a>,
```
hawkw, line 298: the impl layer is much nicer :+1:

console-subscriber/src/builder.rs
```diff
@@ -149,6 +158,283 @@ impl Builder {
...
+    ///
+    /// tracing_subscriber::registry()
+    ///     .with(console_layer)
+    ///     .with(tracing_subscriber::fmt::layer())
+    /// //  .with(..potential additional layer..)
+    ///     .init();
+    /// ```
+    ///
+    /// [`Targets`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/filter/struct.Targets.html
+    pub fn init(self) {
```
hawkw, line 220: nit/tioli: it would also maybe be nice to have a `try_init` variant that returns `Result` rather than panicking? but we can add that in a follow-up PR.

console-subscriber/src/builder.rs
```diff
@@ -149,6 +158,283 @@ impl Builder {
...
+///
+/// let console_layer = console_subscriber::spawn();
+///
+/// tracing_subscriber::registry()
+///     .with(console_layer)
+///     .with(tracing_subscriber::fmt::layer())
+/// //  .with(..potential additional layer..)
+///     .init();
+/// ```
+///
```
hawkw, line 395: in general, i find that when there's a simple entry-point functions that just calls a couple other functions in the public API, it's often useful to explicitly document what that function does, so that users can easily reproduce the default configuration when switching to other APIs. so, we could maybe add something like this:
```suggestion
///
/// Calling `console_subscriber::init` is equivalent to the following:
/// ```rust
/// use console_subscriber::TasksLayer;
///
/// TasksLayer::builder().with_default_env().init();
/// ```
```

console-subscriber/src/builder.rs
```diff
@@ -149,6 +158,283 @@ impl Builder {
...
+
+/// Returns a new `tracing_subscriber` [`Layer`] configured with a [`TasksLayer`]
+/// and a [filter] that enables the spans and events required by the console.
+///
+/// This function spawns the console subscriber's [`Server`] in its own Tokio
+/// runtime in a background thread.
+///
+/// Unlike [`init`], this function does not set the default subscriber, allowing
+/// additional [`Layer`]s to be added.
+///
```
hawkw, line 415: similarly, it might be worth documenting something like
```suggestion
///
/// This function is equivalent to the following:
/// ```
/// use console_subscriber::TasksLayer;
///
/// let layer = TasksLayer::builder().with_default_env().spawn();
/// # drop(layer); // to suppress must_use warnings
/// ```
```

console-subscriber/src/builder.rs
```diff
@@ -149,6 +158,283 @@ impl Builder {
...
+/// //  .with(...)
+///     .init();
+/// ```
+/// [`.init()`]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/util/trait.SubscriberInitExt.html
+/// [default]: https://docs.rs/tracing/latest/tracing/dispatcher/index.html#setting-the-default-subscriber
+/// [sub]: https://docs.rs/tracing/latest/tracing/trait.Subscriber.html
+/// [`tracing_subscriber::fmt`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/fmt/index.html
+/// [`fmt::Layer`]: https://docs.rs/tracing-subscriber/latest/tracing-subscriber/fmt/struct.Layer.html
+/// [`console_subscriber::init`]: crate::init()
+#[must_use = "spawn() without init() will not set the default tracing subscriber"]
```
hawkw: should maybe change this as well:
```suggestion
#[must_use = "a `Layer` must be added to a `tracing::Subscriber`in order to be used"]
```

console-subscriber/src/builder.rs
asonix: I renamed it to `spawn` as well but maybe another name would be preferable

console-subscriber/src/builder.rs
```diff
@@ -149,6 +158,283 @@ impl Builder {
...
+    /// runtime in a background thread.
+    ///
+    /// Unlike [`init`], this function does not set the default subscriber, allowing
+    /// additional [`Layer`]s to be added.
+    ///
+    /// [subscriber]: https://docs.rs/tracing/latest/tracing/subscriber/trait.Subscriber.html
+    /// [filter]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.Targets.html
+    /// [`Layer`]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/trait.Layer.html
+    /// [`Server`]: crate::Server
+    ///
```
hawkw, line 253: nit: maybe worth adding something like:
```suggestion
    ///
    /// ## Panics
    ///
    /// * If the subscriber's background thread could not be spawned.
    ///
```

console-subscriber/src/builder.rs
```diff
@@ -149,6 +158,283 @@ impl Builder {
...
+    /// In addition to the [`TasksLayer`], which collects instrumentation data
+    /// consumed by the console, the default [`Subscriber`][sub] initialized by this
+    /// function also includes a [`tracing_subscriber::fmt`] layer, which logs
+    /// tracing spans and events to stdout. Which spans and events are logged will
+    /// be determined by the `RUST_LOG` environment variable.
+    ///
+    /// **Note**: this function sets the [default `tracing` subscriber][default]
+    /// for your application. If you need to add additional layers to a subscriber,
+    /// see [`spawn`].
+    ///
```
hawkw, line 177: nit: maybe worth adding:
```suggestion
    ///
    /// ## Panics
    ///
    /// * If the subscriber's background thread could not be spawned.
    /// * If the [default `tracing` subscriber][default] has already been set.
    ///
```

console-subscriber/src/builder.rs
```diff
@@ -149,6 +158,283 @@ impl Builder {
...
+/// In addition to the [`TasksLayer`], which collects instrumentation data
+/// consumed by the console, the default [`Subscriber`][sub] initialized by this
+/// function also includes a [`tracing_subscriber::fmt`] layer, which logs
+/// tracing spans and events to stdout. Which spans and events are logged will
+/// be determined by the `RUST_LOG` environment variable.
+///
+/// **Note**: this function sets the [default `tracing` subscriber][default]
+/// for your application. If you need to add additional layers to a subscriber,
+/// see [`spawn`].
+///
```
hawkw, line 354: nit: maybe worth adding:
```suggestion
///
/// ## Panics
///
/// * If the subscriber's background thread could not be spawned.
/// * If the [default `tracing` subscriber][default] has already been set.
///
```

console-subscriber/src/builder.rs
```diff
@@ -149,6 +158,283 @@ impl Builder {
...
+/// This function spawns the console subscriber's [`Server`] in its own Tokio
+/// runtime in a background thread.
+///
+/// Unlike [`init`], this function does not set the default subscriber, allowing
+/// additional [`Layer`]s to be added.
+///
+/// [filter]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.Targets.html
+/// [`Layer`]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/trait.Layer.html
+/// [`Server`]: crate::Server
+///
```
hawkw, line 427: nit: maybe worth adding:
```suggestion
///
/// ## Panics
///
/// * If the subscriber's background thread could not be spawned.
///
```

hawkw (approved):
okay, this looks great to me --- i commented on some minor docs suggestions that i can go ahead and apply, and then we can merge this!

a couple of thoughts for follow-ups:
- do we _really_ need an explicit `with_default_env`? would it be better to just have `Builder::default()` read the default env variables if they're set? i could be convinced either way on that one.
- we should probably rename `TasksLayer` to `ConsoleLayer` or something, since it now also collects data about resources and async ops :)

but those should be addressed separately.

console-subscriber/src/builder.rs
hawkw: i think `impl Layer` was the right choice. it also now means that we can (theoretically) use this with subscribers other than the default `tracing_subscriber::Registry` type!

hawkw:
Thanks again to @asonix (and @seanmonstar) for working on this! <3

console-subscriber/src/builder.rs
```diff
@@ -149,6 +158,314 @@ impl Builder {
...
+/// Unlike [`init`], this function does not set the default subscriber, allowing
+/// additional [`Layer`]s to be added.
+///
+/// This function is equivalent to the following:
+/// ```
+/// use console_subscriber::TasksLayer;
+///
+/// let layer = TasksLayer::builder().with_default_env().spawn();
+/// # drop(layer); // to suppress must_use warnings
```
hawkw: ugh, the docs i added here broke the build, lol...dropping the layer is insufficient because its type is not inferred:
```suggestion
/// # use tracing_subscriber::prelude::*;
/// # tracing_subscriber::registry().with(layer).init(); // to suppress must_use warnings
```

PR #220, opened 2021-12-17
chore(nix): refactor for flake support and add package build

jrobsonchase:
Pins `nixpkgs` to a known working version for the flake.
Things should be unchanged for non-flake-users.


shell.nix
```diff
@@ -1,19 +1,43 @@
...
+      cacert
+      gcc
+      cmake
+      pkg-config
+      openssl
+    ]
+    ++ lib.optional haveGlibcLocales glibcLocales-utf8
+    ++ lib.optionals stdenv.isDarwin [ libiconv ]
+    ++ tokio-console.buildInputs
+    ++ tokio-console.nativeBuildInputs;
```
jrobsonchase: There's probably a better way to accomplish this.

shell.nix
```diff
@@ -1,19 +1,43 @@
...
+      bashInteractive
+      docker
+      cacert
+      gcc
+      cmake
+      pkg-config
+      openssl
+    ]
+    ++ lib.optional haveGlibcLocales glibcLocales-utf8
+    ++ lib.optionals stdenv.isDarwin [ libiconv ]
```
jrobsonchase: I removed a `Security` package from this list of optional deps. `nix flake check` couldn't find it, and I couldn't find a reference to it anywhere in `nixpkgs`. If it's from an overlay somewhere, I can add that as a flake input and bring the package back.

shell.nix
```diff
@@ -1,19 +1,43 @@
...
+      direnv
+      binutils
+      stdenv
+      bashInteractive
+      docker
+      cacert
+      gcc
+      cmake
+      pkg-config
+      openssl
```
jrobsonchase: Not actually sure if any of this is needed. 

* direnv should already exist, since it seems a little silly to start a nix shell manually, load the direnv hook from the shell environment, and then let direnv take over from there.
* Most of the build tools seem to either be unnecessary or already brought in by the tokio-console inputs.

Docker seems like the only useful thing that this adds.

shell.nix
```diff
@@ -1,19 +1,43 @@
 scope@{ pkgs ? import <nixpkgs> { } }:
+with pkgs;
+let
+  tokio-console = import ./default.nix { inherit pkgs; };
 
-let env = (import ./default.nix scope);
+  haveGlibcLocales = pkgs.glibcLocales != null && stdenv.hostPlatform.libc == "glibc";
+  glibcLocales-utf8 = (glibcLocales.override { locales = [ "en_US.UTF-8" ]; });
```
jrobsonchase: Same "not sure if actually needed" comment regarding all of the glibc locale stuff that's going on here.

shell.nix
```diff
@@ -1,19 +1,43 @@
...
   LC_ALL = "en_US.UTF-8";
   OPENSSL_DIR = "${openssl.dev}";
   OPENSSL_LIB_DIR = "${openssl.out}/lib";
   SSL_CERT_FILE = "${cacert}/etc/ssl/certs/ca-bundle.crt";
   GIT_SSL_CAINFO = "${cacert}/etc/ssl/certs/ca-bundle.crt";
   CURL_CA_BUNDLE = "${cacert}/etc/ca-bundle.crt";
   CARGO_TERM_COLOR = "always";
-  RUST_BACKTRACE = "1";
-  buildInputs = [ (import ./default.nix { inherit pkgs; }) ];
+  RUST_BACKTRACE = "full";
```
jrobsonchase: Aaaand one final "not sure if needed." The package builds fine without all of these variables set. Maybe just the `RUST_BACKTRACE` and `CARGO_TERM_COLOR`?

shell.nix
hawkw: we probably don't need docker, fwiw, it was only there because i copied and pasted from a different nix derivation, and forgot to remove it. we should probably remove the docker dep!

shell.nix
hawkw: i'm fine with removing anything that isn't necessary to build the package --- again, i copied and pasted the config from a different nix derivation originally, and forgot to remove some stuff!

i think the `PROTOC` and `PROTOC_INCLUDE` env vars are probably necessary, but I think we can safely remove all the openssl stuff...

tokio-console.nix
```diff
@@ -0,0 +1,56 @@
...
+
+  inherit src;
+
+  cargoLock = {
+    lockFile = "${src}/Cargo.lock";
+  };
+
+  meta = {
+    inherit (cargoTOML.package) description homepage license;
+    maintainers = cargoTOML.package.authors;
```
hawkw: oh TIL you can do this, that's awesome!

default.nix
```diff
@@ -1,37 +1,2 @@
...
-    OPENSSL_DIR = "${openssl.dev}";
-    OPENSSL_LIB_DIR = "${openssl.out}/lib";
-    SSL_CERT_FILE = "${cacert}/etc/ssl/certs/ca-bundle.crt";
-    GIT_SSL_CAINFO = "${cacert}/etc/ssl/certs/ca-bundle.crt";
-    CURL_CA_BUNDLE = "${cacert}/etc/ca-bundle.crt";
-    CARGO_TERM_COLOR = "always";
-    RUST_BACKTRACE = "full";
-  };
-}
+{ pkgs ? import <nixpkgs> { } }:
```
hawkw: hmm, there are now enough nix configs in the repository that i wonder if we should stick them in `console/nix/` instead of having them all in `console/`? not a hard blocker though...

hawkw (approved):
thanks for working on this, this is great!

i had initially added the Nix derivations just so that i could have a dev environment with direnv, but the plan was always to _eventually_ actually have a nix package for building the console and hopefully submit it to upstream nixpkgs. thank you for working on that!

i'm not totally sure what nixpkgs' preferred procedure for packages is in cases where the package's repository contains a nix derivation --- can we add something to nixpkgs that depends on the derivation in this repository? IMO that would be preferable, since we can use the same nix configuration for the dev environment, and not have to keep two separate configs in sync...but, we can figure this out later when we actually submit to nixpkgs. :)

besides that, i commented on a couple minor things, but overall this looks good to me!

default.nix
jrobsonchase: Kinda depends on how people are going to be using `nix`. I think the `flake.{nix,lock}` are the only things that actually *have* to live at the repo root. `shell.nix` and `default.nix` might be happier at the root if people are going to be invoking `nix-build` and `nix-shell` manually, since the `nix-*` commands will look in their current directory for their respective files first. Otherwise, you'd have to give them a path to their file.

If `.envrc` is going to be the main nix entrypoint for most, then everything but the flake can definitely be relocated.

shell.nix
jrobsonchase: Cool, I'll clean this up then :+1:

default.nix
hawkw: :woman_shrugging: we can figure that out later, i think leaving them all in the root is fine for now.

jrobsonchase: Eh, I went ahead and moved them. I figure since the only thing that was here before was the `.envrc` and `shell.nix`, it's pretty unlikely that anyone's existing workflow will be significantly disrupted. I don't know anyone who manually uses `nix-shell` anymore, and there was nothing to `nix-build` to begin with :grin:

jrobsonchase:
> i'm not totally sure what nixpkgs' preferred procedure for packages is in cases where the package's repository contains a nix derivation --- can we add something to nixpkgs that depends on the derivation in this repository? IMO that would be preferable, since we can use the same nix configuration for the dev environment, and not have to keep two separate configs in sync...but, we can figure this out later when we actually submit to nixpkgs. :)

I think there are a few options here. Most of the rust crates I've seen in `nixpkgs` are built from crates.io, but you can also use a git repository as the source. I agree that it would be preferable to use the `tokio-console.nix` derivation directly. It's technically possible to do something like

`nixpkgs/path/to/tokio-console.nix`:
```nix
builtins.fetchGit {
    url = "https://github.com/tokio-rs/console";
    name = "tokio-console";
    ref = "refs/tags/tokio-console-v0.1.0";
    rev = "2c80d8a0378fba651f89c7a6029108b6d8dd7aba";
    submodules = false;
} + "/nix/tokio-console.nix"
```

Somewhere else in `nixpkgs`:
```nix
tokio-console = pkgs.callPackage ./tokio-console.nix { };
```

But I'm not sure if that would pass review, since the things for review are entirely outside of the main `nixpkgs` repo.

hawkw (approved):
this looks great to me, thanks again (assuming you tested that it still builds with the various unnecessary build inputs removed :))

jrobsonchase:
Yep, all looks good from my end :smiley:

hawkw:
Hmm, this commit seems to have broken `rust-analyzer` for me. The error message suggests that the sysroot no longer contains the `rust-src` component:
```
rust-analyzer failed to load workspace: Failed to find sysroot for Cargo.toml file /home/eliza/Code/console/Cargo.toml. Is rust-src installed?: can't load standard library from sysroot
/nix/store/449lmdsx8dv2q5fn4xbkxk51d4cg2zih-rustc-1.56.1
(discovered via `rustc --print sysroot`)
try installing the Rust source the same way you installed rustc
```

I think this is due to the change from including `rustup` in `paths` to using `rustPlatform.buildRustPackage` in `tokio-console.nix`?

PR #222, opened 2021-12-17
fix(console): don't enable crossterm mouse capture

### LINKED ISSUE #167: Running the console in alacritty blocks selections and copy-paste
My guess is that the terminal is led to believe that a mouse-supporting application is running. I'm not aware of any mouse features in the tokio console, though.

hawkw:
Currently, the console captures mouse input, but it doesn't actually do
anything with mouse input yet. This is a shame, because it means you
can't do things like copying and pasting text out of the console.

This commit disables mouse capture.

In the future, if we add mouse support to the UI, we can re-enable mouse
capture...but it would be nice to make sure copying and pasting still
works (we may have to reimplement it?).

Fixes #167

### DIFF
--- a/console/src/term.rs
+++ b/console/src/term.rs
@@ -4,23 +4,20 @@ pub use tui::{backend::CrosstermBackend, Terminal};
 
 pub fn init_crossterm() -> color_eyre::Result<(Terminal<CrosstermBackend<io::Stdout>>, OnShutdown)>
 {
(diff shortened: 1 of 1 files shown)


pnkfelix:
I was *wondering* why it was such a struggle to select text to cut and paste!!!

pnkfelix (approved):

Issue #131, opened 2021-09-10
console: filter tasks list


hawkw:
It would be nice to be able to filter which tasks are displayed in the tasks list.

Some ways we might want to filter the tasks list:
- By name (with a regex?)
- By target (with a regex?)
- By fields:
    - by the presence of a field with a given name
    - by the value of a field with a given name (regex?)

Since we would want to allow filtering on different columns in the task list, we would need some UI for selecting which column is being filtered on, as well as for writing the actual filter. Here's a simple proposal:
- When the user presses a particular key, a text entry box is opened to type in a filter pattern
- Using the arrow keys in the filter entry mode selects which column is _filtered_ on, rather than which column sorts the task list
- pressing some other key (enter?) adds the filter, and returns the user to the mode where the arrow keys select the column to sort by

Questions:
- Do we want to support multiple filters? We probably do, but that complicates the UI
    * Opening the filter entry box again would then create a new filter, rather than editing the current filter
    * There would need to be some way to edit an existing filter as well
    * Perhaps we should add a UI box that lists all the current filters?
    * Some kind of command for clearing all filters
 - How to combine patterns in filters? e.g. users might want to have `OR` filters or `AND` filters...

Most of these questions are things we can punt on for an MVP implementation, especially if we only allow a single filter at a time. But, we'll want to figure them out eventually.

UI inspiration:
- `htop` (which only allows filtering on _one_ column, the process's name):
   ![image](https://user-images.githubusercontent.com/2796466/132888547-222e21e0-ac7e-4e9d-b49d-4b07e8b07e44.png)
- `atuin` history search:
   ![image](https://user-images.githubusercontent.com/2796466/132888648-f29387f5-c651-4693-b905-84217a38be17.png)

hawkw:
I marked this as `E-hard` because a _complete, full-featured_ filtering implementation is going to be a big pile of work (see the "questions" section above). However, an initial MVP where there's only one active filter at a time is probably relatively easy...

seanmonstar:
There's also other things we'd probably want to be able to "enter" dynamically, like rewinding to a certain spot, or changing a config value, etc.

hawkw:
> There's also other things we'd probably want to be able to "enter" dynamically, like rewinding to a certain spot, or changing a config value, etc.

Yeah, it would make sense to have stuff like text boxes and other UI components eventually be general-purpose...

zaharidichev:
[...]

PR #223, opened 2021-12-17
feat(console): Add vi style keybinds for navigating tables

k0nserv:
Adds `hjkl` as alternatives to the arrow keys and `G` and `gg` to navigate
to the first and last rows in tables, respectively.

Changes to the UI:

image

### DIFF
--- a/console/src/view/table.rs
+++ b/console/src/view/table.rs
@@ -40,6 +40,8 @@ pub(crate) struct TableListState<T: TableList> {
     pub(crate) selected_column: usize,
     pub(crate) sort_descending: bool,
     pub(crate) table_state: TableState,
+
+    last_key_event: Option<input::KeyEvent>,
 }
 
 pub(crate) struct Controls {
@@ -66,18 +68,19 @@ impl<T: TableList> TableListState<T> {
         }
     }
 
-    pub(in crate::view) fn key_input(&mut self, input::KeyEvent { code, .. }: input::KeyEvent) {
+    pub(in crate::view) fn key_input(&mut self, event: input::KeyEvent) {
         use input::KeyCode::*;
         let header_len = T::HEADER.len();
+        let code = event.code;
         match code {
-            Left => {
+            Left | Char('h') => {
                 if self.selected_column == 0 {
                     self.selected_column = header_len - 1;
                 } else {
                     self.selected_column -= 1;
                 }
             }
-            Right => {
+            Right | Char('l') => {
                 if self.selected_column == header_len - 1 {
                     self.selected_column = 0;
                 } else {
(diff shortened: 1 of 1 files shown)


hawkw (commented):
nice, i'm sure this will make a lot of people very happy! :)

we also display the keybindings in the UI, so maybe we should also add the vi-style keybindings to the table controls we display, here:
https://github.com/tokio-rs/console/blob/1aa9b594f30e42098c6c6bbf41eb1d2b01dc0426/console/src/view/table.rs#L162-L176

also, I wonder if it's worth making a CLI option for toggling between vi only, arrow keys only, and enabling both? but, we could also do that in a follow-up PR.

k0nserv:
Thanks for the suggestions, I considered both those options as I wrote this. 

For adding the bindings to the UI I wasn't sure how to do it so I wanted to raise it here instead.

One possibility is `\u{2190}(h)\u{2192}(l)` for left/right and likewise for up and down. I'm happy to add this, my concern was that it might be a bit messy. I guess for `G` and `gg` those could just be added directly.

e.g. like this
image


An option to consider if this display of keybinding does get too unwieldy is `?` to bring up a floating window that displays them all.

About the CLI toggling: My thinking was there's no harm in having both by default, folks will use the ones they know/prefer? I'm not sure there's any benefit to allowing them to be disabled, but I can certainly add that in a follow up PR if you wish

hawkw:
> One possibility is `\u{2190}(h)\u{2192}(l)` for left/right and likewise for up and down. 

How about `\u{2190}/\u{2192} or h/l` and similar, instead? (ideally, with the `or` and the slashes not bolded). But, honestly, I'm fine with whatever. I think just displaying `gg` and `G` separately the way you did in your screenshot looks correct.

> An option to consider if this display of keybinding does get too unwieldy is `?` to bring up a floating window that displays them all.

Yeah, that might be a good idea, especially if we add more controls for things like filtering (#131) in the future. A simpler solution might just be to make it two lines rather than one line, but that eats up more vertical real estate...

> About the CLI toggling: My thinking was there's no harm in having both by default, folks will use the ones they know/prefer? I'm not sure there's any benefit to allowing them to be disabled, but I can certainly add that in a follow up PR if you wish

Honestly, the only real benefit I had in mind is that it makes the controls list shorter to disable one set of keybindings :) I really don't think people are likely to disable or enable keybindings with a command-line option. I think that if we eventually add a _config file_, it might be more likely to want to configure keybindings there, but in that case, I'd imagine users would want to specify custom keybindings, rather than just turning them on or off. All of that is just something to think about for the future, though.

Once we've added the vi keybindings to the table controls, I'll be happy to merge this!

k0nserv:
image


Here's what it looks like now

hawkw:
> image
> 
> Here's what it looks like now

Yup, that looks good to me.  I noticed that the controls line is now quite long, and probably won't fit in a narrower terminal...and that we truncate the line instead of wrapping it. So, I went ahead and implemented wrapping in PR #231, so that we can add more text to the controls and still have it display nicely on narrower terminals. Once that's merged, let's go ahead and merge this PR!

hawkw:
@k0nserv any updates here? It would be nice to get this updated with the latest `main` branch, and then I'll be happy to merge it!

k0nserv:
I saw that you had merged main into this branch and I thought you'd merge when #231 was merged. 

What do you need me to do before we can merge?

hawkw:
> What do you need me to do before we can merge?

It looks like the change for adding the keybindings to the help text hasn't actually been added to this branch yet? Once that's done, I'll merge this.

k0nserv:
Huh, I thought I pushed that, but I guess not then. Shall resolve

k0nserv:
Rebased on `main` and pushed now

k0nserv:
I believe you need to approve the some of the workflows for them to run

image

hawkw (approved):
okay, looks good to me! thanks again for working on this!

k0nserv:
Thanks for merging :+1:

Issue #180, opened 2021-11-04
thread 'console_subscriber' panicked at 'attempt to subtract with overflow'


pnkfelix:
thread 'console_subscriber' panicked at 'attempt to subtract with overflow', console/console-subscriber/src/aggregator/mod.rs:242:9

This is while attaching the console to some web-crawler demo code I've been working on. (It's deliberately naive code.)

Here is that code base:

https://github.com/pnkfelix/huaa-exsets/tree/console-play/crawler

The nature of that current web crawler is that it will visit pages in undetermined order, which means that the overflow panic is not always the first panic it encounters. So, it currently does not deterministically reproduce the bug here. But I bet I can narrow it down.

pnkfelix:
Oh actually I think I just encountered this on a simpler test case. I'll try to produce that shortly.

pnkfelix:
Simpler test case (and should be deterministic now, though it does take a while for the problem to surface; about 90 seconds or so on my desktop):

```rust
#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
    console_subscriber::init();

    let mut handles = Vec::new();
    for i in 1..1000 {
        handles.push(tokio::task::spawn(async move {
            loop { fact_yielding(i).await; }
        }));
    }

    for h in handles {
        h.await?;
    }

    Ok(())
}

async fn fact_yielding(n: u32) -> f64 {
    let mut i = 0_u32;
    let mut accum = 1_f64;
    loop {
        if i == n { break; }
        tokio::task::yield_now().await;
        i += 1;
        accum *= i as f64;
    }
    return accum;
}
```

I am running this atop Eliza's branch with the fix for compute-intensive task(s), i.e. branch https://github.com/tokio-rs/console/tree/eliza/approaching-capacity

hawkw:
Hmm, the panic occurs on this line: https://github.com/tokio-rs/console/blob/c1111bc72a616f17c67e546a035d513214dd52e4/console-subscriber/src/aggregator/mod.rs#L242

That code seems to assume the number of polls will never be zero when handling a poll-end event, because it's incremented when starting a poll, and a poll can't end without having been started. This makes me wonder if the panic occurs because the event that started the poll was dropped because the event channel was at capacity, but the event for ending the poll occurred when there was capacity for it.

[...]

PR #234, opened 2021-12-20
fix(subscriber): use saturating arithmetic for attribute updates

hawkw:
Currently, an issue exists where a subtraction attribute update for a
resource or async op attribute may overflow, if a previous add update
was dropped due to event buffer capacity. This may result in the
aggregator task panicking. See
https://github.com/tokio-rs/console/issues/180#issuecomment-998132325
for details.

This branch changes all resource updates to use saturating arithmetic.
In practice, I think it's _much_ more likely for a subtract to overflow
due to missed adds than it is for an add to overflow due to missed subs,
but I made the addition saturating as well, just in case.

We should probably also increase the default event buffer capacity,
since it seems like lots of people are hitting problems with this...


zaharidichev (approved):
Good catch!

Issue #225, opened 2021-12-17
console: emoji/symbols glossary


hawkw:
When the terminal supports Unicode, the console CLI will use a number of emojis and other symbols as icons etc. This is good, because it's easier to visually scan the UI this way, and in many cases, it uses fewer characters, allowing more screen real estate for things like paths that are potentially long. However, in some cases, the precise meaning of a particular emoji may not be _quite_ as obvious as text. We might want to add some kind of glossary explaining the meanings of some of the emoji/symbols we use.

This could be in the documentation, or somewhere in the UI itself (a floating help window you toggle by pressing a key?)

bIgBV:
A floating help window in general would be a good idea. It could be part of building the glossary I think.

bIgBV:
Looking at the code and thinking a spending some more time thinking about the implementation: 

Different views would have different help information to display (including the glossary of emojis if they are being used), which pertains specifically to the view. If every view had a way to display help information (probably in the bottom row in one or two lines) then we could hook the event handler to display the right help information based on the current view state.

bIgBV:
I looked into whether or not it was possible to do floating windows using corssterm, but it doesn't look like there's an easy way to do that.

javihernant:
I think this can be closed as PR [#432 ](https://github.com/tokio-rs/console/pull/432) was already merged

jonatan-zeidler-bl:
The modal help view prints the controls, but does not explain emojis/symbols as of #432. I don't think, this should be closed.

PR #243, opened 2021-12-25
Introduce new help view to display information on any view

bIgBV:
As a part of addressing #225 , I'm introducing a new floating pane which can show information relevant to the current in focus pane. This can include shortcuts, key bindings and glossary information when using symbols and emojis..


console/src/view/mod.rs
```diff
@@ -164,6 +171,17 @@ impl View {
                     }
                 }
             }
+            _ => {
```
bIgBV: Just realized that this is unreachable. I guess I will have to create a new `ViewState` state to indicate that the help state is present. I think as long as I store what the previous state was, I can go back to it when `esc` (or any other key) is pressed.

bIgBV:
Does anyone have thoughts on what the contents of the help popup should be in different views? For now I'm working on the mechanism to pump the information from the view to the popup. This is what it looks like currently:

image

hawkw:
@bIgBV this is awesome, thanks for working on this!

As for what should be displayed in the help text:

 - I think we should probably display the controls for the current view, even though they're also displayed in the header...perhaps once the help text feature is merged, we should consider hiding the controls from the header by default, and just display what key is needed to open the help text?
- It could be useful to explain the meanings of different emojis/symbols in the current view?
- Eventually, the same popup system could maybe be used for displaying detailed descriptions of warnings, but that's something we would add later (since we don't have a system for associating help with warnings yet...)

bIgBV:
* I think including the controls and what the symbols mean in a given page in the popup would be a good start.
* And I love the idea of displaying the meaning behind/expanding on warnings in the future. And it should be fairly straightforward to extend this to perform that function.

console/src/view/help.rs
```diff
@@ -53,15 +59,28 @@ where
...
+                height += 1
+            }
+        }
+
+        let content_layout = layout::Layout::default()
+            .direction(layout::Direction::Vertical)
+            .margin(0);
+
+        let content_area = content_layout
+            .constraints([layout::Constraint::Length(height)])
```
bIgBV: I wanted to create an area within the popup to render the text into, which would be responsive, but doing it like this doesn't seem to be working.

hawkw: what isn't working, exactly? is the problem that the text isn't wrapping?

console/src/view/help.rs
```diff
@@ -0,0 +1,86 @@
...
+            .split(popup_area)[0];
+
+        let display_text = Paragraph::new(content).block(styles.border_block().title("Help"));
```
hawkw: if you want the text to wrap, you need to call `.wrap()` on the `Paragraph`:
```suggestion
        let display_text = Paragraph::new(content)
            .block(styles.border_block().title("Help"))
            .wrap(Wrap { trim: true });
```

bIgBV: This was exactly what I was looking for. Thanks!

bIgBV:
@hawkw got the text in the help popup to wrap properly (thanks again for the suggestion). 

image

I also updated the text in the detail views to be (slightly) less generic. I think this is in a good place to be merged and I could separately open a new PR once we finalize the content of the help text for different views.

hawkw:
@bIgBV sorry it took me so long to look at this, I'd really like to actually give it a review now. It looks like this branch is now out of date with `main`, can I get you to rebase or merge `main`?

bIgBV:
@hawkw no worries. I just rebased this branch on top of main.

tokio-console/src/view/help.rs
```diff
@@ -0,0 +1,70 @@
...
+        frame: &mut tui::terminal::Frame<B>,
+        _area: layout::Rect,
+        _state: &mut State,
+    ) {
+        let r = frame.size();
+        let content = self
+            .help_text
+            .take()
+            .expect("help_text should be initialized")
+            .into();
```
hawkw, line 36: it's kind of weird to me that `help_text` is an `Option` that we have to `take` here? Is there a reason the `render` method has to take an `&mut self`? It could just take `self` --- this isn't a trait method or anything.

tokio-console/src/view/help.rs
```diff
@@ -0,0 +1,70 @@
...
+            help_text: Some(help_text),
+        }
+    }
+
+    pub(crate) fn render<B: tui::backend::Backend>(
+        &mut self,
+        styles: &view::Styles,
+        frame: &mut tui::terminal::Frame<B>,
+        _area: layout::Rect,
+        _state: &mut State,
```
hawkw, line 29: is there a reason this has these arguments, when it doesn't use them? since this isn't a trait method, I think we should just not take these arguments just because they are passed to other views' `render` methods.
```suggestion
```

tokio-console/src/view/mod.rs
```diff
@@ -168,32 +177,56 @@ impl View {
         update_kind
     }
 
+    fn handl_help_popup(&mut self, event: crossterm::event::Event) {
```
hawkw, line 180: typo: should probably be
```suggestion
    fn handle_help_popup(&mut self, event: crossterm::event::Event) {
```

tokio-console/src/view/mod.rs
```diff
@@ -168,32 +177,56 @@ impl View {
...
+        if input::is_help_toggle(&event) {
+            // TODO: Pause state if we are about to show the help toggle
+            self.show_help_toggle = !self.show_help_toggle;
+        }
+        if self.show_help_toggle && !input::is_help_toggle(&event) {
```
hawkw, line 185: nit, take it or leave it: if we used `if ... else if`, we could avoid the second call to `input::is_help_toggle`:
```suggestion
        } else if self.show_help_toggle {
```

tokio-console/src/view/mod.rs
```diff
@@ -168,32 +177,56 @@ impl View {
         update_kind
     }
 
+    fn handl_help_popup(&mut self, event: crossterm::event::Event) {
+        if input::is_help_toggle(&event) {
+            // TODO: Pause state if we are about to show the help toggle
```
hawkw, line 182: we should probably do this before merging?

tokio-console/src/view/mod.rs
```diff
@@ -168,32 +177,56 @@ impl View {
...
             ViewState::TaskInstance(ref mut view) => {
                 let now = state
                     .last_updated_at()
                     .expect("task view implies we've received an update");
                 view.render(&self.styles, frame, area, now);
+                help_content = HelpView::new(TaskView::render_help_content(&self.styles));
             }
             ViewState::ResourceInstance(ref mut view) => {
                 view.render(&self.styles, frame, area, state);
+                help_content = HelpView::new(ResourceView::render_help_content(&self.styles));
```
hawkw, line 221: I notice that this will construct the help content for each view *every* time we render a screen, even if we're not showing the help window. That seems a little bit inefficient, since we have to allocate a big `vec` and a bunch of strings.

I think it might be more efficient to add a second `match self.state` inside of the `if self.show_help_toggle` box, and only build the help text if we are actually going to display a help box. That might make the method a bit longer, but I think it's worth doing to avoid having to format the help text when we're not going to display it...

tokio-console/src/view/task.rs
```diff
@@ -267,6 +267,10 @@ impl TaskView {
         frame.render_widget(fields_widget, fields_area);
         frame.render_widget(percentiles_widget, percentiles_area);
     }
+
+    pub(in crate::view) fn render_help_content(_styles: &view::Styles) -> Spans<'static> {
+        Spans::from(vec![Span::raw("A view to diplay help data for a task")])
```
hawkw, line 272: should this also include controls?

hawkw (changes requested):
Sorry it took me such a long time to review this, but it looks like a good start. I had some small suggestions.

Thanks for working on this! If you no longer have the time to continue working on it, I could make the remaining suggested changes myself and go ahead and merge this branch. Let me know if you'd like that. Thanks again!

hawkw:
@bIgBV quick ping — still interested in working on this?

bIgBV:
@hawkw I completely missed my notifications on this. While I'm still interested in working on this, my job has kept me pretty busy, and will keep doing so for the next couple of weeks. So if you can pull in my changes and add the ones you suggested on top of that, please go for it!

erwanor:
@bIgBV Hey, would you feel comfortable giving me write access to your fork repo ("Adding contributors" in settings). Otherwise, I'd need to start a fork from scratch and apply your changes etc. which might get a bit gnarly, and we'd lose this convo history.

bIgBV:
@erwanor I've sent you an invitation.

hds:
As no further progress had been made on this PR, a new one was opened (#432) inspired by it and has now been merged.

PR #251, opened 2022-01-06
fix(console): don't make details requests with rewritten IDs

hawkw:
PR #244 moved the rewriting of `tracing` span IDs to sequential
low-number IDs from the `console-subscriber` crate to the console CLI.
However, this introduced a bug with task details, because that PR didn't
change the part of the console that makes `watch_details` RPCs to use
the `tracing` span ID recieved from the remote --- it continued to use
the `Task` and `Resource` structs' `id` fields. These are now different
from the `tracing` ID recieved from the remote, because they're
rewritten in the console CLI. This means that `watch_details` RPCs would
generally recieve an error, or (in the off-chance that a rewritten ID
collides with a low-numbered `tracing` ID) the details for anoter task
or resource.

This branch fixes the bug by changing the `Task` and `Resource` structs
to store both the span ID received from the remote _and_ the rewritten
pretty ID. This way, when we make `watch_details` RPC calls, we do it
with the span ID we received from the remote process.

I also changed some naming to make the distinction between rewritten IDs
and span IDs clearer.

### DIFF
--- a/console/src/state/resources.rs
+++ b/console/src/state/resources.rs
@@ -36,7 +36,15 @@ pub(crate) enum SortBy {
 
 #[derive(Debug)]
 pub(crate) struct Resource {
-    id: u64,
+    /// The resource's pretty (console-generated, sequential) ID.
+    ///
+    /// This is NOT the `tracing::span::Id` for the resource's `tracing` span on the
+    /// remote.
+    num: u64,
+    /// The `tracing::span::Id` on the remote process for this resource's span.
+    ///
+    /// This is used when requesting a resource details stream.
+    span_id: u64,
     id_str: InternedStr,
     parent: InternedStr,
     parent_id: InternedStr,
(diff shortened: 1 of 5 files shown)


console/src/state/resources.rs
```diff
@@ -36,7 +36,15 @@ pub(crate) enum SortBy {
...
-    id: u64,
+    /// The resource's pretty (console-generated, sequential) ID.
+    ///
+    /// This is NOT the `tracing::span::Id` for the resource's `tracing` span on the
+    /// remote.
+    num: u64,
+    /// The `tracing::span::Id` on the remote process for this resource's span.
+    ///
+    /// This is used when requesting a resource details stream.
+    span_id: u64,
```
zaharidichev, line 47: maybe we can rename it to just pretty_id or display_id?

zaharidichev (approved):
LGTM

zaharidichev (approved):
LGTM

console/src/state/resources.rs
hawkw: i'm also fine with a name like that; i went with `num` because they're sequential numbers, but I'd be happy to change it.

hawkw:
A follow-up that I didn't address in this branch, but that we probably ought to add, is something that's displayed in the UI when we can't get details for a task. Currently, when a `watch_details` RPC returns an error, we log it...but logs from the console aren't visible in the UI and can only be viewed if the user redirects them to a file. So, we probably want to add something in the UI showing that there was an error.

Issue #246, opened 2021-12-29
consider redesigning resource states to use `tracing` more idiomatically


hawkw:
Currently, state attributes for resources and async ops are represented by `tracing` events with a particular structure, within the resource's span. This is somewhat complicated, both in the instrumentation and in `console-subscriber`. These events consist of a field with the attribute's name and a value, and two additional optional fields for numeric values: a field `<attribute_name>.op`, which can be "add", "sub", or "override", and a field `<attribute_name>.unit`, which is used to optionally add a unit such as `ms`. The `op` field is used to record numeric values as a delta rather than as an absolute value.

Meanwhile, `tracing` already has a mechanism for attaching field values to spans, and recording new values for those fields, via the [`Span::record`](https://docs.rs/tracing/latest/tracing/span/struct.Span.html#method.record) method. I think we should change state attributes for resources/async ops to use this instead.

## Motivation

Here are some of the issues and questions I've noticed about the current approach.

 - It's fairly complicated
 - When spans for resources and async ops are created, there's no way to know all of the state attributes that will exist for that particular resource. 
 
   This means that as individual `state_update` events occur, the attributes are added to a hashmap one-by-one as the console layer becomes aware of those attributes. This isn't very efficient, as we might reallocate the hashmap multiple times. Also, if these maps are moved out of the aggregator task and updated in the `ConsoleLayer`, as in #238, they would have to be mutated to insert new attribute names, which would require adding a lock, even if the values themselves can all be updated atomically (since they're all integers or booleans). 

   If all these fields were declared up front on the span, however, we would know about all the fields that exist when the resource/async op is created, even if their values aren't populated yet. That way, we could create a single hashmap (or array) with all the attributes already present, and insert it into the span's extensions. This would only need a single allocation. If the values of those attributes are updated atomically, we wouldn't need any kind of locking to update or read from them.
 - I don't understand the advantage of recording changes in numeric values as deltas rather than as absolute values (the `.op` field and "add"/"sub"). All the values will eventually be reported to the `console` as absolute values in the wire format anyway. If we want to use these to display the "direction" in which a value has changed since the last update, we would have to implement that logic in the UI anyway, because the delta value is lost when sending the message on the wire.

   What do we gain from recording these values as deltas? As far as I can tell, it just makes the implementation much more complex.
 - I'm not sure if recording units as a structured field really buys us anything.
 
[...]

PR #255, opened 2022-01-07
subscriber: use span fields for state attr updates

zaharidichev:
This PR is an attempt to follow the suggestions in #246 to redesign the state update logic.
This change enables us to start switching to this new approach in tokio while still maintaining
the old way of emitting state updates. Now we can simply drive state updates by creating and
recording fields that look like `state.field_name.unit.delta`. 


zaharidichev:
closing in favor of #264

PR #257, opened 2022-01-09
console: fix task lookup in async ops view

zaharidichev:
#244 moved rewriting of ids to the console. We have however forgot to look up
the pretty ID when trying to figure out which task is running a particular async op.
The result of that was that we were displaying the raw ID for the task and we were
not able to properly resolve the name of the task if specified.

Before: 
Screenshot 2022-01-09 at 19 52 27

After: 
Screenshot 2022-01-09 at 19 53 24


hawkw (approved):
Thanks! This looks good to me!

PR #269, opened 2022-01-28
Fix histogram display

### LINKED ISSUE #262: Histogram empty with `tokio-console 0.1.1`
image

Also note that the ID seems to be different, while the rest of the stats is the same.
# Last working release

`tokio-console 0.1.0`

# Steps to reproduce

1. `cargo install tokio-console`
2. `tokio-console -V` -> `tokio-console 0.1.1`
3. `git clone https://github.com/tokio-rs/console`
4. `cd console\console-subscriber\examples`
5. `cargo run --example app`
6. `tokio-console`
image

# System Info
`rustc -V` -> `rustc 1.58.0 (02072b482 2022-01-11)`
Windows 10 21H2 build `19044.1466`

jtgeibel:
Task `Details` are tracked by the remote's `span::Id` while the
`Task::id()` is the friendly id.

The comparison in the view is changed to use the remote span ids and the
field in `Details` is renamed to match.

Fixes: #262

### DIFF
--- a/tokio-console/src/state/mod.rs
+++ b/tokio-console/src/state/mod.rs
@@ -215,7 +215,7 @@ impl State {
     pub(crate) fn update_task_details(&mut self, update: proto::tasks::TaskDetails) {
         if let Some(id) = update.task_id {
             let details = Details {
-                task_id: id.id,
+                span_id: id.id,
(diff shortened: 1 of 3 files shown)


hawkw (approved):
whoops! good catch, thanks for the fix!

PR #271, opened 2022-02-04
fix(api): fix accidental exhaustive matching on `metadata::Kind`

### LINKED ISSUE #270: Broken compile with tracing-core 0.1.22
A recent change to `tracing-core` in tokio-rs/tracing#1891 which is used by recently released `tracing` 0.1.29 causes broken compiles due to Kind being changed to a u8 rather than an enum. Manually pinning `tracing-core` to 0.1.21 fixes the issue.

```
error[E0004]: non-exhaustive patterns: `Kind(0_u8)` and `Kind(3_u8..=u8::MAX)` not covered
  --> /home/arlyon/.cargo/registry/src/github.com-1ecc6299db9ec823/console-api-0.1.1/src/common.rs:20:15
   |
20 |         match kind {
   |               ^^^^ patterns `Kind(0_u8)` and `Kind(3_u8..=u8::MAX)` not covered
   |
   = help: ensure that all possible cases are being handled, possibly by adding wildcards or more match arms
   = note: the matched value is of type `tracing_core::Kind`
```

Happy to open a PR; how should we handle cases where the `tracing_core::metadata::Kind` is not covered? Maybe we should also use a u8 to match?

hawkw:
The `tracing_core::metadata::Kind` type was not intended to be matched
exhaustively, so that its internal representation could be changed
without causing a breaking change. However, some fairly recent compiler
changes broke this, and made the type exhaustively matched. This
resulted in `tracing-core` 0.1.22 breaking `console-api`.

This PR fixes the breakage by changing `console-api` to avoid
exhaustively matching on `Kind`s.

Fixes #270

### DIFF
--- a/console-api/src/common.rs
+++ b/console-api/src/common.rs
@@ -17,9 +17,20 @@ impl From<tracing_core::Level> for metadata::Level {
 
 impl From<tracing_core::metadata::Kind> for metadata::Kind {
     fn from(kind: tracing_core::metadata::Kind) -> Self {
-        match kind {
-            tracing_core::metadata::Kind::SPAN => metadata::Kind::Span,
-            tracing_core::metadata::Kind::EVENT => metadata::Kind::Event,
+        // /!\ Note that this is intentionally *not* an exhaustive match. the
+        // `metadata::Kind` struct in `tracing_core` is not intended to be
+        // exhaustively matched, but compiler changes made past versions of it
+        // capable of being matched exhaustively, which was never intended.
+        //
+        // Therefore, we cannot write a match with both variants without a
+        // wildcard arm. However, on versions of `tracing_core` where the
+        // compiler believes the type to be exhaustive, a wildcard arm will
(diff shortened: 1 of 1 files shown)


hawkw:
Once this is published, I think we may need to yank all previous versions of `console-api`, so that `cargo-install` always uses the latest version? Or, we can bump `tokio-console`'s dependency and publish a new version of it as well.

guswynn:
This looks good to me, but is there any reason not to just delete the whole `From` impl? it appears to be unused? Is it just because it now technically a public api?

edit: answered in the other issue

hawkw:
> This looks good to me, but is there any reason not to just delete the whole `From` impl? it appears to be unused? Is it just because it now technically a public api?
> 
> edit: answered in the other issue

yeah, deleting the `From` impl would require releasing `console-api` 0.2, which i don't super want to do...

seanmonstar (approved):

PR #272, opened 2022-02-04
Reword comment on the tracing_core::metadata::Kind match

dtolnay:
I don't think it's accurate that a compiler change made this exhaustively matchable when it wasn't before. All versions of rustc since 1.0.0 have allowed exhaustively matching on that kind of data structure from outside of the defining crate.

### DIFF
--- a/console-api/src/common.rs
+++ b/console-api/src/common.rs
@@ -17,16 +17,16 @@ impl From<tracing_core::Level> for metadata::Level {
 
 impl From<tracing_core::metadata::Kind> for metadata::Kind {
     fn from(kind: tracing_core::metadata::Kind) -> Self {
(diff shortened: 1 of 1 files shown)


hawkw:
Thanks, I could have sworn this was not always exhaustively matched, but my memory is probably just bad.

hawkw (approved):

dtolnay:
The correct pattern it sounds like you meant to use is:

```rust
pub struct Kind(KindInner);

enum KindInner {
    Event,
    Span,
    // reserved for future use:
    SomethingElse,
}

impl Kind {
    pub const EVENT: Self = Kind(KindInner::Event);
    pub const SPAN: Self = Kind(KindInner::Span);
}
```

hawkw:
Ah, yeah, I think that was the problem.

PR #276, opened 2022-02-08
feat(subscriber) add filter env var builder parameter

### LINKED ISSUE #206: subscriber: consider adding filter env var builder parameter
>what do you think about making the filter environment variable a builder parameter? so, by default, it uses `RUST_LOG`, but there's a separate builder method to change the filter env var? something like:
>
>```rust
>let console = console_subscriber::Builder::new()
>    .with_filter_env_variable("MYAPP_LOG")
>    .build_console_layer();
>```
>
> _Originally posted by @hawkw in https://github.com/tokio-rs/console/pull/195#discussion_r769848327_

mujpao:
Add a new builder method to `console_subscriber::Builder` to configure the filter environment variable to use for `tracing` events.

closes #206


console-subscriber/src/builder.rs
```diff
@@ -145,6 +149,16 @@ impl Builder {
         }
     }
 
+    /// Sets the filter environment variable for tracing events.
+    ///
+    /// By default, this is `RUST_LOG`.
+    pub fn filter_env_variable(self, filter_env_variable: impl Into<String>) -> Self {
```
hawkw: I think maybe we should rename this to `filter_env_var` rather than `filter_env_variable`? It feels a little bit weird to me that we abbreviate "environment" to "env" but we don't abbreviate "variable" to "var", especially since the stdlib function is `std::env::var` rather than `std::env::variable`.

i know i was the one who suggested this name originally, sorry about that.

console-subscriber/src/builder.rs
```diff
@@ -145,6 +149,16 @@ impl Builder {
         }
     }
 
+    /// Sets the filter environment variable for tracing events.
+    ///
+    /// By default, this is `RUST_LOG`.
```
hawkw: some notes on the documentation here:
 - this specifically filters the tracing events that are _logged to stdout_ by the `fmt` layer added in `Builder::init`, it does not effect the events that are recorded by the console itself. I think the documentation should probably state that explicitly. 
- only `Builder::init` adds a logging layer; `Builder::build` and `Builder::spawn` will not. I think this documentation should state that clearly.
- there should probably be a link to the `tracing` documentation explaining the syntax for the environment variable.

here's my suggestion:
```suggestion
    /// Sets the environment variable used to configure which `tracing` events
    /// are logged to stdout.
    ///
    /// The [`Builder::init`] method configures the default `tracing`
    /// subscriber. In addition to a [`ConsoleLayer`], the subscriber
    /// constructed by `init` includes a [`fmt::Layer`] for logging events to
    /// stdout. What `tracing` events that layer will log is determined by the
    /// value of an environment variable; this method configures which
    /// environment variable is read to determine the log filter.
    ///
    /// This environment variable does not effect what spans and events are
    /// recorded by the [`ConsoleLayer`]. Therefore, this method will have no
    /// effect if the builder is used with [`Builder::spawn`] or
    /// [`Builder::build`].
    ///
    /// The default environment variable is `RUST_LOG`. See [here] for details
    /// on the syntax for configuring the filter.
    ///
    /// [`fmt::Layer`]: https://docs.rs/tracing-subscriber/0.3/tracing_subscriber/fmt/index.html
    /// [here]: https://docs.rs/tracing-subscriber/0.3/tracing_subscriber/filter/targets/struct.Targets.html
```

hawkw (changes requested):
Overall, the implementation looks right. I had some naming and documentation suggestions.

It would probably also be good to mention something about this method in the documentation for `Builder::init`, where it discusses the `RUST_LOG` environment variable: 
https://github.com/tokio-rs/console/blob/a77c6f4b6fd8585f219f4cb582c1e07938d6d8c2/console-subscriber/src/builder.rs#L205-L209

and maybe also here: 
https://github.com/tokio-rs/console/blob/a77c6f4b6fd8585f219f4cb582c1e07938d6d8c2/console-subscriber/src/builder.rs#L237

it would be good if that documentation mentioned that the environment variable can be changed using this method.

Thanks for working on this; once the docs and naming changes are addressed, I think this will be good to merge!

hawkw (approved):
i went ahead and made the remaining changes, this LGTM!

PR #277, opened 2022-02-09
api: move timestamp proto file

Freax13:
This PR moves `console-api/proto/google/protobuf/timestamp/timestamp.proto` to `console-api/proto/google/protobuf/timestamp.proto` to match the directory structure in https://github.com/protocolbuffers/protobuf.
This may not be important for generating code for Rust, however other languages such as Javascript need the path to match the official definitions in order to work with the official [google-protobuf](https://www.npmjs.com/package/google-protobuf) library.


hawkw (approved):
this seems fine to me!

i have to admit, i'm curious about why you're generating JavaScript bindings for the console protobufs...working on something exciting? :)

Freax13:
> i have to admit, i'm curious about why you're generating JavaScript bindings for the console protobufs...working on something exciting? :)

I'm generating TypeScript bindings for a console clone running in the browser. I already got a basic version of the overview table working :)

Screenshot from 2022-02-10 08-00-25

Given that the console cli is already very good, I'm not sure if there would be enough interest for it to be worth publishing. It's a nice exercise to learn about how tokio-console works though.

PR #282, opened 2022-02-14
fix(subscriber): fix compilation on targets without 64-bit atomics

### LINKED ISSUE #279: Support for platforms where AtomicU64 does not exist
When cross-compile to mipsel I get this error:

```
$ cross build --release --target mipsel-unknown-linux-musl
...
error[E0432]: unresolved import `std::sync::atomic::AtomicU64`
 --> /cargo/registry/src/github.com-1ecc6299db9ec823/console-subscriber-0.1.2/src/stats.rs:8:26
  |
8 |     atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering::*},
  |                          ^^^^^^^^^
  |                          |
  |                          no `AtomicU64` in `sync::atomic`
  |                          help: a similar name exists in the module: `AtomicU8`
```

See also:

https://doc.rust-lang.org/std/sync/atomic/index.html#portability

hawkw:
## Motivation

Currently, the `console-subscriber` crate does not compile on targets
without 64-bit atomic operations, such as `mipsel-unknown-linux-musl`.
This is because a 64-bit atomic is used to store the last task ID that
polled a resource.

## Solution

This branch changes this code to use the `AtomicCell` type from
`crossbeam-utils`, instead. `AtomicCell` will use atomic operations on
targets that support 64-bit atomics, and fall back to using a mutex on
platforms that don't have 64-bit atomic operations. This should fix the
build on those platforms.

Fixes #279


hawkw:
@spacemeowx2, would you mind checking out this branch and verifying that it does, in fact, fix the build issues you reported in #279? I wasn't able to easily verify the fix, as I'm running NixOS, and the `cross` tool doesn't currently work on NixOS (cross-rs/cross#524).

Thank you!

spacemeowx2:
It compiles with e0afa98801a74cc8362349935a6245e71ffa9f6f

Thanks!

hawkw:
@spacemeowx2 great, thanks for checking that for me!

PR #284, opened 2022-02-16
add information on where to put .cargo/config.toml

Frederik-Baetens:
I ran into an issue where putting my .cargo/config.toml with the tokio_unstable rustflag in the sub-project of my workspace where I needed to use tokio-console led to full rebuilds on every cargo build invocation.

This was caused by rust-analyzer invoking its cargo commands in the workspace root independent of the directory where the editor (e.g. nvim) is launched from. This led to rust-analyzer not to use the cargo config, while my manual cargo build invocations did use the config, hence requiring full rebuilds on every run.

This documentation informs new users of this pitfall.

If this documentation is found to be a little on the long side, I can try to move part of this warning into the cargo documentation, which I'm linking to anyways.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,11 +104,15 @@ notes:
   ```shell
   $ RUSTFLAGS="--cfg tokio_unstable" cargo build
   ```
-  or add the following to your `.cargo/config` file:
+  or add the following to your `.cargo/config.toml` file:
   ```toml
   [build]
   rustflags = ["--cfg", "tokio_unstable"]
   ```
+
+  For more information on the appropriate location of your `.cargo/config.toml` file,
+  especially when using workspaces, see the
+  [console-subscriber readme](console-subscriber/README.md#enabling-tokio-instrumentation).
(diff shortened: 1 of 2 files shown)


hawkw:
Hi @Frederik-Baetens, thanks for working on this!

The `console-subscriber` crate's README also has a section on how to enable the console instrumentation: https://github.com/tokio-rs/console/blob/main/console-subscriber/README.md#enabling-tokio-instrumentation

I think this should be documented there, as well, since that README will also be present in `console-subscriber`'s documentation on crates.io and docs.rs, while the repo's root README is only visible if you're actually looking at this repository on GitHub.

Perhaps, rather than adding more detailed information to the root README, we should add it to the `console-subscriber` README instead, and have a link in the root README  that says "see the `console-subscriber` documentation for more details on configuring RUSTFLAGS in workspaces" or something? What do you think?

Frederik-Baetens:
I've now moved this information, and added a link to it in the main readme.

hawkw (approved):
looks great, thank you!

PR #285, opened 2022-02-16
docs: fix broken links in READMEs and subscriber doc comment

thombles:
Documentation about setting the default subscriber in tracing and the
builder interface for console-subscriber moved, resulting in a few
broken links. Updates those links to their current locations.


hawkw (approved):
thanks, this looks good to me!

PR #289, opened 2022-02-17
fix(subscriber): record timestamps for updates last

### LINKED ISSUE #266: The console randomly die with an unwrap
The console randomly die after some time. Running with `RUST_LOG=debug`  and `RUST_BACKTRACE=full` produce this log file: [console.txt](https://github.com/tokio-rs/console/files/7924035/console.txt) (I suggest printing the log file in a terminal so you get the colors and other formatting)

## Error
```
called `Result::unwrap()` on an `Err` value: SystemTimeError(836.154µs)
```

## Metadata
|key|value|
|--|--|
|**version**|0.1.1|
|**location**|tokio-console/src/state/async_ops.rs:214:61|

hawkw:
## Motivation

Currently, when constructing an update message to send over the wire, a
timestamp is taken first, and then the protobuf data is constructed.
This can lead to issues where the "now" timestamp is actually _before_
timestamps present in the stats sent in that update, since the stats for
a particular task/resource/async op might be updated on another thread
after taking the update's "now" timestamp. This results in issues like
#266.

## Solution

There's no actual reason to take those timestamps *before* we assemble
the update. This branch changes the aggregator to build all the various
data updates in an update message, and *then* record the update's "now"
timestamp. Any timestamps for tasks/resources/async ops that are
recorded after the update's "now" timestamp will now be included in the
*next* update.

Fixes #266
Depends on #290

### DIFF
--- a/console-subscriber/src/aggregator/id_data.rs
+++ b/console-subscriber/src/aggregator/id_data.rs
@@ -7,6 +7,7 @@ pub(crate) struct IdData<T> {
     data: ShrinkMap<Id, T>,
 }
 
+#[derive(Copy, Clone, Eq, PartialEq)]
 pub(crate) enum Include {
     All,
     UpdatedOnly,
@@ -43,6 +44,23 @@ impl<T: Unsent> IdData<T> {
 
     pub(crate) fn get(&self, id: &Id) -> Option<&T> {
         self.data.get(id)
+    }
+
+    pub(crate) fn as_proto_list(
+        &mut self,
+        include: Include,
+        base_time: &TimeAnchor,
+    ) -> Vec<T::Output>
+    where
+        T: ToProto,
+    {
+        match include {
+            Include::UpdatedOnly => self
+                .since_last_update()
+                .map(|(_, d)| d.to_proto(base_time))
+                .collect(),
+            Include::All => self.all().map(|(_, d)| d.to_proto(base_time)).collect(),
+        }
     }
 
     pub(crate) fn as_proto(
--- a/console-subscriber/src/aggregator/mod.rs
+++ b/console-subscriber/src/aggregator/mod.rs
@@ -261,44 +261,52 @@ impl Aggregator {
     fn add_instrument_subscription(&mut self, subscription: Watch<proto::instrument::Update>) {
         tracing::debug!("new instrument subscription");
         let now = Instant::now();
-        // Send the initial state --- if this fails, the subscription is already dead
+
+        let task_update = Some(self.task_update(Include::All));
+        let resource_update = Some(self.resource_update(Include::All));
+        let async_op_update = Some(self.async_op_update(Include::All));
         let update = &proto::instrument::Update {
-            task_update: Some(proto::tasks::TaskUpdate {
-                new_tasks: self
-                    .tasks
-                    .all()
-                    .map(|(_, value)| value.to_proto(&self.base_time))
-                    .collect(),
-                stats_update: self.task_stats.as_proto(Include::All, &self.base_time),
-                dropped_events: self.shared.dropped_tasks.swap(0, AcqRel) as u64,
-            }),
-            resource_update: Some(proto::resources::ResourceUpdate {
-                new_resources: self
-                    .resources
-                    .all()
-                    .map(|(_, value)| value.to_proto(&self.base_time))
-                    .collect(),
-                stats_update: self.resource_stats.as_proto(Include::All, &self.base_time),
-                new_poll_ops: (*self.all_poll_ops).clone(),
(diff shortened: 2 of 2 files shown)


seanmonstar:
> There's no actual reason to take those timestamps before we assemble the update.

I'd think the reason for taking the timestamp first is because all the other stuff can take some time to compute, and we want to record as close as possible when the event actually happened, not when we finished understanding what the event means.

hawkw:
> > There's no actual reason to take those timestamps before we assemble the update.
> 
> I'd think the reason for taking the timestamp first is because all the other stuff can take some time to compute, and we want to record as close as possible when the event actually happened, not when we finished understanding what the event means.

Right, but the issue is that converting the data to the protobuf format takes time, and the data may be updated while we're doing that. If the data is updated by another thread after when we take the "now" timestamp and when we actually construct the protobuf representation of that data, it might have a timestamp that occurs _after_ "now". #266 explains why this is a problem.

We could, alternatively, handle this kind of thing in the console UI; this _felt_ like a better solution to me, though, since it wouldn't have to be solved again in another consumer of the wire format.

hawkw:
I went ahead and split out the refactoring into a separate branch (#290) so we can go ahead and merge that regardless of what we decide is the best way to fix the bug.

seanmonstar (approved):

thombles:
I've been fiddling with this crash in parallel (I can trigger it pretty reliably) and was investigating the same change. I tried checking out the tip of this branch (713a7befc3dcb9b7dae5a6e5b4fbe220e47a6eb6) and I'm still getting a similar crash/backtrace sadly.

[console-crash-713a7befc3dcb9b7dae5a6e5b4fbe220e47a6eb6.txt](https://github.com/tokio-rs/console/files/8092430/console-crash-713a7befc3dcb9b7dae5a6e5b4fbe220e47a6eb6.txt)

hawkw:
> I've been fiddling with this crash in parallel (I can trigger it pretty reliably) and was investigating the same change. I tried checking out the tip of this branch ([713a7be](https://github.com/tokio-rs/console/commit/713a7befc3dcb9b7dae5a6e5b4fbe220e47a6eb6)) and I'm still getting a similar crash/backtrace sadly.
> 
> [console-crash-713a7befc3dcb9b7dae5a6e5b4fbe220e47a6eb6.txt](https://github.com/tokio-rs/console/files/8092430/console-crash-713a7befc3dcb9b7dae5a6e5b4fbe220e47a6eb6.txt)

Hmm, that's a shame that this didn't fix it. Can you share how you're able to reliably trigger the panic?

thombles:
Not really, sorry. My app under test just has 50–70 long-running tasks ticking along mostly idle and it crashes within 10 minutes or so looking at at the tasks list view.

thombles:
If it helps account for any particular skew, the instrumented device is an iPhone I'm connected to over WiFi, with console running on my Mac.

hawkw:
@thombles just to confirm, the iPhone's `console-subscriber` dependency is from this branch on commit 713a7befc3dcb9b7dae5a6e5b4fbe220e47a6eb6? Not just the version of the console UI running on your mac?

hawkw:
Oh, one other question: when the panic occurs, do you know if it's _always_ in `AsyncOpsState::retain_active`, or does it ever panic in `TasksState::retain_active` or `ResourcesState::retain_active`?

thombles:
> @thombles just to confirm, the iPhone's `console-subscriber` dependency is from this branch on commit [713a7be](https://github.com/tokio-rs/console/commit/713a7befc3dcb9b7dae5a6e5b4fbe220e47a6eb6)? Not just the version of the console UI running on your mac?

Whoops, great guess - yes I forgot to update the subscriber end and obviously this code is in the subscriber crate. Sorry to mislead. I've now updated both device to `main` and it's rock solid. It does make me wonder whether console should defensively do a saturating duration_since just in case it gets bogus data but I guess it's not critical?

> when the panic occurs, do you know if it's _always_ in `AsyncOpsState::retain_active`, or does it ever panic in `TasksState::retain_active` or `ResourcesState::retain_active`?

So far I've only noticed it in the `AsyncOpsState` but I haven't analysed whether there's any special reason for that.

hawkw:
@thombles 
> Whoops, great guess - yes I forgot to update the subscriber end and obviously this code is in the subscriber crate. Sorry to mislead. I've now updated both device to `main` and it's rock solid.

Ah, yup, there's the problem --- thanks for double-checking. So, this change probably does fix the issue, which is good, because I felt like I was losing my mind there for a second. :)

> It does make me wonder whether console should defensively do a saturating duration_since just in case it gets bogus data but I guess it's not critical?

Yeah, it probably should — we should give the whole console UI a pass for potential panics in the face of weird data from the instrumented application.

PR #291, opened 2022-02-18
docs: document minimum Tokio versions

### LINKED ISSUE #281: console docs should prominently state required version of tokio runtime
I've seen some tokio users hypothesize that console "works" on old versions of tokio. (We're talking version 0.1 here.) After being corrected, they responded that maybe the minimum version should be mentioned more prominently.

I skimmed https://tokio.rs/blog/2021-12-announcing-tokio-console and didn't see any mention of the minimum tokio version required for use.

Then I looked in the docs for tokio-console itself, and saw no mention of the minimum tokio version required:
https://docs.rs/tokio-console/latest/tokio_console/

Then I skimmed the README for the tokio-console github repository, which is the biggest thing ones sees when visiting the repository, and saw no mention of the minimum tokio version required: 
https://github.com/tokio-rs/console/blob/main/README.md

I *do* see that the minimum version is mentioned with the console_subscriber docs:
https://docs.rs/console-subscriber/latest/console_subscriber/#getting-started

The problem is that someone has to be someone deep into the decision to *use*  tokio-console before they might potentially see that.

Now, I personally know that the nature of the tokio-console architecture means that the minimum tokio version is really a property of the console_subscriber crate, not so much a property of the tokio-console, and that readily explains how the documentation ended up this way.

But I think newcomers to tokio-console cannot be expected to have that level of insight.

It would probably be good for the minimum tokio runtime version to be mentioned someone on the common landing pages for someone investigating whether tokio-console can be used with their application: i.e., the github README linked above, and in the tokio-console docs.

 * It could include the caveat there that this tokio-version requirement can be lifted for people who somehow inject tokio-console telemetry without using console-subscriber. But the common expected case is for people to use console-subscriber, and it would be good to directly address that scenario by stating the requirement inline.

hawkw:
This branch adds documentation to the `console-subscriber` _and_
`tokio-console` READMEs and RustDoc explaining the minimum Tokio
versions required by the console. Since instrumentation for different
aspects of the runtime has been added over time, this isn't as simple as
"the console requires version x.y.z or later"; instead, we provide a
list of which console features require which Tokio versions. The console
should be capable of at least limited operation with any Tokio version
>= 1.0, but some data will not be available (such as wakers, resource
>instrumentation, etc).

Closes #281

### DIFF
--- a/console-subscriber/README.md
+++ b/console-subscriber/README.md
@@ -114,7 +114,7 @@ Other instrumentation is added in later Tokio releases:
   as `Mutex`es, `RwLock`s, `Semaphore`s, `oneshot` channels, `mpsc` channels, et
   cetera. 
      
-[Tokio v1.0.0]:
+[Tokio v1.0.0]: https://github.com/tokio-rs/tokio/releases/tag/tokio-1.0.0
 [Tokio v1.7.0]: https://github.com/tokio-rs/tokio/releases/tag/tokio-1.7.0
 [Tokio v1.12.0]:https://github.com/tokio-rs/tokio/releases/tag/tokio-1.12.0
 [`Runtime::block_on`]: https://docs.rs/tokio/1/tokio/runtime/struct.Runtime.html#method.block_on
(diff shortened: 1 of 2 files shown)


pnkfelix:
awesome, this went above and beyond my expectations here!

PR #295, opened 2022-02-20
fix(console): prevent panics if subscriber reports out-of-order times

thombles:
Changes duration_since and unchecked subtractions to assume
Duration::ZERO (via Default) if the alternative would be a crash.

If the subscriber reports nonsensical data this can result in negative
Durations in console's calculations, leading to panics. This has already
happened inadvertently. All changed code relates directly to received
protobuf data whose validity isn't guaranteed.

Relates to #266 (crash) and #289 (fix for root cause)

---

There are a handful of other unwraps/expects I'm not so sure about. These ones looked like pretty clear-cut cases where panicking is possible, undesirable, and clamping to zero is a reasonable response.

### DIFF
--- a/tokio-console/src/state/async_ops.rs
+++ b/tokio-console/src/state/async_ops.rs
@@ -211,7 +211,7 @@ impl AsyncOpsState {
                 .stats
                 .dropped_at
                 .map(|d| {
-                    let dropped_for = now.duration_since(d).unwrap();
+                    let dropped_for = now.duration_since(d).unwrap_or_default();
                     retain_for > dropped_for
                 })
                 .unwrap_or(true)
@@ -245,25 +245,29 @@ impl AsyncOp {
     }
 
     pub(crate) fn total(&self, since: SystemTime) -> Duration {
-        self.stats
-            .total
-            .unwrap_or_else(|| since.duration_since(self.stats.created_at).unwrap())
+        self.stats.total.unwrap_or_else(|| {
+            since
+                .duration_since(self.stats.created_at)
+                .unwrap_or_default()
+        })
     }
 
     pub(crate) fn busy(&self, since: SystemTime) -> Duration {
         if let (Some(last_poll_started), None) =
             (self.stats.last_poll_started, self.stats.last_poll_ended)
         {
-            let current_time_in_poll = since.duration_since(last_poll_started).unwrap();
+            let current_time_in_poll = since.duration_since(last_poll_started).unwrap_or_default();
             return self.stats.busy + current_time_in_poll;
         }
         self.stats.busy
     }
 
     pub(crate) fn idle(&self, since: SystemTime) -> Duration {
-        self.stats
-            .idle
-            .unwrap_or_else(|| self.total(since) - self.busy(since))
+        self.stats.idle.unwrap_or_else(|| {
+            self.total(since)
+                .checked_sub(self.busy(since))
+                .unwrap_or_default()
+        })
     }
 
     pub(crate) fn total_polls(&self) -> u64 {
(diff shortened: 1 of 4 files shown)


tokio-console/src/state/async_ops.rs
```diff
@@ -245,25 +245,29 @@ impl AsyncOp {
...
 
     pub(crate) fn total(&self, since: SystemTime) -> Duration {
-        self.stats
-            .total
-            .unwrap_or_else(|| since.duration_since(self.stats.created_at).unwrap())
+        self.stats.total.unwrap_or_else(|| {
+            since
+                .duration_since(self.stats.created_at)
+                .unwrap_or_default()
+        })
```
hawkw: nit, take it or leave it: this might look a bit nicer as
```suggestion
        self.stats
            .total
            .or_else(|| since.duration_since(self.stats.created_at).ok())
            .unwrap_or_default()
```

tokio-console/src/state/async_ops.rs
```diff
@@ -245,25 +245,29 @@ impl AsyncOp {
...
 
     pub(crate) fn idle(&self, since: SystemTime) -> Duration {
-        self.stats
-            .idle
-            .unwrap_or_else(|| self.total(since) - self.busy(since))
+        self.stats.idle.unwrap_or_else(|| {
+            self.total(since)
+                .checked_sub(self.busy(since))
+                .unwrap_or_default()
+        })
```
hawkw: nit, take it or leave it: again, we could move the `unwrap_or_default` out using an `or_else` instead, which i think maybe reads a little nicer?
```suggestion
        self.stats
            .idle
            .or_else(|| self.total(since).checked_sub(self.busy(since)))
            .unwrap_or_default()
```

tokio-console/src/state/resources.rs
```diff
@@ -296,9 +296,11 @@ impl Resource {
...
 
     pub(crate) fn total(&self, since: SystemTime) -> Duration {
-        self.stats
-            .total
-            .unwrap_or_else(|| since.duration_since(self.stats.created_at).unwrap())
+        self.stats.total.unwrap_or_else(|| {
+            since
+                .duration_since(self.stats.created_at)
+                .unwrap_or_default()
+        })
```
hawkw, line 303: nit, take it or leave it: consider
```suggestion
        self.stats
            .total
            .or_else(|| since.duration_since(self.stats.created_at).ok())
            .unwrap_or_default()
```

tokio-console/src/state/tasks.rs
```diff
@@ -279,26 +279,30 @@ impl Task {
...
 
     pub(crate) fn total(&self, since: SystemTime) -> Duration {
-        self.stats
-            .total
-            .unwrap_or_else(|| since.duration_since(self.stats.created_at).unwrap())
+        self.stats.total.unwrap_or_else(|| {
+            since
+                .duration_since(self.stats.created_at)
+                .unwrap_or_default()
+        })
```
hawkw: nit, take it or leave it: consider:
```suggestion
        self.stats
            .total
            .or_else(|| since.duration_since(self.stats.created_at).ok())
            .unwrap_or_default()
```

tokio-console/src/state/tasks.rs
```diff
@@ -279,26 +279,30 @@ impl Task {
...
 
     pub(crate) fn idle(&self, since: SystemTime) -> Duration {
-        self.stats
-            .idle
-            .unwrap_or_else(|| self.total(since) - self.busy(since))
+        self.stats.idle.unwrap_or_else(|| {
+            self.total(since)
+                .checked_sub(self.busy(since))
+                .unwrap_or_default()
+        })
```
hawkw: nit, take it or leave it: consider
```suggestion
        self.stats
            .idle
            .or_else(|| self.total(since).checked_sub(self.busy(since)))
            .unwrap_or_default()
```

hawkw (approved):
overall, this looks good to me --- i left some minor style suggestions, but it's up to you whether you think my suggested code is more readable or not. thanks!

Issue #300, opened 2022-03-04
Want to represent sorting state


nrskt:
I want to display whether the selected columns are in ascending or descending order like as htop.

hawkw:
Closed by #301

PR #301, opened 2022-03-04
feat(console): Add icon that sorting state

nrskt:
Add `▵` or `▿` icon that represents the sorting state.

|  |  |
| ---- | ---- |
| Screenshot_20220304_183030 | Screenshot_20220304_183046 |

Relates to #300

### DIFF
--- a/tokio-console/src/view/resources.rs
+++ b/tokio-console/src/view/resources.rs
@@ -20,12 +20,12 @@ use tui::{
 #[derive(Debug, Default)]
 pub(crate) struct ResourcesTable {}
 
-impl TableList for ResourcesTable {
+impl TableList<9> for ResourcesTable {
     type Row = Resource;
     type Sort = SortBy;
     type Context = ();
 
-    const HEADER: &'static [&'static str] = &[
+    const HEADER: &'static [&'static str; 9] = &[
         "ID",
         "Parent",
         "Kind",
@@ -37,8 +37,20 @@ impl TableList for ResourcesTable {
         "Attributes",
     ];
 
+    const WIDTHS: &'static [usize; 9] = &[
+        Self::HEADER[0].len() + 1,
+        Self::HEADER[1].len() + 1,
+        Self::HEADER[2].len() + 1,
+        Self::HEADER[3].len() + 1,
+        Self::HEADER[4].len() + 1,
+        Self::HEADER[5].len() + 1,
+        Self::HEADER[6].len() + 1,
+        Self::HEADER[7].len() + 1,
+        Self::HEADER[8].len() + 1,
+    ];
+
     fn render<B: tui::backend::Backend>(
-        table_list_state: &mut TableListState<Self>,
+        table_list_state: &mut TableListState<Self, 9>,
         styles: &view::Styles,
         frame: &mut tui::terminal::Frame<B>,
         area: layout::Rect,
@@ -59,15 +71,15 @@ impl TableList for ResourcesTable {
             .sort_by
             .sort(now, &mut table_list_state.sorted_items);
 
-        let viz_len: u16 = Self::HEADER[6].len() as u16;
+        let viz_len: u16 = Self::WIDTHS[6] as u16;
 
-        let mut id_width = view::Width::new(Self::HEADER[0].len() as u16);
-        let mut parent_width = view::Width::new(Self::HEADER[1].len() as u16);
+        let mut id_width = view::Width::new(Self::WIDTHS[0] as u16);
+        let mut parent_width = view::Width::new(Self::WIDTHS[1] as u16);
 
-        let mut kind_width = view::Width::new(Self::HEADER[2].len() as u16);
-        let mut target_width = view::Width::new(Self::HEADER[4].len() as u16);
-        let mut type_width = view::Width::new(Self::HEADER[5].len() as u16);
-        let mut location_width = view::Width::new(Self::HEADER[7].len() as u16);
+        let mut kind_width = view::Width::new(Self::WIDTHS[2] as u16);
+        let mut target_width = view::Width::new(Self::WIDTHS[4] as u16);
+        let mut type_width = view::Width::new(Self::WIDTHS[5] as u16);
+        let mut location_width = view::Width::new(Self::WIDTHS[7] as u16);
 
         let rows = {
             let id_width = &mut id_width;
@@ -120,22 +132,22 @@ impl TableList for ResourcesTable {
                 })
         };
 
-        let (selected_style, header_style) = if let Some(cyan) = styles.color(Color::Cyan) {
-            (Style::default().fg(cyan), Style::default())
+        let header_style = if styles.color(Color::Cyan).is_some() {
+            Style::default()
         } else {
-            (
-                Style::default().remove_modifier(style::Modifier::REVERSED),
-                Style::default().add_modifier(style::Modifier::REVERSED),
-            )
+            Style::default().add_modifier(style::Modifier::REVERSED)
         };
         let header_style = header_style.add_modifier(style::Modifier::BOLD);
 
         let header = Row::new(Self::HEADER.iter().enumerate().map(|(idx, &value)| {
-            let cell = Cell::from(value);
             if idx == table_list_state.selected_column {
-                cell.style(selected_style)
+                if table_list_state.sort_descending {
+                    Cell::from(styles.ascending(value))
+                } else {
+                    Cell::from(styles.descending(value))
+                }
             } else {
-                cell
+                Cell::from(value)
             }
         }))
         .height(1)
--- a/tokio-console/src/view/styles.rs
+++ b/tokio-console/src/view/styles.rs
@@ -127,7 +127,7 @@ impl Styles {
         )
     }
 
-    pub fn selected<'a>(&self, value: &str) -> Span<'static> {
+    pub fn selected(&self, value: &str) -> Span<'static> {
         let style = if let Some(cyan) = self.color(Color::Cyan) {
             Style::default().fg(cyan)
         } else {
--- a/tokio-console/src/view/styles.rs
+++ b/tokio-console/src/view/styles.rs
@@ -127,6 +127,25 @@ impl Styles {
         )
     }
 
+    pub fn selected<'a>(&self, value: &str) -> Span<'static> {
+        let style = if let Some(cyan) = self.color(Color::Cyan) {
+            Style::default().fg(cyan)
+        } else {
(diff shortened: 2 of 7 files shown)


tokio-console/src/view/tasks.rs
```diff
@@ -149,11 +161,15 @@ impl TableList for TasksTable {
         let header_style = header_style.add_modifier(style::Modifier::BOLD);
 
         let header = Row::new(Self::HEADER.iter().enumerate().map(|(idx, &value)| {
-            let cell = Cell::from(value);
             if idx == table_list_state.selected_column {
-                cell.style(selected_style)
+                let suffix = if table_list_state.sort_descending {
+                    "▵"
+                } else {
+                    "▿"
```
hawkw: The console supports running in an ASCII-only mode, so we need to be able to switch between these Unicode characters and ASCII-only text. For other Unicode characters we use as icons, we use the `if_utf8` method on the `Styles` type to switch between Unicode and ASCII variants:
https://github.com/tokio-rs/console/blob/900a5c2bd5b610e9b939a5f824af1ac1a11267d0/tokio-console/src/view/styles.rs#L57-L63

Here, I think we should do something like this:
```suggestion
                    styles.if_utf8("▵", "+")
                } else {
                   styles.if_utf8("▿", "-") 
```
using `+` for ascending and `-` for descending in ASCII-only mode. alternatively, we could use something like `(asc.)`/`(desc.)` for ascending/descending, if we don't think a single character is descriptive enough...

note that if we use multiple characters in the ASCII-only mode, we'll also have to add that much additional length to the column width, rather than just adding a single character in every case.

tokio-console/src/view/tasks.rs
```diff
@@ -72,6 +72,18 @@ impl TableList for TasksTable {
...
+        match table_list_state.sort_by {
+            SortBy::Warns => warn_width.update_len(warn_width.len() + 1),
+            SortBy::Tid => id_width.update_len(id_width.len() + 1),
+            SortBy::State => state_len += 1,
+            SortBy::Name => name_width.update_len(name_width.len() + 1),
+            SortBy::Polls => polls_width.update_len(polls_width.len() + 1),
+            SortBy::Target => target_width.update_len(target_width.len() + 1),
+            SortBy::Location => location_width.update_len(location_width.len() + 1),
+            SortBy::Total | SortBy::Busy | SortBy::Idle => (),
+        };
```
hawkw: one note about this approach is that it means that changing the sorted column has the potential to make that column one character wider than it was previously, if none of the text in the column was longer than the header. i feel like this has the potential to look a little bit jarring if selecting a new column to sort by makes the columns shift around. it might honestly be better to just always add an additional character of width to every column?

but, i'm definitely willing to be convinced otherwise.

hawkw (changes requested):
This is definitely a good idea, thanks for working on it!

The main thing I think we need to change before this is ready to merge is supporting ASCII-only mode, which I left a comment about. Besides that, this seems more or less correct --- I think it might be worth changing how we add extra length to column titles, so that we don't change column width as often, but that's up to you.

tokio-console/src/view/tasks.rs
nrskt: Thank you for the clear explanation. I will fix it.

tokio-console/src/view/tasks.rs
nrskt: > i feel like this has the potential to look a little bit jarring if selecting a new column to sort by makes the columns shift around. it might honestly be better to just always add an additional character of width to every column?

I agree. I feel it would be more clear to add an additional character of width to every column. However, I am confused because there are a few places in the current implementation where the width is adjusted. I would like to fix it at another opportunity and understand the whole implementation a little bit.

hawkw: We can always address that later!

tokio-console/src/view/tasks.rs
nrskt: I fixed c29d6c95c7a529fb8713141ccb7630fe5240e916.

hawkw (changes requested):
oh, I just realized something else: we should make the same change for the resource list view (https://github.com/tokio-rs/console/blob/main/tokio-console/src/view/resources.rs) and the async ops list view (https://github.com/tokio-rs/console/blob/main/tokio-console/src/view/async_ops.rs). that should be pretty straightforward, but it would be nice to add this to every table view as part of this change.

since we'll be using the same icons for ascending and descending sorts on all three views, we may want to just add methods to the `Styles` type for returning the ascending and descending icons based on whether or not Unicode is enabled. There are existing methods on `Styles` for other commonly-used symbols, like the warning symbol: https://github.com/tokio-rs/console/blob/b4e6b6a61eb8fc03a623c7d33a98830c2109fe0d/tokio-console/src/view/styles.rs#L123-L128

nrskt:
`cargo clippy` raise a few warnings of `unnecessary use of to_owned`. I have tried to fix it but caused compile error...
Should I fix it? If so, how to fix it?

hawkw:
> `cargo clippy` raise a few warnings of `unnecessary use of to_owned`. I have tried to fix it but caused compile error... Should I fix it? If so, how to fix it?

This is fixed on the `main` branch, by PR #305. I'll update this branch from `main` and we should get a passing CI build now.

tokio-console/src/view/tasks.rs
nrskt: I realize an idea.

1. Add const WIDTHS param to TableList trait definition.
2. Set the widths on the implementation.

```rust
pub(crate) trait TableList<const N: usize> {
    type Row;
    type Sort: SortBy + TryFrom<usize>;
    type Context;

    const HEADER: &'static [&'static str; N];
    const WIDTHS: &'static [usize; N];

    fn render<B: tui::backend::Backend>(
        state: &mut TableListState<Self, N>,
        styles: &view::Styles,
        frame: &mut tui::terminal::Frame<B>,
        area: layout::Rect,
        state: &mut state::State,
        cx: Self::Context,
    ) where
        Self: Sized;
}
```

```rust
impl TableList<11> for TasksTable {
    type Row = Task;
    type Sort = SortBy;
    type Context = ();

    const HEADER: &'static [&'static str; 11] = &[
        "Warn", "ID", "State", "Name", "Total", "Busy", "Idle", "Polls", "Target", "Location",
        "Fields",
    ];

    const WIDTHS: &'static [usize; 11] = &[
        Self::HEADER[0].len() + 1,
        Self::HEADER[1].len() + 1,
        Self::HEADER[2].len() + 1,
        Self::HEADER[3].len() + 1,
        Self::HEADER[4].len() + 1,
        Self::HEADER[5].len() + 1,
        Self::HEADER[6].len() + 1,
        Self::HEADER[7].len() + 1,
        Self::HEADER[8].len() + 1,
        Self::HEADER[9].len() + 1,
        Self::HEADER[10].len() + 1,
    ];
```

What do you think it?
(I worry about implement WIDTHS param is complex and boring)

nrskt:
@hawkw 

- I tried the idea about (https://github.com/tokio-rs/console/pull/301#discussion_r820071449).
- Fix resources and async_ops tables.
  - If I misunderstood your advice (https://github.com/tokio-rs/console/pull/301#pullrequestreview-900668895), feel free to point it out

Screenshot_20220305_185330
Screenshot_20220305_185359
Screenshot_20220305_185415

tokio-console/src/view/styles.rs
```diff
@@ -127,6 +127,25 @@ impl Styles {
...
+        Span::styled(value.to_string(), style)
+    }
+
+    pub fn ascending(&self, value: &str) -> Span<'static> {
+        let value = format!("{}{}", value, self.if_utf8("▵", "+"));
+        self.selected(&value)
+    }
+
+    pub fn descending(&self, value: &str) -> Span<'static> {
+        let value = format!("{}{}", value, self.if_utf8("▿", "-"));
```
hawkw, line 145: i think it might look a little bit nicer if we added a space after the symbol in the case where it's a plus or minus sign, but not a blocker.

tokio-console/src/view/resources.rs
```diff
@@ -59,15 +71,15 @@ impl TableList for ResourcesTable {
...
+        let mut parent_width = view::Width::new(Self::WIDTHS[1] as u16);
 
-        let mut kind_width = view::Width::new(Self::HEADER[2].len() as u16);
-        let mut target_width = view::Width::new(Self::HEADER[4].len() as u16);
-        let mut type_width = view::Width::new(Self::HEADER[5].len() as u16);
-        let mut location_width = view::Width::new(Self::HEADER[7].len() as u16);
+        let mut kind_width = view::Width::new(Self::WIDTHS[2] as u16);
+        let mut target_width = view::Width::new(Self::WIDTHS[4] as u16);
+        let mut type_width = view::Width::new(Self::WIDTHS[5] as u16);
+        let mut location_width = view::Width::new(Self::WIDTHS[7] as u16);
```
hawkw, line 82: The `WIDTHS` array seems fine to me, but, for what it's worth, think it would *also* be fine to just change this code to something like
```rust
        let viz_len: u16 = (Self::HEADER[6].len() + 1) as u16;

        let mut id_width = view::Width::new((Self::HEADER[0].len() + 1) as u16);
        let mut parent_width = view::Width::new((Self::HEADER[1].len() + 1) as u16);
        // ... and so on ...
```

Since the `HEADER` array is a const, and the `1` is also a constant, this calculation should probably get const-folded, so I don't think there would be a performance difference between this approach and creating a separate const array. I think the code might be a bit simpler if we avoided the separate array.

But, this is fine either way --- it's not a big deal.

hawkw (approved):
This looks good to me, thanks! 

I had some suggestions, but you can feel free to take them or leave them. Let me know what you think!

nrskt:
@hawkw 
Thank you for reviewing.
Your suggestions are very helpful to me. However, I would like to merge this PR.

I’m interested in this project and I hope I can reflect your suggestions in another PR.

hawkw:
> @hawkw Thank you for reviewing. Your suggestions are very helpful to me. However, I would like to merge this PR.
> 
> I’m interested in this project and I hope I can reflect your suggestions in another PR.

Sounds good, I'm happy to go ahead and merge this now!

tokio-console/src/view/styles.rs
```diff
@@ -127,6 +127,25 @@ impl Styles {
         )
     }
 
+    pub fn selected<'a>(&self, value: &str) -> Span<'static> {
```
hawkw: this should fix CI:
```suggestion
    pub fn selected(&self, value: &str) -> Span<'static> {
```

Issue #310, opened 2022-03-11
console: configuration file


hawkw:
### What problem are you trying to solve?

Currently, the `tokio-console` CLI honors a number of command-line arguments and environment variables to configure its behavior. These include view options, such as colors and UTF-8 support: https://github.com/tokio-rs/console/blob/3c559121e3c5ad175471718a3cf87ada0146a7cd/tokio-console/src/config.rs#L71-L107

In general, these kinds of configurations are typically not the sort of thing that users would want to pass a bunch of CLI flags for. Instead, users might want to set these configurations globally for their system.

### How should the problem be solved?

We should support a configuration file for configuring the `tokio-console` CLI.  

Currently, the main things we would set in a config file are view options like colors and Unicode support, but as we add more features, there may be additional configuration surface area. In particular, #148 proposes the ability to control what warnings are enabled/disabled; we would almost certainly want to be able to configure this from a config file.

### Any alternatives you've considered?

We could just not do this.

### How would users interact with this feature?

TOML is probably a good configuration language for our purposes: as it's widely used for Rust tools and most Rust developers are probably already familiar with it; and we don't need to represent particularly complex configurations. 

On Unix systems, we should probably use the [`XDG_CONFIG_HOME` environment variable](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) to determine where to look for the config file. This way, if the user's systems stores configs in a non-standard location, we'll be able to find the config file.

We may also want to support overriding parts of the configuration with a config file in the current working directory. That way, users could have per-project configs for different projects, and (potentially) check them into source control. This could allow different projects to do things like enable or disable different sets of warnings. We would probably want to combine local configs with global ones in a way where the local configs override the global config in the case of conflicts. This could be a separate issue once basic config file support is implemented.

### Would you like to work on this feature?

no

hawkw:
This would probably be a good first issue for anyone interested in helping contribute to the console.

nrskt:
i’d like to take it.

hawkw:
@nrskt That would be great! Please let me know if you have any questions!

nrskt:
@hawkw 

[Q1] Do you want to use the view options from the config file only? (Can I disable the options no_colors, ascii_only and etc?)

I think both options are realized is ideal. But it's complex for me because I don't know how to use the default value from the config file using clap derive-style.

First implementation:

- Disable arguments as follows.
    - no_colors
    - lang
    - ascii_only
    - truecolor
    - palette
[...]

PR #312, opened 2022-03-17
(WIP) Support configuration file

nrskt:
This pull request is a draft of the supporting configuration file.
I'm worried about implementations as follows.

- two kinds of types of configuration files are supported.
    - `$XDG_CONFIG_HOME/tokio-console/console.toml`
    - current configuration (`./console.toml`) 

I want some advice about these implementations.
(Should I rewrite clap builder api style?)

Relates to https://github.com/tokio-rs/console/issues/310

### DIFF
--- a/console.toml
+++ b/console.toml
@@ -1,0 +1,9 @@
+no_colors = false
+lang = "en_us.UTF-8"
+ascii_only = true
+palette = "8"
+
+
+[toggles]
+color_durations = true
+color_terminated = true
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -1,5 +1,9 @@
 use crate::view::Palette;
 use clap::{ArgGroup, Parser as Clap, ValueHint};
+use serde::{Deserialize, Serialize};
+use std::env;
+use std::fs;
+use std::path::Path;
 use std::process::Command;
 use std::str::FromStr;
 use std::time::Duration;
@@ -28,7 +32,7 @@ pub struct Config {
     #[clap(long = "log", env = "RUST_LOG", default_value = "off")]
     pub(crate) env_filter: tracing_subscriber::EnvFilter,
 
-    #[clap(flatten)]
+    #[clap(skip)]
     pub(crate) view_options: ViewOptions,
 
     /// How long to continue displaying completed tasks and dropped resources
@@ -66,7 +70,7 @@ pub struct Config {
 #[derive(Debug)]
 struct RetainFor(Option<Duration>);
 
-#[derive(Clap, Debug, Clone)]
+#[derive(Clap, Debug, Clone, Default, Deserialize, Serialize)]
 #[clap(group = ArgGroup::new("colors").conflicts_with("no-colors"))]
 pub struct ViewOptions {
     /// Disable ANSI colors entirely.
@@ -107,7 +111,7 @@ pub struct ViewOptions {
 }
 
 /// Toggles on and off color coding for individual UI elements.
-#[derive(Clap, Debug, Copy, Clone)]
+#[derive(Clap, Debug, Copy, Clone, Deserialize, Serialize)]
 pub struct ColorToggles {
     /// Disable color-coding for duration units.
     #[clap(long = "no-duration-colors", parse(from_flag = std::ops::Not::not), group = "colors")]
@@ -118,9 +122,37 @@ pub struct ColorToggles {
     pub(crate) color_terminated: bool,
 }
 
+impl Default for ColorToggles {
+    fn default() -> Self {
+        Self {
+            color_durations: true,
+            color_terminated: true,
+        }
+    }
+}
+
 // === impl Config ===
 
 impl Config {
+    pub fn from_config() -> color_eyre::Result<Self> {
+        let xdg_config_path = env::var_os("XDG_CONFIG_HOME").map(|mut base| {
+            base.push("/tokio-console/console.toml");
+            base
+        });
+        let xdg_view_opt = xdg_config_path.and_then(ViewOptions::from_config);
+        let current_view_opt = ViewOptions::from_config("console.toml");
+
+        let config = Config::parse();
+
+        match xdg_view_opt.or(current_view_opt) {
+            None => Ok(config),
+            Some(view_opt) => Ok(Self {
+                view_options: view_opt,
+                ..config
+            }),
+        }
+    }
+
     pub fn trace_init(&mut self) -> color_eyre::Result<()> {
         let filter = std::mem::take(&mut self.env_filter);
         use tracing_subscriber::prelude::*;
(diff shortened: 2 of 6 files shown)


tokio-console/src/config.rs
```diff
@@ -28,7 +32,7 @@ pub struct Config {
     #[clap(long = "log", env = "RUST_LOG", default_value = "off")]
     pub(crate) env_filter: tracing_subscriber::EnvFilter,
 
-    #[clap(flatten)]
+    #[clap(skip)]
```
nrskt, line 35: The reason I skipped it:
- I don't know how to use default values from config files.
- If the parameter is enabled, I should implement that:
    - If command-line options are given, then need to use them instead of the config file's value.

tokio-console/src/config.rs
```diff
@@ -118,9 +122,37 @@ pub struct ColorToggles {
...
+        let config = Config::parse();
+
+        match xdg_view_opt.or(current_view_opt) {
+            None => Ok(config),
+            Some(view_opt) => Ok(Self {
+                view_options: view_opt,
+                ..config
+            }),
+        }
+    }
```
nrskt, line 154: I'm worried:

- I don't know woking it on macOS or Windows.
- Some rules are broken.
    - ex) it can use `--palette` and `--colorterm`.

console.toml
```diff
@@ -0,0 +1,9 @@
+no_colors = false
+lang = "en_us.UTF-8"
+ascii_only = true
+palette = "8"
+
+
+[toggles]
+color_durations = true
+color_terminated = true
```
hawkw, line 9: Was this checked in for testing purposes? Eventually, I think it could be good to have an example config file with the default options (and comments explaining them) checked into the repo, so that users can see how to configure the console. But, I don't know if it should be "live" --- it might be better to name the example something like "console.example.toml" or similar, so that running the console from the repo directory doesn't actually use those settings...

console.toml
```diff
@@ -0,0 +1,9 @@
+no_colors = false
+lang = "en_us.UTF-8"
+ascii_only = true
+palette = "8"
```
hawkw, line 4: not sure if it makes sense for the config file to have separate keys for `no_colors` and `palette`, if we can also write `palette = "off"` to disable colors. i'd prefer to not have two separate ways to write the same thing in the config file.

console.toml
hawkw, line 9: I'm not sure if it makes sense to match the layout of the config structs exactly in the config file. instead, i think it might be  good to group the different configurations into TOML tables, maybe something like this:

```toml
[charset]
lang = "en_us.UTF-8"
ascii_only = false

[colors]
enabled = true
palette = "8"

[colors.enable]
durations = true
terminated = true
```
or something like that?

tokio-console/src/config.rs
```diff
@@ -118,9 +122,37 @@ pub struct ColorToggles {
...
+
+        let config = Config::parse();
+
+        match xdg_view_opt.or(current_view_opt) {
+            None => Ok(config),
+            Some(view_opt) => Ok(Self {
+                view_options: view_opt,
+                ..config
+            }),
+        }
```
hawkw, line 153: eventually, I think what we'll want is a function on the `Config` type to _merge_ two `Config`s, with one overriding the other. I think the behavior we actually want here is something like this:
- read the config file in the platform-specific config location (e.g. `XDG_CONFIG_HOME`) as the "base" config, or use the defaults if there is no config file there
- if there is a config file in the current directory, any configs set in that config file override the settings in the "base" config
- any command-line arguments that are set override the configs from both the current directory and the base directory

does that make sense?

tokio-console/src/config.rs
hawkw: > * I don't know woking it on macOS or Windows.

I think we might want to look into using the [`dirs` crate](https://lib.rs/crates/dirs) for reading configs from the correct location on multiple platforms. That library has a [`config_dir` function](https://docs.rs/dirs/latest/dirs/fn.config_dir.html) that returns the correct operating-system-specific config file location.

On macOS, we might also want to additionally check for `~/.console.toml` (and maybe also `~/.config/console.toml`), which are both Linux style locations --- although Apple suggests the use of `~/Library/Application Support` for config files, a lot of people still like to put config files for command-like utilities in `~` or `~.config/`...

hawkw (commented):
This looks like a good start. I left some suggestions inline.

One higher-level thought is that I'm not _sure_ if it really makes sense to have one set of structs that implements both `serde::Deserialize` _and_ `clap::Parser`. We might want to structure the command-line arguments and configuration files differently, so we may want to define separate types for reading configs from the command line and from a file, and then combine them into one internal config type that's used to actually configure the console.

On the other hand, it's possible that we can use `#[clap(...)]` and `#[serde(...)]` attributes to implement any differences between the structure of the command-line arguments and the config file. If this is practical, it might be a bit nicer to not have to convert between multiple structs! But, if the best solution requires us to have separate structs for the CLI and config file, I think that's fine.

tokio-console/src/config.rs
nrskt: `config_dir` is exactly what i have wanted. I'll use it.

console.toml
nrskt: Sorry I added it unintentionally. 
Eventually, I provide an example config file or add the explanation to README.

nrskt: That sounds good, I'll try it.

tokio-console/src/config.rs
nrskt: Yes, that makes sense. Ok, I'll try it again.

nrskt:
@hawkw Thank you for reviewing.

>  We might want to structure the command-line arguments and configuration files differently, so we may want to define separate types for reading configs from the command line and from a file, and then combine them into one internal config type that's used to actually configure the console.

Your input has inspired me to change my implementation approach.

I'll try as follows:

- Define the type of config from a file.
- Define merge config function.

On the other hand, I have a problem.

>  - read the config file in the platform-specific config location (e.g. XDG_CONFIG_HOME) as the "base" config, or use the defaults if there is no config file there
> - if there is a config file in the current directory, any configs set in that config file override the settings in the "base" config
> - any command-line arguments that are set override the configs from both the current directory and the base directory

I can implement the first and second points, but I am having difficulty with the third suggestion.
because I don't know how to identify the command-line argument set when using `clap::Parser::parse()` was used. I think if the value does not set, then use the default value that defines by `#[clap(...)]`. So, I don't know what value should be overridden. 

Do you have any ideas in this regard?

nrskt:
I'm trying to rewrite the config parser using clap Builder-API style. It will take some time because I'm doing it while checking to document it.

Maybe I'll submit it as another Pull Request. (Should I close this PR?)

hawkw:
> I can implement the first and second points, but I am having difficulty with the third suggestion. because I don't know how to identify the command-line argument set when using `clap::Parser::parse()` was used. I think if the value does not set, then use the default value that defines by `#[clap(...)]`. So, I don't know what value should be overridden.
> 
> Do you have any ideas in this regard?

I think the simplest solution would be to just change the structs that derive `clap::Parser` so that all their values are `Option`s, and remove any default values from the struct. Then, we would merge configurations by using any values that are `Some` in the command line arguments, and falling back to the config file for anything that is `None` in the command-line arguments.

I don't think this will require the use of the `clap` builder API.

nrskt:
> I think the simplest solution would be to just change the structs that derive clap::Parser so that all their values are Options, and remove any default values from the struct.

That sounds great! I'll try your suggestion.

nrskt:
@hawkw Hi, I create a new pull request about this PR. please review it.

https://github.com/tokio-rs/console/pull/320

PR #313, opened 2022-03-22
docs: add tokio-console installation section

nickgerace:
## Description

This PR adds a section for installing the tokio-console crate as well as fixes some broken console links. All changes were made to the README. Even though the tokio-console crate has its own README that includes installation, users will likely stumble upon the root README first. Thus, the purpose of this PR is to aid folks who may not wish to clone the repository when trying out the project.

## Commit Notes
- Add tokio-console installation section in the README as the primary
  method of using tokio-console
- Move the local cargo run option to its own alternative method section
- Update broken console links to use tokio-console in the README
- Add ".idea" to gitignore
- Update cargo install to use "--locked" flag in the nested README

## Additional Notes

Excellent project! Thank you for this.


README.md
```diff
@@ -132,6 +132,18 @@ $ cargo run
 ```
 in this repository.
 
+> **alternative method:** install the tool locally
+>
+> ```shell
+> $ cargo install --locked tokio-console
```
nickgerace: I see that the tokio-console README does not have the `--locked` flag for its installation instructions. I'd be happy to remove this, but thought I'd keep it for now in case reviewers wanted to go one way or another.

hawkw: If using `--locked` is generally considered preferable, this is fine with me --- we may want to update the `tokio-console` crate's README as well. I don't have a strong opinion here.

README.md
```diff
@@ -132,6 +132,18 @@ $ cargo run
 ```
 in this repository.
 
+> **alternative method:** install the tool locally
```
hawkw: nit, take it or leave it: how about specifically noting 
```suggestion
> **alternative method:** install `tokio-console` from crates.io:
```

hawkw: also, i might consider changing the "running the console" README section to present installing the CLI from crates.io _first_, as the primary suggested way to run the console, and then suggest running from a local checkout of the repository as an alternative approach for console development.

this README was written while the project was in early development and hadn't been published to crates.io yet, so we suggested running from a local git checkout as it was the only approach available at the time. now that the console is published, it would make sense to change this.

hawkw (commented):
thanks for the PR, this is definitely a good idea! i had a few small suggestions.

README.md
nickgerace: Looking at the [cargo install docs](https://doc.rust-lang.org/cargo/commands/cargo-install.html)...

> By default, the `Cargo.lock` file that is included with the package will be ignored. This means that Cargo will recompute which versions of dependencies to use, possibly using newer versions that have been released since the package was published. The `--locked` flag can be used to force Cargo to use the packaged `Cargo.lock` file if it is available. This may be useful for ensuring reproducible builds, to use the exact same set of dependencies that were available when the package was published. It may also be useful if a newer version of a dependency is published that no longer builds on your system, or has other problems. The downside to using `--locked` is that you will not receive any fixes or updates to any dependency. 

I think it might be worth including the `--locked` flag to ensure the user experience matches what's expected. However, I see the benefit of getting security / performance / bug fixes. I could go either way too, but lean towards `--locked`. What are your thoughts?

README.md
nickgerace: Good points! Perhaps, it might even be a good idea to move the suggestion to run from a local checkout of the repository to a development/contributing file/section. I'll take a look.

nickgerace:
@hawkw thanks for your review! I made changes based on the above discussions and would be happy to change more, as needed/desired.

.gitignore
```diff
@@ -2,3 +2,4 @@
 **/target
 **/result
 **/*.swp
+**/.idea
```
nickgerace, line 5: This is new since the first review. I switched back to CLion and IntelliJ for Rust dev 😅. I could just drop this part from the commit, but it could be useful for others.

README.md
```diff
@@ -126,11 +126,23 @@ notes:
...
-$ cargo run
+$ tokio-console
 ```
-in this repository.
+
+> **alternative method:** run the tool from a local checkout of this repository
+>
+> ```shell
+> $ cargo run
+> ```
```
nickgerace, line 148: Swapped the alternative and primary section contents based on previous discussion: https://github.com/tokio-rs/console/pull/313#discussion_r835945170

tokio-console/README.md
```diff
@@ -72,7 +72,7 @@ use the [Tokio] runtime, this means that:
 Once the application is instrumented, install the console CLI using
 
 ```shell
-cargo install tokio-console
+cargo install --locked tokio-console
```
nickgerace, line 75: Chose the `--locked` side to land on, but I do not feel _that_ strongly about it

(Original thread: https://github.com/tokio-rs/console/pull/313#discussion_r839926585)

.gitignore
hawkw: i'm fine with having editor-specific gitignore configurations, i think, but i do have a preference for merging this kind of change separately, so that it's not part of the git history for the unrelated documentation changes in this PR?

README.md
```diff
@@ -126,11 +126,23 @@ notes:
...
+
+and run locally
+
 ```shell
-$ cargo run
+$ tokio-console
 ```
-in this repository.
+
+> **alternative method:** run the tool from a local checkout of this repository
```
hawkw, line 144: maybe worth noting here that this is primarily intended for developers, or for using currently-unreleased changes? not a blocker.

tokio-console/README.md
hawkw, line 75: we should _probably_ make the use of `$` in shell examples consistent between the two READMEs --- I notice that the repo root readme always uses a `$` to indicate shell commands, while the `tokio-console` crate README doesn't. 

i have a _slight_ preference for omitting the `$` to make it easier to copy and paste the command into your shell, but it's not a strong preference.

in any case, we can address this in a follow-up change.

hawkw (approved):
this looks good to me! i had a couple more suggestions, but they're not blockers.

nickgerace:
Whoops! Apologies @hawkw, I had meant to make those changes but did not have the time recently. Feel free to do whatever in a follow-up commit or I can open another PR when I get the chance. Thanks for the reviews :)

PR #314, opened 2022-03-24
Don't trace tasks spawned through the console server

g2p:
The console subscriber's server generates a lot of async activity,
introspecting or logging that drowns out the activity the program
has outside of instrumentation.

Set a thread-local subscriber that drops events coming from the server thread.

Implementation requires a simple wrapper for NoSubscriber, because the dispatch implementation uses it as a sentinel value (https://github.com/tokio-rs/tracing/pull/2001 explains the history) and logs through the global dispatcher when a thread-local NoSubscriber is set.

### DIFF
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -39,6 +39,9 @@ pub struct Builder {
 
     /// The filter environment variable to use for `tracing` events.
     pub(super) filter_env_var: String,
+
+    /// Whether to trace events coming from the subscriber thread
+    self_trace: bool,
 }
 
 impl Default for Builder {
@@ -51,6 +54,7 @@ impl Default for Builder {
             server_addr: SocketAddr::new(Server::DEFAULT_IP, Server::DEFAULT_PORT),
             recording_path: None,
             filter_env_var: "RUST_LOG".to_string(),
+            self_trace: false,
         }
     }
 }
@@ -174,6 +178,13 @@ impl Builder {
             filter_env_var: filter_env_var.into(),
             ..self
         }
+    }
+
+    /// Sets whether we are are tracing events coming from the console subscriber
+    ///
+    /// The default is to drop events coming from the console subscriber thread.
+    pub fn enable_self_trace(self, self_trace: bool) -> Self {
+        Self { self_trace, ..self }
     }
 
     /// Completes the builder, returning a [`ConsoleLayer`] and [`Server`] task.
@@ -368,6 +379,8 @@ impl Builder {
             meta.name().starts_with("runtime.") || meta.target().starts_with("tokio")
         }
 
+        let self_trace = self.self_trace;
+
         let (layer, server) = self.build();
         let filter =
             FilterFn::new(console_filter as for<'r, 's> fn(&'r tracing::Metadata<'s>) -> bool);
@@ -376,6 +389,12 @@ impl Builder {
         thread::Builder::new()
             .name("console_subscriber".into())
             .spawn(move || {
+                let _subscriber_guard;
+                if !self_trace {
+                    _subscriber_guard = tracing::subscriber::set_default(
+                        tracing_core::subscriber::NoSubscriber::default(),
+                    );
+                }
                 let runtime = runtime::Builder::new_current_thread()
                     .enable_io()
                     .enable_time()
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -180,9 +180,11 @@ impl Builder {
         }
     }
 
-    /// Sets whether we are are tracing events coming from the console subscriber
+    /// Sets whether tasks, resources, and async ops from the console
+    /// subscriber thread are recorded.
     ///
-    /// The default is to drop events coming from the console subscriber thread.
+    /// By default, events from the console subscriber are discarded and
+    /// not exported to clients.
     pub fn enable_self_trace(self, self_trace: bool) -> Self {
         Self { self_trace, ..self }
     }
--- a/console-subscriber/Cargo.toml
+++ b/console-subscriber/Cargo.toml
@@ -36,7 +36,7 @@ tokio-stream = "0.1"
 thread_local = "1.1.3"
 console-api = { version = "0.2.0", path = "../console-api", features = ["transport"] }
 tonic = { version = "0.7", features = ["transport"] }
-tracing-core = "0.1.18"
+tracing-core = "0.1.24"
 tracing = "0.1.26"
 tracing-subscriber = { version = "0.3.0", default-features = false, features = ["fmt", "registry"] }
 futures = { version = "0.3", default-features = false }


g2p:
(Now that https://github.com/tokio-rs/tracing/pull/2001 is in a release, console could also require that release instead of the NoSubscriber wrapper)

seanmonstar:
I kinda think the task being shown _is_ useful, since it could help you realize how much the console is affecting your app, or reveal potential bugs in the console server. But I understand wanting to hide it usually. I don't know what the best solution is. Maybe we can dim it, or have a thing at the bottom "2 tasks hidden by default", or some other idea?

g2p:
I could provide a way to make this configurable, I would put it in `console_subscriber::Builder`, but the default would be to hide so that a console user gets something relevant with less effort and knowledge.  A runtime annotation that could be filtered client-side is also much harder than the thread-local filter in this PR, the annotations don't exist right now.

Aside: Having hyper tasks coming from the subscriber definitely made the console unusable for me at release.  Currently tasks like this cannot be distinguished if the app also uses hyper, though better context annotations will hopefully be provided at some point.

Figuring out console bugs or overheads is more relevant to a tokio-console developer, someone also more able to find and configure an "enable_self_trace" flag.

hawkw:
I agree that it feels a bit philosophically wrong to totally exclude the console's tasks, because then we are not reporting a completely accurate view of the application's real state. But, pragmatically, if the console tasks make using the console more difficult, I'm willing to consider excluding them --- doing so may also improve performance and/or reduce dropped events. So, I agree with @seanmonstar --- we should allow the user to choose whether the console's tasks are recorded or not.

I think in an _ideal_ world, we would want to configure whether or not the console's tasks are recorded in the console UI, using a general-purpose runtime filtering system as described in #131. However, implementing that is a pretty big project --- it would require UI work as well as potential changes to the wire protocol and `console-subscriber` crates. So, if the console tasks are currently causing problems for users trying to debug their applications, I think a builder option for enabling recording the console's tasks would be a fine stop-gap solution until the more general-purpose filtering system is implemented.

console-subscriber/src/builder.rs
```diff
@@ -14,6 +14,54 @@ use tracing_subscriber::{
...
+    }
+
+    fn enter(&self, span: &tracing_core::span::Id) {
+        self.0.enter(span)
+    }
+
+    fn exit(&self, span: &tracing_core::span::Id) {
+        self.0.exit(span)
+    }
+}
```
hawkw: this issue was fixed upstream in [`tracing-core` 0.1.24`](https://github.com/tokio-rs/tracing/releases/tag/tracing-core-0.1.24), so if we change the `console-subscriber` crate's minumum `tracing-core` dependency to 0.1.24, we can get rid of this workaround.

console-subscriber/src/builder.rs
```diff
@@ -376,17 +424,19 @@ impl Builder {
...
+                        .enable_time()
+                        .build()
+                        .expect("console subscriber runtime initialization failed");
+
+                    runtime.block_on(async move {
+                        server
+                            .serve()
+                            .await
+                            .expect("console subscriber server failed")
+                    });
```
hawkw: nit, take it or leave it: we could avoid adding an additional level of indentation if we used [`tracing::subscriber::set_default`](https://docs.rs/tracing/latest/tracing/subscriber/fn.set_default.html), which returns a `Drop` guard, rather than `with_default` here:
```suggestion
                let _subscriber = tracing::subscriber::set_default(NoNoSubscriber::default());
                let runtime = runtime::Builder::new_current_thread()
                    .enable_io()
                    .enable_time()
                    .build()
                    .expect("console subscriber runtime initialization failed");

                runtime.block_on(async move {
                    server
                        .serve()
                        .await
                        .expect("console subscriber server failed")
```

I have a slight preference for doing it this way, because it makes the Git diff for this change more focused (currently, the indentation change makes all the code inside the `with_default` closure appear to be "added" in this PR). But, it's not a big deal, and I won't block merging on that.

hawkw (changes requested):
I think that, rather than _always_ disabling recording on the `console-subscriber` thread, we should add a builder option to control whether or not it's disabled, and only set `NoSubscriber` when that builder option is set. I would be fine with making the `console-subscriber` thread disabled by default, but I think it ought to be _possible_ to enable recording it.

Also, I commented with a couple smaller suggestions, but neither of those are hard blockers for merging this branch.

console-subscriber/src/builder.rs
g2p: Git has a flag for that (-b and -w), GitHub doesn't.
Changed, the guard works better when setting the subscriber is conditional.

console-subscriber/src/builder.rs
g2p: I've removed the workaround

g2p:
Addressed comments and made this configurable.

console-subscriber/src/builder.rs
```diff
@@ -176,6 +180,13 @@ impl Builder {
         }
     }
 
+    /// Sets whether we are are tracing events coming from the console subscriber
+    ///
+    /// The default is to drop events coming from the console subscriber thread.
```
hawkw: nit: how about:
```suggestion
    /// Sets whether tasks, resources, and async ops from the console 
    /// subscriber thread are recorded. 
    ///
    /// By default, events from the console subscriber are discarded and
    /// not exported to clients.
```

console-subscriber/src/builder.rs
```diff
@@ -376,6 +389,12 @@ impl Builder {
         thread::Builder::new()
             .name("console_subscriber".into())
             .spawn(move || {
+                let _subscriber_guard;
+                if !self_trace {
+                    _subscriber_guard = tracing::subscriber::set_default(
+                        tracing_core::subscriber::NoSubscriber::default(),
+                    );
+                }
```
hawkw, line 399: nit, take it or leave it: this approach seems fine to me. a slightly more concise alternative could be to use [`bool::then`](https://doc.rust-lang.org/stable/std/primitive.bool.html#method.then) to create an `Option<DefaultGuard>`:
```suggestion
                let _subscriber_guard = self_trace.then(|| {
                    tracing::subscriber::set_default(tracing_core::subscriber::NoSubscriber::default())
                });
```
but the current implementation might be clearer, so feel free to take it or leave it!

hawkw (approved):
looks good to me! thank you!

hawkw:
~~@g2p mind merging the `main` branch into this PR so we can get a passing CI build? I'll be happy to merge it afterwards --- it looks like I can't update this branch from `main` in the GitHub UI, for some reason.~~

**Edit**: never mind, I took care of it.

console-subscriber/src/builder.rs
```diff
@@ -176,6 +180,15 @@ impl Builder {
         }
     }
 
+    /// Sets whether tasks, resources, and async ops from the console 
+    /// subscriber thread are recorded.
```
hawkw: gah, rustfmt is mad that i accidentally added trailing whitespace here
```suggestion
    /// subscriber thread are recorded.
```

console-subscriber/src/builder.rs
g2p: Interesting! Maybe not clear enough as the test is negated.

g2p:
(I just rebased to appease CI.  Feel free to merge when that works)

PR #316, opened 2022-03-31
feat(console): surface dropped events if there are any

bobrik:
During normal operation it won't show anything:

```
connection: http://127.0.0.1:6669/ (CONNECTED)
```

But if there are any drops detected, it will show a cumulative amount:

```
connection: http://127.0.0.1:6669/ (CONNECTED) dropped events: 2507
```

image


tokio-console/src/main.rs
```diff
@@ -127,6 +127,13 @@ async fn main() -> color_eyre::Result<()> {
                     .0
                     .push(Span::styled(" PAUSED", view.styles.fg(Color::Red)));
             }
+            let dropped_events = state.async_ops_state().dropped_events();
+            if dropped_events > 0 {
+                header_text.0.push(Span::styled(
+                    format!(" dropped events: {}", dropped_events),
+                    view.styles.fg(Color::Red),
+                ));
+            }
```
hawkw: It looks like this will only display the number of dropped async ops. But, we also have separate counters of [dropped tasks](https://docs.rs/console-api/latest/console_api/tasks/struct.TaskUpdate.html#structfield.dropped_events) and [dropped resources](https://docs.rs/console-api/latest/console_api/resources/struct.ResourceUpdate.html#structfield.dropped_events). If we only display the number of dropped async ops, we won't inform the user if any new tasks or resources were dropped. We should probably also include those in the UI.

hawkw (commented):
Thanks for working on this, this PR looks like a great start!

One thing I notice is that we're currently only showing the count of dropped async ops, and not dropped tasks or resources. We should probably fix that.

Also, some ideas for the future:
 - I had initially wondered if it might be nice to show a count of dropped events as a warning in the warnings UI box, but I think showing it here is fine for now.
 - I also wonder if we want to consider showing the number of dropped events over a time window, like `43 events dropped over the last 20 seconds (3489 dropped events total)` or something?

But, those can both be addressed in follow-ups --- for now, I think it's important to just actually start showing this data, as it's not currently exposed at all...

tokio-console/src/main.rs
bobrik: Should it be a one number with the sum of three or a separate counter for each one of them?

hawkw: I would probably have a separate counter for each.

hawkw:
@bobrik FYI, you generally don't need to force push new changes to your PRs. We merge pull requests by squashing, so individual work-in-progress commits to a branch are fine. See https://github.com/tokio-rs/console/blob/main/CONTRIBUTING.md#commit-squashing for details.

tokio-console/src/main.rs
```diff
@@ -127,6 +127,15 @@ async fn main() -> color_eyre::Result<()> {
...
             }
+            let dropped_updates = state.async_ops_state().dropped_events()
+                + state.tasks_state().dropped_events()
+                + state.resources_state().dropped_events();
+            if dropped_updates > 0 {
+                header_text.0.push(Span::styled(
+                    format!(" dropped updates: {}", dropped_updates),
+                    view.styles.fg(Color::Red),
+                ));
+            }
```
hawkw: I think it's probably better to show a separate count for each category of dropped events. We may be in a state where we have not dropped any tasks, but we have dropped several async ops and several resources (for example). Summing all the counters hides these details.

hawkw (changes requested):

hawkw:
@bobrik quick ping --- are you still planning to work on this? I'd like to get it merged! If you don't have the time to finish this, I'd be happy to wrap it up myself, just let me know!

bobrik:
Apologies for the delay. I pushed separate counters and a bugfix for overcounting of `resources_state` drops.

tokio-console/src/main.rs
```diff
@@ -127,6 +127,27 @@ async fn main() -> color_eyre::Result<()> {
...
+            if dropped_tasks_state > 0 {
+                header_text.0.push(Span::styled(
+                    format!(" dropped tasks_state: {}", dropped_tasks_state),
+                    view.styles.fg(Color::Red),
+                ));
+            }
+            let dropped_resources_state = state.resources_state().dropped_events();
+            if dropped_resources_state > 0 {
+                header_text.0.push(Span::styled(
+                    format!(" dropped resources_state: {}", dropped_resources_state),
```
hawkw: nit: i would probably change these to just say
```suggestion
            if dropped_async_ops_state > 0 {
                header_text.0.push(Span::styled(
                    format!(" dropped async ops: {}", dropped_async_ops_state),
                    view.styles.fg(Color::Red),
                ));
            }
            let dropped_tasks_state = state.tasks_state().dropped_events();
            if dropped_tasks_state > 0 {
                header_text.0.push(Span::styled(
                    format!(" dropped tasks: {}", dropped_tasks_state),
                    view.styles.fg(Color::Red),
                ));
            }
            let dropped_resources_state = state.resources_state().dropped_events();
            if dropped_resources_state > 0 {
                header_text.0.push(Span::styled(
                    format!(" dropped resources: {}", dropped_resources_state),
```

tokio-console/src/main.rs
```diff
@@ -127,6 +127,27 @@ async fn main() -> color_eyre::Result<()> {
...
+                    view.styles.fg(Color::Red),
+                ));
+            }
```
hawkw: take it or leave it: we could probably save a little horizontal real estate by formatting the text like
```
dropped XXXX tasks, YYYY resources, ZZZZ async ops
```
or something?

instead of repeating the word "dropped" three times, we could check if any counter is non-zero, print a single "dropped", and then print the actual counts. not a blocker, though, as the current code should fit on most terminals.

hawkw (approved):
I had a couple minor suggestions, but overall, this looks good to me. Thanks for working on it!

tokio-console/src/main.rs
bobrik: Good idea, updated.

hawkw (approved):

PR #318, opened 2022-04-02
feat(console-api): Update `tonic` to `0.7`

LucioFranco:
This change updates `tonic` and `prost` to their latest versions. In
addition this PR changes how the generated protobuf is built from being
done in a `build.rs` file to a `tests/boostrap.rs` test. This removes
the need for downstream consumers to compile/provide `protoc` for
`prost` to generate the protobuf files (which was changed in `0.10`).

The bootstrap test will first attempt to compile the protobuf files and
check if there is a git diff between the checked in version. If there is
a difference it will fail the test waiting for the files to be checked
in. This will fail PRs that have updated the protobuf files but not
checked in the new versions.

BREAKING CHANGE: 
`console-api` is now no longer compatible with projects using `prost`
0.9 or `tonic` 0.7. These crates must be updated to use `console-api`
0.2.


LucioFranco:
I tried adding some code to download protoc to speed up CI but didn't realize that deps were cached so we only pay the cost on the first build. Going to back out my changes.

.github/workflows/ci.yaml
```diff
@@ -25,11 +25,9 @@ env:
...
   # Don't emit giant backtraces in the CI logs.
   RUST_BACKTRACE: short
-  # Make Rust 1.56.0 the minimum supported Rust version.
-  #
-  # Some of our dependencies require `edition = "2021"` which is only supported
-  # on 1.56.0+.
-  minrust: 1.56.0
+  # Make 1.58 MSRV due to <=1.57 failing to build console-api due to a bug
+  # with cargo version resolution. 
+  minrust: 1.58.0
```
LucioFranco, line 30: Would be good to find the bug here 🤷

hawkw (commented):
I notice that the generated code is now checked in --- is this a recommended practice in Tonic 0.7-land?

LucioFranco:
So `prost` 0.10 requires at least a c compiler and to compile protobuf from source if `protoc` is not present. So for end users of the `console-api` checking in the code will avoid having to compile C code.

hawkw (approved):
this seems legit to me!

PR #320, opened 2022-04-05
feat(console): Support config file

nrskt:
Support config files as follows.

- $XDG_CONFIG_HOME/tokio-console/console.toml
- current configuration (`./console.toml`)

## Test

Execute and check debug log.

`env -u LANG -u COLORTERM RUST_LOG=debug cargo run`

### case1) no config files

```
DEBUG tokio_console: args.target_addr=http://127.0.0.1:6669/ args.view_options=
ViewOptions { 
    no_colors: None,
    lang: None,
    ascii_only: None,
    truecolor: None,
    palette: None,
    toggles: ColorToggles { 
        color_durations: None, 
        color_terminated: None
    }
}
```

### case2) current config file

```toml
[charset]
lang = "en_us.UTF-8"
ascii_only = true

[colors]
enabled = true
palette = "all"
truecolor = true

[colors.enable]
durations = true 
terminated = true
```

```
DEBUG tokio_console: args.target_addr=http://127.0.0.1:6669/ args.view_options=
ViewOptions { 
    no_colors: Some(false), 
    lang: Some("en_us.UTF-8"), 
    ascii_only: Some(true), 
    truecolor: Some(true), 
    palette: Some(All), 
    toggles: ColorToggles { 
        color_durations: Some(false), 
        color_terminated: Some(false) 
    } 
}
```

### case3) XDG config file

```toml
[charset]
lang = "en_us.UTF-8"
ascii_only = true

[colors]
enabled = true
palette = "all"
truecolor = true

[colors.enable]
durations = true 
terminated = true
```

```
DEBUG tokio_console: args.target_addr=http://127.0.0.1:6669/ args.view_options=
ViewOptions { 
    no_colors: Some(false), 
    lang: Some("en_us.UTF-8"), 
    ascii_only: Some(true), 
    truecolor: Some(true), 
    palette: Some(All), 
    toggles: ColorToggles { 
        color_durations: Some(false), 
        color_terminated: Some(false) 
    } 
}
```

### case4) XDG config file and current config file

XDG_CONFIG/console.toml

```toml
[charset]
lang = "en_us.UTF-8"
ascii_only = true

[colors]
enabled = true
palette = "all"
truecolor = true

[colors.enable]
durations = true 
terminated = true
```

./console.toml

```
[charset]
lang = "en_us.UTF-8"
ascii_only = false
```

Override the current config file.

```
DEBUG tokio_console: args.target_addr=http://127.0.0.1:6669/ args.view_options=
ViewOptions { 
    no_colors: Some(false), 
    lang: Some("en_us.UTF-8"), 
    ascii_only: Some(false),     <-- override 
    truecolor: Some(true), 
    palette: Some(All), 
    toggles: ColorToggles { 
    color_durations: Some(false), 
    color_terminated: Some(false) 
    } 
}
```

### case5) XDG config file, current config file, and command-line option

XDG_CONFIG/console.toml

```toml
[charset]
lang = "en_us.UTF-8"
ascii_only = true

[colors]
enabled = true
palette = "all"
truecolor = true

[colors.enable]
durations = true 
terminated = true
```

./console.toml

```
[charset]
lang = "en_us.UTF-8"
ascii_only = false
```

`env -u LANG -u COLORTERM RUST_LOG=debug cargo run -- --palette 8`

Override the current config file and command-line option.

```
DEBUG tokio_console: args.target_addr=http://127.0.0.1:6669/ args.view_options=
ViewOptions { 
    no_colors: Some(false), 
    lang: Some("en_us.UTF-8"), 
    ascii_only: Some(false),     <-- override
    truecolor: Some(true), 
    palette: Some(Ansi8),        <-- override
    toggles: ColorToggles { 
        color_durations: Some(false), 
        color_terminated: Some(false) 
    } 
}
```

Relates to 
- https://github.com/tokio-rs/console/issues/310
- https://github.com/tokio-rs/console/pull/312
    - https://github.com/tokio-rs/console/pull/312#issuecomment-1075421420


tokio-console/src/config.rs
```diff
@@ -163,11 +206,43 @@ impl Config {
     }
 }
 
+fn merge_view_options(base: &mut ViewOptions, command_line: ViewOptions) {
```
hawkw: nit. take it or leave it: IMO, this seems like it would make more sense as a method on `ViewOptions`:
```rust
impl ViewOptions {
    fn merge_with(&mut self, command_line: ViewOptions) {
        // ...
    }
}
```
this way, it's clearer which is the base and which is the override.

tokio-console/src/config.rs
```diff
@@ -163,11 +206,43 @@ impl Config {
...
+        base.toggles.color_terminated = command_line.toggles.color_terminated;
+    }
+}
+
 // === impl ViewOptions ===
 
 impl ViewOptions {
     pub fn is_utf8(&self) -> bool {
-        self.lang.ends_with("UTF-8") && !self.ascii_only
+        let lang = self.lang.clone().unwrap_or_default();
```
hawkw: i don't think we need to clone `self.lang` here; we can borrow it instead:
```suggestion
        let lang = self.lang.as_ref().unwrap_or_default();
```

tokio-console/src/config.rs
```diff
@@ -110,17 +113,57 @@ pub struct ViewOptions {
...
+    truecolor: bool,
+    palette: Palette,
+    enable: ColorsEnable,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct ColorsEnable {
+    durations: bool,
+    terminated: bool,
 }
```
hawkw: this struct seems very similar to `ColorToggles`, is it really necessary to have separate `ColorToggles` and `ColorsEnable` structs? can we just derive `Deserialize` for `ColorToggles`, and continue inverting the command-line arguments as part of the `parse` function when parsing CLI args?

tokio-console/src/config.rs
```diff
@@ -110,17 +113,57 @@ pub struct ViewOptions {
...
+    ascii_only: bool,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct ColorsConfig {
+    enabled: bool,
+    truecolor: bool,
+    palette: Palette,
+    enable: ColorsEnable,
+}
```
hawkw: It looks like the the `charset` and `colors` fields are `Option`s, but the fields _on_ those types are all not optional. So, if I understand correctly, it looks like we can successfully parse a config file that does not include a `charset` section or a `colors` section, but if those sections are present, _all_ their fields *must* be included in the config file.

For example, I believe a config file that contains only

```rust
[colors]
enable = true; 
```
will result in a parse error, because all the fields on `ColorsConfig` are required.

We should probably change this so that we can accept config files that don't set _all_ the config file keys. So, we should probably make all the fields on these types `Option`s.

tokio-console/src/config.rs
```diff
@@ -163,11 +206,43 @@ impl Config {
     }
 }
 
+fn merge_view_options(base: &mut ViewOptions, command_line: ViewOptions) {
+    if command_line.no_colors.is_some() {
+        base.no_colors = command_line.no_colors;
+    }
```
hawkw: nit, take it or leave it: i think we could potentially simplify these using [`Option::or`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.or):
```suggestion
   base.no_colors = command_line.no_colors.or(base.no_colors.take());
```

tokio-console/src/config.rs
```diff
@@ -238,3 +313,67 @@ impl FromStr for RetainFor {
...
+
+    pub fn color_terminated(&self) -> bool {
+        self.color_durations.map(std::ops::Not::not).unwrap_or(true)
+    }
+}
+
+// === impl ColorToggles ===
+
+impl ConfigFile {
+    fn from_config() -> Option<Self> {
```
hawkw: can we change this to return a `Result<Option<Self>, ...>`? I think that if parsing the either config file fails, we should print the error to the user. of course, if there just weren't any config files present, we should still return an `Option`.

tokio-console/src/config.rs
```diff
@@ -238,3 +313,67 @@ impl FromStr for RetainFor {
...
+        let mut base = dirs::config_dir();
+        if let Some(path) = base.as_mut() {
+            path.push("tokio-console/console.toml");
+        }
+        let base = base.and_then(|path| fs::read_to_string(path).ok());
+        let base_file: Option<ConfigFile> = base.and_then(|raw| toml::from_str(&raw).ok());
```
hawkw: i think we could simplify this a bit by moving more of this code into the `if let`. here's my suggestion (assuming that we change this method to return a `Result`, as I suggested)
```suggestion
        let base_file = if let Some(path) = base.as_mut() {
            path.push("tokio-console/console.toml");
            fs::read_to_string(path).ok()
                .map(|raw| toml::from_str::<ConfigFile>(&raw))
                .transpose()?
        } else {
            None
        }
```

tokio-console/src/config.rs
```diff
@@ -238,3 +313,67 @@ impl FromStr for RetainFor {
...
+                    .colors
+                    .as_ref()
+                    .map(|config| Not::not(config.enable.durations)),
+                color_terminated: self.colors.map(|config| Not::not(config.enable.terminated)),
+            },
+        }
+    }
+}
+
+fn merge_config_file(before: Option<ConfigFile>, after: Option<ConfigFile>) -> Option<ConfigFile> {
```
hawkw: nit, take it or leave it: again, this seems like it could be better off as a method on `ConfigFile`?

hawkw (changes requested):
thanks for looking on this, this looks like a good start! i had some suggestions.

tokio-console/src/config.rs
nrskt: I understood your suggestion. I thought the config file does not accept partial settings.
I'll fix it.

tokio-console/src/config.rs
nrskt: Ok, I'll try it. thank you!

tokio-console/src/config.rs
nrskt: Ok, Certainly it is easy for users to understand.
I'll fix it.

tokio-console/src/config.rs
nrskt: > this struct seems very similar to ColorToggles, is it really necessary to have separate ColorToggles and ColorsEnable structs?

No, it does not necessary. I think it can be fixed as you suggested.
I made a mistake that `durations` and `terminated` are required in the config file.

> can we just derive Deserialize for ColorToggles, and continue inverting the command-line arguments as part of the parse function when parsing CLI args?

I think to invert the bool value with the getter method. 

```
// === impl ColorToggles ===

impl ColorToggles {
    pub fn color_durations(&self) -> bool {
        self.color_durations.map(std::ops::Not::not).unwrap_or(true)
    }

    pub fn color_terminated(&self) -> bool {
        self.color_durations.map(std::ops::Not::not).unwrap_or(true)
    }
}
```
---

**Why I remove `parse(from_flag = std::ops::Not::not)` attribute.**

First, I tried this approach.

I changed the fields (color_durations, color_terminated) to optional.
But, this approach does not pass the compile. because `from_flag` required `fn(bool) -> T`.

```rust
#[derive(Clap, Debug, Copy, Clone)]
pub struct ColorToggles {
    /// Disable color-coding for duration units.
    #[clap(long = "no-duration-colors", parse(from_flag = std::ops::Not::not), group = "colors")]
    pub(crate) color_durations: Option<bool>,

    /// Disable color-coding for terminated tasks.
    #[clap(long = "no-terminated-colors", parse(from_flag = std::ops::Not::not), group = "colors")]
    pub(crate) color_terminated: Option<bool>,
}
```

So I defined the getter method.

nrskt:
@hawkw 

I fixed the implementations and try to test again.
I have questions about case7 and case8. please check it.

## test
### case1) no config files

```
DEBUG tokio_console: args.target_addr=http://127.0.0.1:6669/ args.view_options=
ViewOptions { 
    no_colors: None, 
    lang: None, 
    ascii_only: None, 
    truecolor: None, 
    palette: None, 
    toggles: ColorToggles { 
        color_durations: None, 
        color_terminated: None 
    } 
}
```

### case2) current config file

```toml
[charset]
lang = "en_us.UTF-8"
ascii_only = true

[colors]
enabled = true
palette = "all"
truecolor = true

[colors.enable]
durations = true 
terminated = true
```

```
DEBUG tokio_console: args.target_addr=http://127.0.0.1:6669/ args.view_options=
ViewOptions { 
    no_colors: Some(false), 
    lang: Some("en_us.UTF-8"), 
    ascii_only: Some(true), 
    truecolor: Some(true),
    palette: Some(All), 
    toggles: ColorToggles { 
        color_durations: Some(false), 
        color_terminated: Some(false) 
    } 
}
```

### case3) XDG config file

```toml
[charset]
lang = "en_us.UTF-8"
ascii_only = true

[colors]
enabled = true
palette = "all"
truecolor = true

[colors.enable]
durations = true 
terminated = true
```

```
DEBUG tokio_console: args.target_addr=http://127.0.0.1:6669/ args.view_options=
ViewOptions { 
    no_colors: Some(false), 
    lang: Some("en_us.UTF-8"), 
    ascii_only: Some(true), 
    truecolor: Some(true), 
    palette: Some(All), 
    toggles: ColorToggles { 
        color_durations: Some(false), 
        color_terminated: Some(false) 
    }
}
```

### case4) XDG config file and current config file

XDG_CONFIG/console.toml

```toml
[charset]
lang = "en_us.UTF-8"
ascii_only = true

[colors]
enabled = true
palette = "all"
truecolor = true

[colors.enable]
durations = true 
terminated = true
```

./console.toml

```
[charset]
lang = "en_us.UTF-8"
ascii_only = false
```

Override the current config file.

```
DEBUG tokio_console: args.target_addr=http://127.0.0.1:6669/ args.view_options=
ViewOptions { 
    no_colors: Some(false), 
    lang: Some("en_us.UTF-8"), 
    ascii_only: Some(false),    <-- override 
    truecolor: Some(true), 
    palette: Some(All), 
    toggles: ColorToggles { 
        color_durations: Some(false), 
        color_terminated: Some(false) 
    } 
}
```

### case5) XDG config file, current config file, and command-line option

XDG_CONFIG/console.toml

```toml
[charset]
lang = "en_us.UTF-8"
ascii_only = true

[colors]
enabled = true
palette = "all"
truecolor = true

[colors.enable]
durations = true 
terminated = true
```

./console.toml

```
[charset]
lang = "en_us.UTF-8"
ascii_only = false
```

`env -u LANG -u COLORTERM RUST_LOG=debug cargo run -- --palette 8`

Override the current config file and command-line option.

```
DEBUG tokio_console: args.target_addr=http://127.0.0.1:6669/ args.view_options=
ViewOptions { 
    no_colors: Some(false), 
    lang: Some("en_us.UTF-8"), 
    ascii_only: Some(false),    <-- override 
    truecolor: Some(true), 
    palette: Some(Ansi8),       <-- override
    toggles: ColorToggles { 
        color_durations: Some(false), 
        color_terminated: Some(false) 
    } 
}
```

## Added

### case6) XDG config file, current config file

XDG_CONFIG/console.toml

```toml
[colors]
palette = "8"
```

./console.toml

```
[charset]
lang = "C"
```

```
DEBUG tokio_console: args.target_addr=http://127.0.0.1:6669/ args.view_options=
ViewOptions { 
    no_colors: None, 
    lang: Some("C"), 
    ascii_only: None, 
    truecolor: None, 
    palette: Some(Ansi8), 
    toggles: ColorToggles { 
        color_durations: None, 
        color_terminated: None 
    }
}
```

### case7) Parse error

[Question] I want you to tell me how to output a better error message.

XDG_CONFIG/console.toml

```toml
[colors]
palette = "8"
```

./console.toml

```
[charset]
ascii_only = "string value"
```

```
❯ env -u LANG -u COLORTERM RUST_LOG=debug cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/tokio-console`
failed to parse config file: Error {
    msg: "failed to parse Some(\"./console.toml\")",
    source: Error {
        inner: ErrorInner {
            kind: Custom,
            line: Some(
                1,
            ),
            col: 13,
            at: Some(
                23,
            ),
            message: "invalid type: string \"string value\", expected a boolean",
            key: [
                "charset",
                "ascii_only",
            ],
        },
    },
}
```

### case8) Ignore unknown config field

[Question] Did you expect this behavior?
I don't know how to check strictly.

XDG_CONFIG/console.toml

```toml
[colors]
palette = "8"
```

./console.toml

```
[charset]
ascii_only = false
wrong_key = false
```

```
DEBUG tokio_console: args.target_addr=http://127.0.0.1:6669/ args.view_options=
ViewOptions { 
    no_colors: None, 
    lang: None, 
    ascii_only: Some(false), 
    truecolor: None, 
    palette: Some(Ansi8), 
    toggles: ColorToggles { 
        color_durations: None, 
        color_terminated: None
    }
}
```

tokio-console/src/config.rs
hawkw: Okay, that seems fine to me, then!

tokio-console/src/config.rs
```diff
@@ -238,3 +312,86 @@ impl FromStr for RetainFor {
...
+// === impl ColorToggles ===
+
+impl ConfigFile {
+    fn from_config(path: ConfigPath) -> color_eyre::Result<Option<Self>> {
+        let config = path
+            .into_path()
+            .and_then(|path| fs::read_to_string(path).ok())
+            .map(|raw| toml::from_str::<ConfigFile>(&raw))
+            .transpose()
+            .wrap_err_with(|| format!("failed to parse {:?}", path.into_path()))?;
```
hawkw: A small note: I think the path will be formatted nicer if we use [`Path::display`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.display) here instead of its `Debug` implementation:
```suggestion
            .wrap_err_with(|| format!("failed to parse {}", path.into_path().display()))?;
```

tokio-console/src/main.rs
```diff
@@ -26,7 +25,13 @@ mod warnings;
...
 #[tokio::main]
 async fn main() -> color_eyre::Result<()> {
-    let mut args = config::Config::parse();
+    let mut args = match config::Config::from_config() {
+        Err(e) => {
+            eprintln!("failed to parse config file: {:#?}", e);
+            std::process::exit(1);
+        }
+        Ok(args) => args,
+    };
```
hawkw: I think that any config parsing error will be displayed more nicely if we format the error with its standard `Debug` (`{:?}`) formatter rather than its alternate-mode `Debug` formatter (`{:#?}`), [according to the `eyre` documentation](https://docs.rs/eyre/0.6.8/eyre/struct.Report.html#display-representations). So, if we change this to 
```suggestion
    let mut args = match config::Config::from_config() {
        Err(e) => {
            eprintln!("failed to parse config file: {:?}", e);
            std::process::exit(1);
        }
        Ok(args) => args,
    };
```
we'll get a much nicer message.

The documentation also notes that
> Note that this is the representation you get by default if you return an error from `fn main` instead of printing it explicitly yourself.

So, we could also just write this and get the same output, which I think is much nicer:
```suggestion
    let mut args = config::Config::from_config()?;
```

When I make either of those changes and run the console with the invalid configuration file from your example, I get output like this, which I think is more helpful than the `:#?` output:
```
:; cargo run
   Compiling tokio-console v0.1.3 (/home/eliza/Code/console/tokio-console)
    Finished dev [unoptimized + debuginfo] target(s) in 5.66s
     Running `target/debug/tokio-console`
Error: failed to parse Some("./console.toml")

Caused by:
    invalid type: string "string value", expected a boolean for key `charset.ascii_only` at line 2 column 14

Location:
    tokio-console/src/config.rs:337:14
```

(unfortunately, the output isn't colored, because we don't install our custom `eyre` handler until _after_ we've parsed the config file...but there's not really a way around that, so this is good enough!)

tokio-console/src/config.rs
```diff
@@ -238,3 +312,86 @@ impl FromStr for RetainFor {
...
+    fn color_terminated(&self) -> Option<bool> {
+        self.colors
+            .as_ref()
+            .and_then(|config| config.enable.map(|toggles| toggles.color_terminated()))
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+enum ConfigPath {
+    Xdg,
```
hawkw: This is a minor nitpick, but I don't think `Xdg` is the best thing to call this variant. The `dirs` trait uses `XDG_CONFIG_HOME` on Linux, but it uses a different directory on Windows or macOS. Calling it `Xdg` kind of suggests to the reader that it is specifically using `XDG_CONFIG_HOME`...maybe we should call it something like `User` or `Home` or something

tokio-console/src/config.rs
```diff
@@ -238,3 +312,86 @@ impl FromStr for RetainFor {
...
+
+// === impl ColorToggles ===
+
+impl ColorToggles {
+    pub fn color_durations(&self) -> bool {
+        self.color_durations.map(std::ops::Not::not).unwrap_or(true)
+    }
+
+    pub fn color_terminated(&self) -> bool {
+        self.color_durations.map(std::ops::Not::not).unwrap_or(true)
```
hawkw: minor style nit: `Not` is imported, so we could just write
```suggestion
        self.color_durations.map(Not::not).unwrap_or(true)
    }

    pub fn color_terminated(&self) -> bool {
        self.color_durations.map(Not::not).unwrap_or(true)
```

hawkw: it might also be worth having a comment or something here explaining why these are inverted?

tokio-console/src/config.rs
```diff
@@ -219,6 +272,27 @@ impl ViewOptions {
...
+        self.palette = command_line.palette.or_else(|| self.palette.take());
+        self.toggles.color_durations = command_line
+            .toggles
+            .color_durations
+            .or_else(|| self.toggles.color_durations.take());
+        self.toggles.color_terminated = command_line
+            .toggles
+            .color_terminated
+            .or_else(|| self.toggles.color_terminated.take());
+    }
```
hawkw: a minor style suggestion, take it or leave it: what if, instead of mutating `self`, we had this function take `self` by value and return a _new_ `ViewOptiions`? like this:
```suggestion
    fn merge_with(self, command_line: ViewOptions) -> ViewOptions {
        Self {
            no_colors: command_line.no_colors.or(self.no_colors),
            lang: command_line.lang.or(self.lang),
            ascii_only: command_line.ascii_only.or(self.ascii_only),
            truecolor: command_line.truecolor.or(self.truecolor),
            palette: command_line.palette.or(self.palette),
            toggles: ColorToggles {
                color_durations: command_line
                    .toggles
                    .color_durations
                    .or(self.toggles.color_durations)
                color_terminated: command_line
                    .toggles
                    .color_terminated
                    .or(self.toggles.color_terminated)
            }
        }
    }
```

This approach has one big advantage, in my opinion: if we add new fields to the `ViewOptions` type in the future, but forget to add code in `merge_with` to merge those fields, the current implementation will just use the values from the config file and ignore the values from the command-line, because it's modifying the _existing_ instance of `ViewOptions`. If we make the change I suggested, though, forgetting to merge any new fields would result in a compile-time error, because we are constructing a _new_ instance of `ViewOptions`, and we wouldn't be adding the new fields to the returned value. It would be nice to be able to prevent that kind of bug at compile-time, IMO.

hawkw: (we probably want to do the same thing in the `Config` struct's `merge_with` function, too...)

tokio-console/src/config.rs
```diff
@@ -107,20 +112,66 @@ pub struct ViewOptions {
...
+    lang: Option<String>,
+    ascii_only: Option<bool>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct ColorsConfig {
+    enabled: Option<bool>,
+    truecolor: Option<bool>,
+    palette: Option<Palette>,
+    enable: Option<ColorToggles>,
```
hawkw: nit: these are only used for parsing the TOML config file, they're not exposed to the rest of the crate, as we turn them into a `Config` before returning it. i think these don't need to be `pub`:
```suggestion
#[derive(Debug, Clone, Deserialize)]
struct ConfigFile {
    charset: Option<CharsetConfig>,
    colors: Option<ColorsConfig>,
}

#[derive(Debug, Clone, Deserialize)]
struct CharsetConfig {
    lang: Option<String>,
    ascii_only: Option<bool>,
}

#[derive(Debug, Clone, Deserialize)]
struct ColorsConfig {
    enabled: Option<bool>,
    truecolor: Option<bool>,
    palette: Option<Palette>,
    enable: Option<ColorToggles>,
```

hawkw: also, it might be worth adding comments explaining that these types are used so that the structure of the config file can be different from how the CLI arguments are structured?

tokio-console/src/config.rs
```diff
@@ -107,20 +112,66 @@ pub struct ViewOptions {
...
 }
 
 // === impl Config ===
 
 impl Config {
+    pub fn from_config() -> color_eyre::Result<Self> {
```
hawkw: i'm not sure if `from_config()` is the best name for this, since it also parses the config from the command line arguments as well?

tokio-console/src/config.rs
```diff
@@ -167,7 +218,9 @@ impl Config {
 
 impl ViewOptions {
     pub fn is_utf8(&self) -> bool {
-        self.lang.ends_with("UTF-8") && !self.ascii_only
+        let lang = self.lang.as_deref().unwrap_or_default();
+        let ascii_only = self.ascii_only.unwrap_or(true);
+        lang.ends_with("UTF-8") && !ascii_only
```
hawkw: nit: it occurs to me that we can simplify this a bit if we check the value of `self.ascii_only` _before_ we touch `self.lang`:
```suggestion
        if !self.ascii_only.unwrap_or(true) {
            return false;
        }
        self.lang.as_deref().unwrap_or_default().ends_with("UTF-8") 
```
this has the advantage of avoiding the string comparison when the `ascii_only` flag is set.

hawkw (changes requested):
This looks really close! I commented on a way to improve the error handling for parse errors, which you asked about in https://github.com/tokio-rs/console/pull/320#issuecomment-1093875927. I also had some small code style suggestions.

hawkw:
@nrskt:
> I fixed the implementations and try to test again.

Thanks! Overall, this looks great! 

> I have questions about case7 and case8. please check it.
> ### case7) Parse error
> 
> [Question] I want you to tell me how to output a better error message.
> 
> XDG_CONFIG/console.toml
> 
> ```toml
> [colors]
> palette = "8"
> ```
> 
> ./console.toml
> 
> ```
> [charset]
> ascii_only = "string value"
> ```
> 
> ```
> ❯ env -u LANG -u COLORTERM RUST_LOG=debug cargo run
>     Finished dev [unoptimized + debuginfo] target(s) in 0.06s
>      Running `target/debug/tokio-console`
> failed to parse config file: Error {
>     msg: "failed to parse Some(\"./console.toml\")",
>     source: Error {
>         inner: ErrorInner {
>             kind: Custom,
>             line: Some(
>                 1,
>             ),
>             col: 13,
>             at: Some(
>                 23,
>             ),
>             message: "invalid type: string \"string value\", expected a boolean",
>             key: [
>                 "charset",
>                 "ascii_only",
>             ],
>         },
>     },
> }
> ```

My suggestions https://github.com/tokio-rs/console/pull/320#discussion_r847555390 and  https://github.com/tokio-rs/console/pull/320#discussion_r847566966 here explain how we can get a nicer error message for parse errors.

> ### case8) Ignore unknown config field
> 
> [Question] Did you expect this behavior? I don't know how to check strictly.
> 
> XDG_CONFIG/console.toml
> 
> ```toml
> 
> ```

Hmm, I think we should probably emit some kind of error or warning on unknown fields... Looking at the `toml` crate, I notice that they do have an error variant for a struct containing unexpected fields: https://github.com/alexcrichton/toml-rs/blob/master/src/de.rs#L184-L192 so, it seems weird that it's not returning an error in this case. Could be a bug upstream?

tokio-console/src/main.rs
nrskt: Thanks, I have tried it and getting the expected result. https://github.com/tokio-rs/console/pull/320/commits/c87ddfce18b6a169cbafa80f8886da8dbc8585e5

```
❯ RUST_LOG=debug cargo run
     Running `target/debug/tokio-console`
Error: failed to parse ./console.toml

Caused by:
    invalid type: string "string value", expected a boolean for key `charset.ascii_only` at line 2 column 14

Location:
    tokio-console/src/config.rs:337:14
```

tokio-console/src/config.rs
nrskt: surely, It's right. I choose `Home`.
https://github.com/tokio-rs/console/pull/320/commits/2da4fdfc4b77ebf098130a8409c9ae6218cfc82c

tokio-console/src/config.rs
nrskt: > This approach has one big advantage

Sounds great! I agree with your opinion.
https://github.com/tokio-rs/console/pull/320/commits/682a2613e1ccfc9d3e715877f2b83de9fdbe7c34

nrskt:
@hawkw 

Thanks reviewing. I have made corrections to the areas you pointed out. Please review again.

> Hmm, I think we should probably emit some kind of error or warning on unknown fields... Looking at the toml crate, I notice that they do have an error variant for a struct containing unexpected fields: https://github.com/alexcrichton/toml-rs/blob/master/src/de.rs#L184-L192 so, it seems weird that it's not returning an error in this case. Could be a bug upstream?

I'll investigate this one after tomorrow.

tokio-console/src/config.rs
```diff
@@ -107,20 +112,62 @@ pub struct ViewOptions {
...
+    truecolor: Option<bool>,
+    palette: Option<Palette>,
+    enable: Option<ColorToggles>,
 }
 
 // === impl Config ===
 
 impl Config {
+    /// Parse from config files and command line options.
+    pub fn parse() -> color_eyre::Result<Self> {
```
hawkw, line 153: one last naming nit, take it or leave it: I'm not the _biggest_ fan of calling this `parse`, as it conflicts with the `Clap::parse` implementation, which is a bit confusing. but, I can't immediately think of a better name for it, so :woman_shrugging:

hawkw (approved):
> Thanks reviewing. I have made corrections to the areas you pointed out. Please review again.

This looks great to me, and I'm excited to go ahead and merge it --- thanks for all your hard work on this, I'm really happy with how it's turned out.

> > Hmm, I think we should probably emit some kind of error or warning on unknown fields... Looking at the toml crate, I notice that they do have an error variant for a struct containing unexpected fields: https://github.com/alexcrichton/toml-rs/blob/master/src/de.rs#L184-L192 so, it seems weird that it's not returning an error in this case. Could be a bug upstream?
> 
> I'll investigate this one after tomorrow.

Since this is probably an issue with an upstream crate, I don't want to block merging this PR on it. Instead, we should just investigate whether it can be changed upstream, and then come back and make any additional changes as necessary. Thanks for looking into it!

hawkw:
Once this is merged, I think we'll probably also want to add documentation about using the config file, but I'm happy to do that in a separate branch.

hawkw:
I opened #323 for some potential follow-up work.

PR #324, opened 2022-04-12
feat(console): add `gen-config` subcommand to generate a config file

hawkw:
This branch builds upon the config file support added in #320 and adds a
new `gen-config` subcommand to the `tokio-console` CLI. This command
generates a new config TOML file with all config options populated with
their default values (overridden by any CLI arguments passed to the
console).

This can be used when generating a new configuration file, so that the
user can see all the default values. We can also use it to generate a
config file for the documentation, which can be automatically kept in
sync with the config file definintion in the app.

This seems much nicer than hand-writing a config file for the docs, as
the struct definitions in the `config` module serve as a single source
of truth for the config file definition.

## Example usage:

```shell
:; cargo run -- gen-config
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/tokio-console gen-config`
[charset]
lang = 'en_US.UTF-8'
ascii_only = false

[colors]
enabled = true
truecolor = true
palette = 'all'

[colors.enable]
durations = true
terminated = true
```

### DIFF
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -1,7 +1,7 @@
 use crate::view::Palette;
-use clap::{ArgGroup, Parser as Clap, ValueHint};
+use clap::{ArgGroup, Parser as Clap, Subcommand, ValueHint};
 use color_eyre::eyre::WrapErr;
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use std::fs;
 use std::ops::Not;
 use std::path::PathBuf;
@@ -16,6 +16,7 @@ use tonic::transport::Uri;
     author,
     about,
     version,
+    propagate_version = true,
 )]
 #[deny(missing_docs)]
 pub struct Config {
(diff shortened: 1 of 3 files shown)


Milo123459:
Could this config file perhaps be used to set things like the port it is served on? And if it uses Unix Domain Socket?

hawkw:
> Could this config file perhaps be used to set things like the port it is served on? And if it uses Unix Domain Socket?

This is a config file for the `tokio-console` command-line interface, not for the application itself.

You can use the [builder API](https://docs.rs/console-subscriber/latest/console_subscriber/struct.Builder.html) to configure the application's instrumentation from values read from a config file. The `console-subscriber` crate doesn't implement configuration file parsing itself, because the goal is for it to be able to be integrated with an application's existing configuration mechanisms. The `tokio-console` command-line interface, on the other hand, is not part of a larger application; it's a stand-alone binary. Therefore, it implements its own configuration system.

Milo123459:
My bad, I misunderstood.

tokio-console/src/config.rs
```diff
@@ -296,6 +325,22 @@ impl ViewOptions {
     }
 }
 
+impl Default for ViewOptions {
+    fn default() -> Self {
+        Self {
+            no_colors: Some(false),
```
seanmonstar, line 331: I could be misunderstanding, but I would normally assume an `Option<bool>` is because we want to represent the state where the user didn't specify a preference. So, should we default to that state?

hawkw: The `Option<bool>` is used here to allow parsing a config file that doesn't specify all the fields --- if it wasn't an `Option`, `serde` would emit an error if the TOML doesn't contain that field. If we defaulted all of these values to `None`, though, the  default config file would be empty. Instead, we want it to represent what the program would actually _do_ if there was no config file.

seanmonstar (approved):

PR #326, opened 2022-04-13
Fix links to console-subscriber's API docs

sebasmagri:
Links to builder and init APIs were broken.


hawkw (approved):
whoops! thanks for the fix!

PR #330, opened 2022-04-14
feat(console): parse error when use unknown fields

nrskt:
This work is related to https://github.com/tokio-rs/console/pull/320#issuecomment-1095416429
return error message if the config file includes unknown fields.  

reference: https://serde.rs/container-attrs.html#deny_unknown_fields

## test

console.toml

```toml
[charset]
lang = "en_us.UTF-8"
ascii_only = true

[colors]
enabled = true
palette = "all"
truecolor = true

[colors.enable]
durations = true 
terminated = true
unknown = true  # unknown field
```

```
❯ cargo run
     Running `target/debug/tokio-console`
Error: failed to parse ./console.toml

Caused by:
    unknown field `unknown`, expected `durations` or `terminated` for key `colors.enable` at line 10 column 1

Location:
    tokio-console/src/config.rs:388:14
```

### DIFF
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -135,6 +135,7 @@ pub struct ViewOptions {
 
 /// Toggles on and off color coding for individual UI elements.
 #[derive(Clap, Debug, Copy, Clone, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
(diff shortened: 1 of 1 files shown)


hawkw (approved):
great, looks good to me! thanks for figuring this out!

Issue #331, opened 2022-04-14
console: add missing configuration options to the config file


hawkw:
### What problem are you trying to solve?

Currently, some configuration options can only be set from the command line: the target address, the console log filter, and the retention interval. This is because these fields are only present on the `Config` struct, and not the `ConfigFile` struct:

https://github.com/tokio-rs/console/blob/3a67d476835e4a1d3557190b85f5a89c760490bb/tokio-console/src/config.rs#L23-L69

It should probably be possible to set at least some of these values from the config file, as well.

### How should the problem be solved?

We should probably add these values to the `ConfigFile` struct and support setting them in the config file.

### Any alternatives you've considered?

I'm somewhat on the fence as to whether it makes sense to set the target address in the config file. I think it makes sense, given that the current config overriding system means that the config file would be setting the _default_ target address, and it would be overridden by the CLI. So, it might make sense to name the config key for this `default_target_addr` or similar.

### Would you like to work on this feature?

maybe

hawkw:
@nrskt perhaps you'd be interested in working on this one as well?

nrskt:
thanks, I'm interesting this work. I’ll take it😁

Issue #323, opened 2022-04-12
console: enable and disable warnings from config files


hawkw:
### What problem are you trying to solve?

The console supports a variety of warnings which indicate potential issues in the application being monitored. However, a warning is just that, a _warning_ --- a warning may in some cases be generated by a task that is behaving correctly. Therefore, in some cases, users may want to disable some warnings for a particular project.

### How should the problem be solved?

PR #320 adds support for reading from a TOML config file to configure the console CLI. Currently, the config file has all of the same configuration options as the console CLI's command-line arguments. 

We should add a new section to the configuration file for enabling or disabling warnings by name.

### Any alternatives you've considered?

We could also configure warnings from a CLI argument. I think it's important to make the warnings configurable from the config file, because it allows a project to enable or disable certain warnings in a way that can be checked into a source control repository for that project. However, it may also be desirable to support enabling/disabling warnings as a CLI argument _as well_ as from the config file.

### How would users interact with this feature?

A sample config file might look like this:

```toml
# ... other configurations ...

[warnings]
# enable the "lost waker" warning
lost_waker = true
# disable the "self wake percent" warning
self_wake_percent = false
```

If a particular project wants to ignore a lint for all developers, that configuration file can be checked into source control, since the console will override the user-level default settings from a directory-local `console.toml`.

### Would you like to work on this feature?

maybe

hawkw:
A potential follow-up might also be to support configuration _values_ associated with warnings, such as the threshold percentage for the self-wake warning.

0xPoe:
I am working on this.

hawkw:
Related: https://github.com/tokio-rs/console/issues/148

PR #334, opened 2022-04-16
feat(console): add missing configuration

nrskt:
This changes supports `target_addr`, `env_filter`, and `retain_for` in configuration file.

```
❯ env -u LANG -u COLORTERM -u RUST_LOG cargo run -- gen-config
     Running `target/debug/tokio-console gen-config`
target_addr = 'http://127.0.0.1:6669/'
env_filter = 'info'
retain_for = '6s'

[charset]
lang = 'en_us.UTF8'
ascii_only = false

[colors]
enabled = true
truecolor = true
palette = 'all'

[colors.enable]
durations = true
terminated = true
```

Related to #331 

## Note
- should I use define the table in the config file, like `[charset]`?
- is it better a field name `log` instead of `env_filter`?
- what is the default value for `env_filter`?
    - I used `info` now.
- these changes may conflict with #323

### DIFF
--- a/tokio-console/args.example
+++ b/tokio-console/args.example
@@ -34,7 +34,7 @@ OPTIONS:
             application runs interactively, stderr should generally be redirected to a file to avoid
             interfering with the console's text output.
             
-            [env: RUST_LOG=info]
+            [env: RUST_LOG=]
 
         --no-colors <no-colors>
             Disable ANSI colors entirely
--- a/tokio-console/args.example
+++ b/tokio-console/args.example
@@ -36,6 +36,8 @@ OPTIONS:
             application runs interactively, stderr should generally be redirected to a file to avoid
             interfering with the console's text output.
             
+            [default: off]
+            
             [env: RUST_LOG=]
 
         --no-colors <no-colors>
@@ -80,6 +82,8 @@ OPTIONS:
             * `months`, `month`, `M` -- defined as 30.44 days
             
             * `years`, `year`, `y` -- defined as 365.25 days
+            
+            [default: 6s]
 
     -V, --version
             Print version information
--- a/tokio-console/args.example
+++ b/tokio-console/args.example
@@ -6,8 +6,6 @@ ARGS:
             The address of a console-enabled process to connect to.
             
             This may be an IP address and port, or a DNS name.
-            
-            [default: http://127.0.0.1:6669]
 
 OPTIONS:
         --ascii-only <ASCII_ONLY>
@@ -36,8 +34,7 @@ OPTIONS:
             application runs interactively, stderr should generally be redirected to a file to avoid
             interfering with the console's text output.
             
-            [env: RUST_LOG=]
-            [default: off]
+            [env: RUST_LOG=info]
 
         --no-colors <no-colors>
             Disable ANSI colors entirely
@@ -81,8 +78,6 @@ OPTIONS:
             * `months`, `month`, `M` -- defined as 30.44 days
             
             * `years`, `year`, `y` -- defined as 365.25 days
-            
-            [default: 6s]
 
     -V, --version
             Print version information
--- a/tokio-console/args.example
+++ b/tokio-console/args.example
@@ -6,6 +6,8 @@ ARGS:
             The address of a console-enabled process to connect to.
             
             This may be an IP address and port, or a DNS name.
+            
+            [default: http://127.0.0.1:6669]
 
 OPTIONS:
         --ascii-only <ASCII_ONLY>
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -177,7 +177,7 @@ pub struct ColorToggles {
 #[derive(Debug, Clone, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
 struct ConfigFile {
-    target_addr: Option<String>,
+    default_target_addr: Option<String>,
     log: Option<String>,
     retention: Option<RetainFor>,
     charset: Option<CharsetConfig>,
@@ -459,14 +459,14 @@ impl ConfigFile {
 
     fn target_addr(&self) -> color_eyre::Result<Option<Uri>> {
         let uri = self
-            .target_addr
+            .default_target_addr
             .as_ref()
             .map(|addr| addr.parse::<Uri>())
             .transpose()
             .wrap_err_with(|| {
                 format!(
                     "failed to parse target address {:?} as URI",
-                    self.target_addr
+                    self.default_target_addr
                 )
             })?;
         Ok(uri)
@@ -508,7 +508,7 @@ impl ConfigFile {
 impl From<Config> for ConfigFile {
     fn from(config: Config) -> Self {
         Self {
-            target_addr: config.target_addr.map(|addr| addr.to_string()),
+            default_target_addr: config.target_addr.map(|addr| addr.to_string()),
             log: config.env_filter.map(|filter| filter.to_string()),
             retention: config.retain_for,
             charset: Some(CharsetConfig {
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -457,7 +457,12 @@ impl ConfigFile {
             .as_ref()
             .map(|addr| addr.parse::<Uri>())
             .transpose()
-            .wrap_err(format!("failed to parse Uri, {:?}", self.target_addr))?;
+            .wrap_err_with(|| {
+                format!(
+                    "failed to parse target address {:?} as URI",
+                    self.target_addr
+                )
+            })?;
         Ok(uri)
     }
 
@@ -467,7 +472,7 @@ impl ConfigFile {
             .as_ref()
             .map(|directive| directive.parse::<tracing_subscriber::EnvFilter>())
             .transpose()
-            .wrap_err(format!("failed to parse EnvFilter, {:?}", self.log))?;
+            .wrap_err_with(|| format!("failed to parse log filter {:?}", self.log))?;
         Ok(env_filter)
     }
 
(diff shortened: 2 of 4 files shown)


tokio-console/args.example
```diff
@@ -6,8 +6,6 @@ ARGS:
             The address of a console-enabled process to connect to.
             
             This may be an IP address and port, or a DNS name.
-            
-            [default: http://127.0.0.1:6669]
```
hawkw: it's a shame that we can no longer show default values in the CLI help. it might be worth manually adding the defaults to the doc comments that generate the help message.

tokio-console/src/config.rs
```diff
@@ -152,6 +162,9 @@ pub struct ColorToggles {
 #[derive(Debug, Clone, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
 struct ConfigFile {
+    target_addr: Option<String>,
+    env_filter: Option<String>,
```
hawkw: as per your question
> is it better a field name log instead of env_filter?

 i think we should probably call this
```suggestion
    log: Option<String>,
```

tokio-console/src/config.rs
```diff
@@ -152,6 +162,9 @@ pub struct ColorToggles {
...
+    retain_for: Option<String>,
```
hawkw: i would probably call this
```suggestion
    retention: Option<String>,
```
in the config file

tokio-console/src/config.rs
```diff
@@ -394,33 +440,34 @@ impl ConfigFile {
...
-                color_terminated: self.color_terminated(),
-            },
-        }
+    fn target_addr(&self) -> color_eyre::Result<Option<Uri>> {
+        let uri = self
+            .target_addr
+            .as_ref()
+            .map(|addr| addr.parse::<Uri>())
+            .transpose()
+            .wrap_err(format!("failed to parse Uri, {:?}", self.target_addr))?;
```
hawkw: nit: let's prefer to use `wrap_err_with` when we're emitting a formatted `String`, as `wrap_err_with` will avoid allocating to format the string when no error occurs
```suggestion
            .wrap_err_with(|| format!("failed to parse Uri, {:?}", self.target_addr))?;
```

hawkw: also, i might change the message a bit:
```suggestion
            .wrap_err_with(|| format!("failed to parse target address {:?} as URI", self.target_addr))?;
```

tokio-console/src/config.rs
```diff
@@ -394,33 +440,34 @@ impl ConfigFile {
...
-                enable: Some(view_options.toggles),
-            }),
-        }
+    fn env_filter(&self) -> color_eyre::Result<Option<tracing_subscriber::EnvFilter>> {
+        let env_filter = self
+            .env_filter
+            .as_ref()
+            .map(|directive| directive.parse::<tracing_subscriber::EnvFilter>())
+            .transpose()
+            .wrap_err(format!("failed to parse EnvFilter, {:?}", self.env_filter))?;
```
hawkw: how about:
```suggestion
            .wrap_err_with(|| format!("failed to parse log filter {:?}", self.env_filter))?;
```

tokio-console/src/config.rs
```diff
@@ -394,33 +440,34 @@ impl ConfigFile {
...
+        Ok(env_filter)
+    }
+
+    fn retain_for(&self) -> color_eyre::Result<Option<RetainFor>> {
+        let retain_for = self
+            .retain_for
+            .as_ref()
+            .map(|retain_for| retain_for.parse::<RetainFor>())
+            .transpose()
+            .wrap_err(format!("failed to parse RetainFor, {:?}", self.retain_for))?;
```
hawkw:
```suggestion
            .wrap_err_with(|| format!("failed to parse retention interval {:?} as a duration", self.retain_for))?;
```

tokio-console/src/config.rs
```diff
@@ -23,16 +24,16 @@ pub struct Config {
     /// The address of a console-enabled process to connect to.
     ///
     /// This may be an IP address and port, or a DNS name.
```
hawkw, line 26: we can probably get the default values displayed in the command-line `--help` output if we just put them in the comments like this:
```suggestion
    /// This may be an IP address and port, or a DNS name.
    ///
    /// [default: http://127.0.0.1:6669/]
```
(and do the same for the other defaults)

tokio-console/src/config.rs
hawkw: also, take it or leave it, but we could add a `Deserialize` implementation for `RetainFor`, and then change this to just be 
```suggestion
    retain_for: Option<RetainFor>,
```
that way, we wouldn't have to handle deserializing it as a string and then parsing the string later in the `ConfigFile::retain_for` accessor.

i would imagine the `Deserialize` impl would just call the `FromStr` impl, like this:
https://github.com/serde-rs/serde/issues/908#issuecomment-298027413

tokio-console/src/config.rs
```diff
@@ -238,10 +250,49 @@ impl Config {
...
+            subcmd: other.subcmd.or(self.subcmd),
+        }
     }
 }
 
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            target_addr: Some(default_target_addr()),
+            env_filter: Some(tracing_subscriber::EnvFilter::new("info")),
```
hawkw: I think this should default to `off`, like it did previously when set from the CLI arguments
```suggestion
            env_filter: Some(tracing_subscriber::EnvFilter::new("off")),
```

tokio-console/src/config.rs
```diff
@@ -238,10 +250,49 @@ impl Config {
...
+            retain_for: Some(RetainFor(Some(Duration::from_secs(6)))),
+            view_options: ViewOptions::default(),
+            subcmd: None,
+        }
+    }
+}
+
+fn default_target_addr() -> Uri {
+    "http://127.0.0.1:6669".parse::<Uri>().unwrap_or_default()
```
hawkw: a very minor nitpick: since this URI is hardcoded to a value we know is valid, it should never fail to parse; I think we should probably just `expect()` this, rather than calling `unwrap_or_default()`. if we were to ever change the default URI to something incorrect, the `unwrap_or_default()` call would just silently return an empty URI. I think it would be preferable to fail loudly in testing if we changed this value to one that didn't parse.

```suggestion
    "http://127.0.0.1:6669".parse::<Uri>().expect("default target address should be a valid URI")
```

hawkw (changes requested):
Sorry I didn't review this earlier, I must have missed it! Overall, the change looks good. I had some small suggestions, but once those are taken care of, I think this should be ready to merge! Thanks!

tokio-console/src/config.rs
nrskt: Resolved. Because I implemented `Deserialize` for `RetainFor`. So we don't need to parse the string it.

tokio-console/args.example
nrskt: I fixed: https://github.com/tokio-rs/console/pull/334/commits/23d396ee7ba1791df1061c2b30126a537e584252

tokio-console/src/config.rs
nrskt: I fixed: https://github.com/tokio-rs/console/pull/334/commits/9473b594cdae614261042c6cf533b79bfaa402da

tokio-console/src/config.rs
nrskt: I fixed: https://github.com/tokio-rs/console/pull/334/commits/00e31553ee118dda9d1e5773c121d542723d49d2

tokio-console/src/config.rs
nrskt: I fixed: https://github.com/tokio-rs/console/pull/334/commits/ded150295c5abfcfd49cba51a321fff3fa39ca28

> nit: let's prefer to use wrap_err_with when we're emitting a formatted String, as wrap_err_with will avoid allocating to format the string when no error occurs

I learned, thank you!

nrskt:
@hawkw
Thanks reviewing. I have made corrections to the areas you pointed out. Please review again.

>Sorry I didn't review this earlier, I must have missed it! 

Don't worry about it:v:

tokio-console/args.example
```diff
@@ -36,8 +36,9 @@ OPTIONS:
             application runs interactively, stderr should generally be redirected to a file to avoid
             interfering with the console's text output.
             
-            [env: RUST_LOG=]
             [default: off]
+            
+            [env: RUST_LOG=]
```
hawkw, line 41: my assumption here is that the extra line of whitespace is inserted by `clap` and we can't get rid of it?

hawkw (approved):
thanks, this looks good to me!

tokio-console/src/config.rs
```diff
@@ -152,6 +177,9 @@ pub struct ColorToggles {
 #[derive(Debug, Clone, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
 struct ConfigFile {
+    target_addr: Option<String>,
```
hawkw: oh, one last thought: it might be a bit clearer for users if we called the config key `default_target_addr` rather than `target_addr`, since a different target can always be connected to by overriding it from the command line?

but, on the other hand, all config keys are overridable, but we don't call the other ones `default_XXX`...

tokio-console/args.example
nrskt: Yes, you're correct. I have tried to remove the extra line but It is difficult.

hawkw: okay, i wouldn't worry about it too much, then. thanks!

tokio-console/src/config.rs
nrskt: Mmm...This is a tough one...Surely `default_target_addr` is clearer for users.
And I think might use `address` instead of `addr`.

Do you want to change `default_target_addr` or `default_target_address`?

hawkw: Either is fine, in my opinion; I think `addr` is a _relatively_ widespread abbreviation, but it's fine for the config file to be a bit more verbose than the CLI args, so `address` is also fine with me.

nrskt: Ok. I fix to use `default_target_addr`.

hawkw: great, thank you!

Issue #325, opened 2022-04-12
console: generate man pages and shell completions


hawkw:
### What problem are you trying to solve?

It would be nice if a man page for `tokio-console` could be generated and installed. Similarly, it would be nice to be able to generate command completion scripts for various shells.

### How should the problem be solved?

The [`clap_mangen`](https://crates.io/crates/clap_mangen) and [`clap_complete`](https://crates.io/crates/clap_complete) crates can be used to generate man pages and shell completions for a `clap` CLI, respectively. We can use these crates to generate man pages and completions for the console.

### Any alternatives you've considered?

I'm not sure the best way to _distribute_ man pages/completions. I've seen some people generating them in a CI job and committing them to the repo when the CLI changes. This could be fine if we expect the console to be distributed primarily by package managers. However, if we expect people to get it using `cargo install`, I don't _believe_ there's a nice way for `cargo install` to install completions or man pages. Instead, we might want to consider having subcommands in the binary itself for generating completions/man pages, so that users can `cargo install` the console and then install the shell completions/man pages?

### Would you like to work on this feature?

maybe

nrskt:
I’d like to take it.

PR #336, opened 2022-04-26
feat(console): add subcommand for gen completions

nrskt:
This PR includes adding subcommand for shell completions. Generating a man page will be done as another PR.

```
❯ tokio-console
--ascii-only            --lang                  --no-duration-colors    --retain-for            -h                      gen-config
--colorterm             --log                   --no-terminated-colors  --version               <TARGET_ADDR>           help
--help                  --no-colors             --palette               -V                      gen-completion

❯ tokio-console gen-completion
--help      --version   -V          -h          bash        elvish      fish        powershell  zsh
```

Related to #325

### DIFF
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -105,7 +105,7 @@ pub enum OptionalCmd {
     ///
     /// For example, using zsh:
     ///
-    ///     $ tokio-console gen-completion zsh > $FPATH/_tokio_console
+    ///     $ tokio-console gen-completion zsh > ~/.zsh_functions/_tokio_console
     ///
     GenCompletion {
         #[clap(arg_enum)]
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -32,13 +32,21 @@ pub struct Config {
 
     /// Log level filter for the console's internal diagnostics.
     ///
-    /// The console will log to stderr if a log level filter is provided. Since
-    /// the console application runs interactively, stderr should generally be
-    /// redirected to a file to avoid interfering with the console's text output.
+    /// Logs are written to a new file at the path given by the `--log-dir`
+    /// argument (or its default value), or to the system journal if
+    /// `systemd-journald` support is enabled.
+    ///
+    /// If this is set to 'off' or is not set, no logs will be written.
     ///
     /// [default: off]
     #[clap(long = "log", env = "RUST_LOG")]
     pub(crate) env_filter: Option<tracing_subscriber::EnvFilter>,
+
+    /// Path to a directory to write the console's internal logs to.
+    ///
+    /// [default: /tmp/tokio-console/logs]
+    #[clap(long = "log-dir", value_hint = ValueHint::DirPath)]
+    pub(crate) log_directory: Option<PathBuf>,
 
     #[clap(flatten)]
     pub(crate) view_options: ViewOptions,
@@ -135,8 +143,8 @@ impl Serialize for RetainFor {
 #[clap(group = ArgGroup::new("colors").conflicts_with("no-colors"))]
 pub struct ViewOptions {
     /// Disable ANSI colors entirely.
-    #[clap(name = "no-colors", long = "no-colors")]
-    no_colors: Option<bool>,
+    #[clap(name = "no-colors", long = "no-colors", takes_value = false)]
+    no_colors: bool,
 
     /// Overrides the terminal's default language.
     #[clap(long = "lang", env = "LANG")]
@@ -192,6 +200,7 @@ pub struct ColorToggles {
 struct ConfigFile {
     default_target_addr: Option<String>,
     log: Option<String>,
+    log_directory: Option<PathBuf>,
     retention: Option<RetainFor>,
     charset: Option<CharsetConfig>,
     colors: Option<ColorsConfig>,
@@ -241,8 +250,13 @@ impl Config {
     }
 
     pub fn trace_init(&mut self) -> color_eyre::Result<()> {
-        let filter = std::mem::take(&mut self.env_filter);
         use tracing_subscriber::prelude::*;
+        let filter = match self.env_filter.take() {
+            // if logging is totally disabled, don't bother even constructing
+            // the subscriber
+            None => return Ok(()),
+            Some(filter) => filter,
+        };
 
         // If we're on a Linux distro with journald, try logging to the system
         // journal so we don't interfere with text output.
@@ -256,13 +270,41 @@ impl Config {
         #[cfg(not(all(feature = "tracing-journald", target_os = "linux")))]
         let should_fmt = true;
 
-        // Otherwise, log to stderr and rely on the user redirecting output.
+        // Otherwise, log to a file.
         let fmt = if should_fmt {
-            Some(
-                tracing_subscriber::fmt::layer()
-                    .with_writer(std::io::stderr)
-                    .with_ansi(atty::is(atty::Stream::Stderr)),
-            )
+            let dir = self
+                .log_directory
+                .take()
+                .unwrap_or_else(default_log_directory);
+
+            // first ensure that the log directory exists
+            fs::create_dir_all(&dir)
+                .with_context(|| format!("creating log directory '{}'", dir.display()))?;
+            color_eyre::eyre::ensure!(
+                dir.is_dir(),
+                "log directory path '{}' is not a directory",
+                dir.display()
+            );
+
+            // now, open a log file
+            let now = std::time::SystemTime::now();
+            // format the current time in a way that's appropriate for a
+            // filename (strip the `:` character, as it is an invalid filename
+            // char on windows)
+            let filename =
+                format!("{}.log", humantime::format_rfc3339_seconds(now)).replace(':', "");
+            let path = dir.join(filename);
+            let file = fs::File::options()
+                .create_new(true)
+                .write(true)
+                .open(&path)
+                .with_context(|| format!("creating log file '{}'", path.display()))?;
+
+            // finally, construct a `fmt` layer to write to that log file
+            let fmt = tracing_subscriber::fmt::layer()
+                .with_ansi(false)
+                .with_writer(file);
+            Some(fmt)
         } else {
             None
         };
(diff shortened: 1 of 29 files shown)


tokio-console/src/config.rs
```diff
@@ -95,6 +96,17 @@ pub enum OptionalCmd {
     ///     $ tokio-console gen-config > console.toml
     ///
     GenConfig,
+
+    /// Generate shell completions
+    ///
+    /// It should be saved in expected directory, depending on the shell used
+    ///
+    ///     $ tokio-console gen-completion zsh > $FPATH/_tokio_console
```
hawkw: since we know the shell from the command-line argument, should we just make the command install the completions to the correct location, instead?

nrskt: @hawkw Maybe it can install the completion but I think that It's better not to do it because the user wants to decide where is put.

tokio-console/src/main.rs
```diff
@@ -34,6 +36,12 @@ async fn main() -> color_eyre::Result<()> {
         return Ok(());
     }
 
+    if let Some(config::OptionalCmd::GenCompletion { shell }) = args.subcmd {
+        let mut app = config::Config::command();
+        generate(shell, &mut app, "tokio-console", &mut std::io::stdout());
+        return Ok(());
+    }
```
hawkw: style, take it or leave it: now that we have multiple subcommands, we might want to consider collapsing this and the `if` block for the `GenConfig` subcommand into a `match`, like
```rust
match args.subcmd {
    Some(config::OptionalCmd::GenConfig) => // ...
    Some(config::OptionalCmd::GenCompletion { shell } => // ...
    None => {},
}
```

tokio-console/src/config.rs
```diff
@@ -95,6 +96,17 @@ pub enum OptionalCmd {
     ///     $ tokio-console gen-config > console.toml
     ///
     GenConfig,
+
+    /// Generate shell completions
```
hawkw, line 108: maybe worth adding:
```suggestion
    /// Generate shell completions
    ///
    /// The completion script will be written to stdout.
```

tokio-console/src/config.rs
```diff
@@ -95,6 +96,17 @@ pub enum OptionalCmd {
...
+    ///
+    /// It should be saved in expected directory, depending on the shell used
```
hawkw: maybe something like this:
```suggestion
    /// The completion script should be saved in the shell's completion directory.
    /// This depends on which shell is in use.
    ///
    /// For example, using zsh:
```

tokio-console/src/config.rs
hawkw: hmm, isn't the completion location determined by the shell, rather than the user? i'm fine with the current approach, but think this is worth a potential follow-up.

tokio-console/src/main.rs
nrskt: fixed: https://github.com/tokio-rs/console/pull/336/commits/b51ce10be556e7371889d0a84ed162c03241b6d9

tokio-console/src/config.rs
nrskt: fixed: https://github.com/tokio-rs/console/pull/336/commits/60fbce5744bf9776157cfbcfce9d714fefb7f460

nrskt:
@hawkw 
> hmm, isn't the completion location determined by the shell, rather than the user? i'm fine with the current approach, but think this is worth a potential follow-up.

Ok, I understand your opinion. I'll check how to completion directory the shell that I don't know (fish, elvish, powershell).
Should I fix it with another pull request? or continually this pull request?

tokio-console/src/config.rs
```diff
@@ -95,6 +96,23 @@ pub enum OptionalCmd {
...
+
+    /// Generate shell completions
+    ///
+    /// The completion script will be written to stdout.
+    /// The completion script should be saved in the shell's completion directory.
+    /// This depends on which shell is in use.
+    ///
+    /// For example, using zsh:
+    ///
+    ///     $ tokio-console gen-completion zsh > ~/.zsh_functions/_tokio_console
```
nrskt: @hawkw 
Sorry, I'm mistaken.
`$FPATH` has multiple path values so it's not work `$FPATH/_tokio_console`

I worried about this sample, because `.zsh_functions` directory name. This name is my config value on .zshrc. I don't know common name.

https://github.com/tokio-rs/console/pull/336/commits/fa91ca9f9f46f7daf6a4ee7c05f69c51cfbe5a1b

tokio-console/src/main.rs
```diff
@@ -214,3 +223,24 @@ async fn watch_details_stream(
...
+                Box::new(std::io::BufWriter::new(f))
+            }
+            _ => color_eyre::eyre::bail!("Not support to install completion script on {}", shell),
+        }
+    } else {
+        Box::new(std::io::stdout())
+    };
+    generate(shell, &mut app, "tokio-console", &mut buf);
+    Ok(())
+}
```
nrskt: @hawkw 
How about this approach?
If the user runs as `tokio-console gen-completion --install <SHELL>` then put the script file.

But I must fix it, `.zsh_functions` directory is not correct. I have a problem.
I have tried to get a directory with `$FPATH` environment variables however I failed.

problem: 
`std::env::var("FPATH")` return an error. I checked on my terminal like this:

```
❯ echo $FPATH
/home/nrskt/.zsh_functions:....some paths

❯ env | rg FPATH
(return nothing)
```
So `FPATH` not working in my environment. I can't determine if it's a problem with my environment or zsh features.

---

In my opinion, first I fix it to not support all shells like this:

```
match shell {
    _ => color_eyre::eyre::bail!("Not support to install completion script on {}", shell),
}
```

And the rest should be treated for potential follow-up.

tokio-console/src/main.rs
```diff
@@ -214,3 +223,24 @@ async fn watch_details_stream(
         }
     }
 }
+
+fn gen_completion(install: bool, shell: Shell) -> color_eyre::Result<()> {
```
hawkw: nit, take it or leave it: I think it might make sense to put this function in the `config` module?

tokio-console/src/main.rs
hawkw: hmm...it seems like automatically installing completions may be a bit more complicated than i'd hoped. i think it's reasonable to just always fail if `--install` is passed, for now, so we can go ahead and merge this branch, and implement support for automatic installations later.

tokio-console/src/main.rs
```diff
@@ -214,3 +223,24 @@ async fn watch_details_stream(
...
+                let f = std::fs::File::create(&home_dir)
+                    .wrap_err_with(|| format!("fail to open {}", home_dir.display()))?;
+                Box::new(std::io::BufWriter::new(f))
+            }
+            _ => color_eyre::eyre::bail!("Not support to install completion script on {}", shell),
+        }
+    } else {
+        Box::new(std::io::stdout())
+    };
+    generate(shell, &mut app, "tokio-console", &mut buf);
```
hawkw: nit, take it or leave it: i would probably not import the `generate` function into this file globally, as it's not _super_ clear what's being generated this way. i would probably have written
```suggestion
    clap_complete::generate(shell, &mut app, "tokio-console", &mut buf);
```

but, it's not a big deal either way.

tokio-console/src/main.rs
```diff
@@ -214,3 +223,24 @@ async fn watch_details_stream(
         }
     }
 }
+
+fn gen_completion(install: bool, shell: Shell) -> color_eyre::Result<()> {
+    let mut app = config::Config::command();
+    let mut buf: Box<dyn std::io::Write> = if install {
+        let mut home_dir = dirs::home_dir()
+            .ok_or_else(|| color_eyre::eyre::eyre!("fail to find home directory"))?;
```
hawkw: nit
```suggestion
            .ok_or_else(|| color_eyre::eyre::eyre!("Failed to find home directory"))?;
```

tokio-console/src/main.rs
```diff
@@ -214,3 +223,24 @@ async fn watch_details_stream(
...
+        match shell {
+            Shell::Zsh => {
+                home_dir.push(".zsh_functions/_tokio_console");
+                let f = std::fs::File::create(&home_dir)
+                    .wrap_err_with(|| format!("fail to open {}", home_dir.display()))?;
```
hawkw:
```suggestion
                    .wrap_err_with(|| format!("Failed to open home directory {}", home_dir.display()))?;
```

tokio-console/src/main.rs
```diff
@@ -214,3 +223,24 @@ async fn watch_details_stream(
...
+                Box::new(std::io::BufWriter::new(f))
+            }
+            _ => color_eyre::eyre::bail!("Not support to install completion script on {}", shell),
```
hawkw: maybe
```suggestion
            _ => color_eyre::eyre::bail!("Automatically installing completion scripts is not currently supported on {}", shell),
```

tokio-console/src/config.rs
hawkw: hmm, if this isn't going to work for everyone, maybe we should remove the sample for now, and add it back later if we can come up with one that will work universally?

hawkw (commented):
Overall, this looks good! I had some small suggestions.

I think it's fine to save the automatic installation of completion scripts for later, since it looks like it's actually somewhat complex. For now, let's go with what you suggested and have that always fail, and the user can still install completions manually.

tokio-console/src/main.rs
nrskt: fixed: https://github.com/tokio-rs/console/pull/336/commits/99f659f44532082f200e9cce0efa7996aaa0fb51

tokio-console/src/main.rs
nrskt: fixed: https://github.com/tokio-rs/console/pull/336/commits/0c02ebfce52d1989034aa4ac140df64f12999bec

nrskt:
@hawkw Fixed based on your suggestion. How about it?

tokio-console/src/config.rs
```diff
@@ -95,6 +96,18 @@ pub enum OptionalCmd {
...
+    /// Generate shell completions
+    ///
+    /// The completion script will be written to stdout.
+    /// The completion script should be saved in the shell's completion directory.
+    /// This depends on which shell is in use.
+    GenCompletion {
+        #[clap(name = "install", long = "install")]
+        install: bool,
+        #[clap(arg_enum)]
+        shell: Shell,
```
hawkw, line 117: i wonder if it's worthwhile to also try to read this from the value of the `$SHELL` environment variable, if it's not provided by the user? we could add that in a follow-up.

hawkw (approved):
okay, this looks good to me! thanks!

PR #338, opened 2022-04-28
docs: Update minimal Rust version

shanesveller:
In #318, the CI version was changed to Rust 1.58+ and installation of tokio-console 0.1.4 on Rust 1.56-1.57 fails on that version as noted in the `ci.yaml` changes in commit 83d8a870bcc40be71bc23d0f45fc374899c636a8.

Please let me know if there are any additional changes needed (i.e. version number bump due to altering `rust-version` in Cargo.toml or changelog entry), contribution process I overlooked , etc.

Thanks for this project!


hawkw (approved):
thanks for catching that!

it would be nice if we had some kind of automation/testing to ensure these are kept in sync with the version on ci...

PR #340, opened 2022-04-29
fix(console): always log to a file instead of `stderr`

### LINKED ISSUE #339: TUI messed up on debian

tokio-console

### What is the issue?

This is what I saw on debian unstable with xfce-terminal.
After this I tried on latest tag and all it's ok.
The SIGWINCH was trapped in right way and the TUI was rendered well only for 1s on resize event.

image

Running xfce-terminal in debian is enought.

### Logs, error output, etc

_No response_

### Versions

472ff52e6445dd2c103b218d60a4e0cad9a1972e

### Possible solution

_No response_

### Additional context

_No response_

maybe

hawkw:
## Motivation

Currently, the `tokio-console` CLI includes `tracing` support for
internal debugging. By default, the logs are written to `stderr`, with
the expectations that users will redirect them to a file if logging is
enabled. 

This isn't great: it means that if a `RUST_LOG` env var is set, but the
user didn't intend to enable the console's logging, logs will be written
to the terminal *while displaying the console TUI*. This results in the
TUI itself getting all messed up (see #339).

If support for `systemd-journald` is enabled, logs are instead written
to the system journal, to avoid having to use this redirection dance. We
should probably change our `tracing` configuration so logs are never
written to stderr/stdout, regardless of whether `journald` is enabled.

## Solution

This branch changes the console so that it logs to a file when
`journald` support is not enabled, rather than to stderr. The log file
location can be overridden from the CLI and config file, and defaults to
`/tmp/tokio-console/logs`. This should fix logging messing up the TUI.

Fixes #339

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1430,7 +1430,7 @@ dependencies = [
 
 [[package]]
 name = "tokio-console"
-version = "0.1.3"
+version = "0.1.4"
 dependencies = [
  "atty",
  "clap",
--- a/tokio-console/args.example
+++ b/tokio-console/args.example
@@ -32,13 +32,14 @@ OPTIONS:
         --log <ENV_FILTER>
             Log level filter for the console's internal diagnostics.
             
-            The console will log to stderr if a log level filter is provided. Since the console
-            application runs interactively, stderr should generally be redirected to a file to avoid
-            interfering with the console's text output.
-            
             [default: off]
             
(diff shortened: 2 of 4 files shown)


hawkw:
confirmed that this does fix #339: https://github.com/tokio-rs/console/issues/339#issuecomment-1113984816

guerinoni (approved):
Good :)

PR #344, opened 2022-05-01
feat(config): default no_colors to false

gelfand:
Default `no_colors` setting to false.

### Before

Input below would produce no colors in tokio console, tho `palette` flag is explicitly passed, and `no_colors` is not explicitly set to `true`.
```bash
tokio-console --palette 256 http://127.0.0.1:6669
```
---
Input below also would produce no colors in tokio-console, tho `COLORTERM` is set to `truecolor`, and `no_colors` is not explicitly set to `true`.
```bash
COLORTERM=truecolor tokio-console http://127.0.0.1:6669 
```
---
Also would produce no colors in tokio-console, tho `truecolor` is explicitly passed.
```bash
tokio-console --truecolor http://127.0.0.1:6669
```
---
So to enable colors you must explicitly set `no_colors` to `false`, like so:
```bash
tokio-console --truecolor --no-colors false http://127.0.0.1:6669
```

### After PR

Colors are enabled. 
```bash
COLORTERM=truecolor tokio-console http://127.0.0.1:6669
```
---
Colors are enabled too.
```bash
tokio-console --palette 256 http://127.0.0.1:6669
```
---
Colors are disabled, because `no_colors` is explicitly passed, tho `COLORTERM` is set to `truecolor`.
```bash
COLORTERM=truecolor tokio-console --no-colors http://127.0.0.1:6669
```

### Explanation

I think it's just not really obvious that you need to explicitly pass `--no-colors false` to enable colors, so it should default to false.

### DIFF
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -130,8 +130,8 @@ impl Serialize for RetainFor {
 #[clap(group = ArgGroup::new("colors").conflicts_with("no-colors"))]
 pub struct ViewOptions {
     /// Disable ANSI colors entirely.
-    #[clap(name = "no-colors", long = "no-colors")]
-    no_colors: Option<bool>,
+    #[clap(name = "no-colors", long = "no-colors", takes_value = false)]
+    no_colors: bool,
 
     /// Overrides the terminal's default language.
     #[clap(long = "lang", env = "LANG")]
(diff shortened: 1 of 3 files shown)


tokio-console/src/config.rs
```diff
@@ -377,7 +377,7 @@ impl ViewOptions {
     /// - Checking the `terminfo` database via `tput`
     pub(crate) fn determine_palette(&self) -> Palette {
         // Did the user explicitly disable colors?
-        if self.no_colors.unwrap_or(true) {
```
hawkw, line 380: Alternatively, I think we could have just changed this to `unwrap_or(false)`...but I think your change is better. I don't love that the previous code had the CLI flag take a value, like `--no-colors true`, which seems quite weird. This is definitely better.

hawkw (approved):

PR #352, opened 2022-05-23
feat(console): dynamically size task details field list

hawkw:
## Motivation

Currently, the task details view always gives the list of fields 50% of
the vertical space. This isn't a particularly effective use of space
when a task has only a small number of fields --- we end up wasting a
lot of the vertical real estate that could be used for something nicer.

## Solution

This branch changes the layout of the task details view to give the list
of fields a `Min` layout constraint with the length of the list of
fields plus the border width. This way, when the list of fields is
short, we don't waste a bunch of vertical space on blank lines.

Now, we have a bunch more vertical resolution for drawing a nicer
histogram.
 
## Screenshots

Before:
image

After:
image


hawkw:
Maybe the histogram looks *too* stretched out in the second case and we should give it a percent constraint additionally...would love to hear if anyone has thoughts on this.

hawkw:
@hds i wonder if it's worth resurrecting this after your changes to add additional histograms to the task details view (#409) merge?

hds:
> @hds i wonder if it's worth resurrecting this after your changes to add additional histograms to the task details view (#409) merge?

@hawkw I think that makes sense. The concerns you raised above are less of an issue with 2 histograms - although we'll still need to improve the constraints so that we don't end up with really, really tall histograms.

I'm happy to look into this after I've finished the scheduled time histogram (and the related docs change). I can pull this branch into my own or if you like you can give me permissions on this branch and I'll work on it here.

hawkw:
> I'm happy to look into this after I've finished the scheduled time histogram (and the related docs change). I can pull this branch into my own or if you like you can give me permissions on this branch and I'll work on it here.

Let's just get #409 through first, and then I can rebase onto `main`. :)

PR #358, opened 2022-07-05
refac(console): use a newtype to separate ID types

hawkw:
The console TUI's state model deals with two different types of IDs:

 - the actual `tracing` span IDs for an object (task, resource, or async
   op), which are *not* assigned sequentially, and may be reused over
   the program's lifetime, and
 - sequential "pretty" IDs assigned by the console, which are mapped to
   span IDs when an object is sent to the console. These are assigned
   based on separate counters for each type of object (so there can be
   both a task 1 and a resource 1, for example).

Remote span IDs are mapped to rewritten sequential IDs by the `Id` type,
which stores a map of span IDs to sequential IDs, and generates new
sequential IDs when a new span ID is recorded. The `Ids::id_for` method
takes a span ID and returns the corresponding sequential ID, and this
must be called before looking up or inserting an object by its remote
span ID.

Currently, *all* of these IDs are represented as `u64`s. This is
unfortunate, as it makes it very easy to accidentally introduce bugs
where the wrong kind of ID is used. For example, it's possible to
accidentally look up a task in the map of active tasks based on its span
ID on the remote application, which is likely to return `None` even if
there is a task with that span ID. PR #251 fixed one such ID-confusion
issue (where `WatchDetails` RPCs were performed with local, rewritten
task IDs rather than the remote's span ID for that task). However, it
would be better if we could just *not have* this class of issue.

This branch refactors the console's `state` module to make ID confusion
issues much harder to write. This is accomplished by adding an `Id`
newtype to represent our rewritten sequential IDs. `Ids::id_for` now
still takes a `u64` (as the remote span IDs are represented as `u64`s in
protobuf, so there's nothing we can do), but it returns an `Id` newtype.
Looking up or inserting objects in a state map now takes the `Id`
newtype. This ensures that all lookups are performed with sequential IDs
at the type level, and the only way to get a sequential ID is to ask the
`Ids` map for one.

Additionally, the `Id` type has a type parameter for the type of object
it identifies. This prevents additional issues where we might look up
the ID of a task in the map of resources, or similar.

### DIFF
--- a/tokio-console/src/state/async_ops.rs
+++ b/tokio-console/src/state/async_ops.rs
@@ -1,6 +1,12 @@
 use crate::{
     intern::{self, InternedStr},
-    state::{pb_duration, Attribute, Field, Ids, Metadata, Visibility},
+    state::{
+        id::{Id, Ids},
+        pb_duration,
+        resources::Resource,
+        tasks::Task,
+        Attribute, Field, Metadata, Visibility,
+    },
     view,
 };
 use console_api as proto;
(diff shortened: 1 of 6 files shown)


hawkw:
cc @jswrenn

PR #363, opened 2022-07-21
console-api,console-subscriber,tokio-console: implement stack traces for tasks

jswrenn:
The task detail view of console now shows the tree of extant consequences stemming from that task; i.e., which spans are open beneath the task. The leaves of the task, in a sense, reflect what the task is busy doing at that moment.

I'm opening this as a draft PR because several issues must be resolved before merging; most importantly:
- the spawn-a-new-thread-to-get-the-global-subscriber technique is horrendous
  - I don't know how to solve this more elegantly.
- `tracing-causality` should not be a git dependency
  - its tests passing are blocked on https://github.com/tokio-rs/tracing/pull/2229 getting backported+released to tracing 0.1.x, or me writing an equality shim
- it needs to be rebased on main


console-api/proto/common.proto
```diff
@@ -78,6 +78,8 @@ message Span {
     repeated Field fields = 3;
     // Timestamp for the span.
     google.protobuf.Timestamp at = 4;
+    // An Id that uniquely identifies this span's parent (if any).
+    optional SpanId parent_id = 5;
```
bryangarza: in proto3, every field (that is a Message) is optional by default, so you don't need the `optional` keyword

console-api/proto/consequences.proto
```diff
@@ -0,0 +1,53 @@
...
+}
+
+message NewIndirect {
+    Span cause = 1;
+    Span indirect_consequences = 3;
+}
+
+message CloseDirect {
+    Span span = 1;
+    optional Span direct_cause = 2;
```
bryangarza, line 41: in proto3, every field (that is a Message) is optional by default, so you don't need the `optional` keyword

console-api/proto/consequences.proto
```diff
@@ -0,0 +1,53 @@
...
+}
+
+message CloseIndirect {
+    Span span = 1;
+    repeated Span indirect_causes = 3;
+}
+
+message CloseCyclic {
+    Span span = 1;
+    optional Span direct_cause = 2;
```
bryangarza, line 51: in proto3, every field (that is a Message) is optional by default, so you don't need the `optional` keyword

console-subscriber/src/aggregator/mod.rs
```diff
@@ -368,19 +387,28 @@ impl Aggregator {
...
+            .retain_and_shrink(|id, updates_and_watchers| {
+                if let Some(task_stats) = stats.get(id) {
+                    let details = proto::tasks::TaskDetails {
+                        task_id: Some(id.clone().into()),
+                        now: Some(self.base_time.to_timestamp(Instant::now())),
+                        poll_times_histogram: task_stats.serialize_histogram(),
+                        causality: vec![],
+                    };
+                    updates_and_watchers.retain(|(updates, watch)| {
+                        let updates = crate::consequences::updates_into_proto(updates);
```
bryangarza, line 401: tiny nit: you don't need `crate::` here

console-subscriber/src/aggregator/mod.rs
```diff
@@ -368,19 +387,28 @@ impl Aggregator {
...
+                        let details = proto::tasks::TaskDetails {
+                            causality: updates,
+                            ..details.clone()
+                        };
```
bryangarza, line 405: maybe make `details` mutable so we can just update it and avoid the clone? or flip the order of the code so that we have the updates variable available before constructing `details`

console-subscriber/src/consequences.rs
```diff
@@ -0,0 +1,105 @@
+use console_api as proto;
+use tracing_causality as causality;
+use tracing_subscriber::Registry;
+
+fn into_proto_span(span: causality::Span) -> proto::consequences::Span {
```
bryangarza, line 56: if you import all the proto stuff but alias them to prefix `proto`, like `protoSpan`, it could look a bit cleaner

console-subscriber/src/aggregator/mod.rs
```diff
@@ -320,19 +322,36 @@ impl Aggregator {
             let (tx, rx) = mpsc::channel(buffer);
             let subscription = Watch(tx);
             let now = Some(self.base_time.to_timestamp(Instant::now()));
+
+            let span = id.clone();
+            // XXX(jswrenn): A new thread is spawned solely for the sake of
```
jswrenn, line 327: I'd like to find a better alternative to this.

console-subscriber/src/aggregator/mod.rs
```diff
@@ -320,19 +322,36 @@ impl Aggregator {
...
+            // getting access to the global subscriber.
+            let (trace, updates) = std::thread::spawn(move || {
+                // XXX(jswrenn): Need to gracefully handle the case where the
+                // task span is already closed.
+                consequences::trace(&span, buffer)
+                    .expect("the subscriber isn't a Registry")
+                    .expect("id is already closed")
+            })
+            .join()
+            .unwrap();
```
jswrenn: I'd like to find a better alternative to this.

jswrenn: I think I found an alternative to this! It's just blocked on https://github.com/tokio-rs/tracing/pull/2247.

console-api/proto/consequences.proto
jswrenn: Huh, that's unfortunate, because here it's semantically significant whether a `direct_cause` is provided or not — some `Span`s *don't* have parents.

bryangarza: I think that’s fine, it will be wrapped in Optional in the generated code, so only if the field is included explicitly will it show up as Some(…)

console-api/proto/consequences.proto
```diff
@@ -0,0 +1,53 @@
...
+import "common.proto";
+
+message Causality {
+    oneof update {
+        Extant extant = 1;
+        OpenDirect open_direct = 2;
+        NewIndirect new_indirect = 3;
+        CloseDirect close_direct = 4;
+        CloseIndirect close_indirect = 5;
+        CloseCyclic close_cyclic = 6;
```
hawkw, line 14: let's add comments to these explaining what they mean, please? the terminology here is not immediately obvious to me, especially because it's not part of the terminology used in the `tracing` documentation; it would be good to explain what we mean when we say "direct" and "indirect", for example.

console-api/proto/tasks.proto
```diff
@@ -49,6 +50,9 @@ message TaskDetails {
 
     // HdrHistogram.rs `Histogram` serialized to binary in the V2 format
     optional bytes poll_times_histogram = 3;
+
+    // What the task is currently busy doing.
+    repeated consequences.Causality causality = 4;
```
hawkw, line 55: i don't love calling this field `causality` --- that suggests to me that a task's `causality` is that which _caused_ the task to be spawned. i might call this something like `spans` or `trace` or something, since it's specifically a list of the span trees that the task is in, i believe?

console-subscriber/examples/deep.rs
```diff
@@ -0,0 +1,44 @@
+use tokio::task;
+
+#[tokio::main]
```
hawkw: would be nice to have a doc comment somewhere in here explaining what this example demonstrates.

console-subscriber/examples/deep.rs
```diff
@@ -0,0 +1,44 @@
...
+async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    use tracing_subscriber::prelude::*;
+
+    let subscriber = tracing_subscriber::registry()
+        .with(console_subscriber::spawn())
+        .with(tracing_subscriber::filter::filter_fn(|_| true));
```
hawkw: this could alternatively be `LevelFilter::TRACE` to enable everything explicitly; either is fine by me though.

tokio-console/src/state/mod.rs
```diff
@@ -218,16 +229,41 @@ impl State {
...
-                span_id: id.id,
-                poll_times_histogram: update.poll_times_histogram.and_then(|data| {
-                    hdrhistogram::serialization::Deserializer::new()
-                        .deserialize(&mut Cursor::new(&data))
-                        .ok()
-                }),
+            let span_id = id.id;
+            let task_id = self.tasks_state.ids.id_for(span_id);
+            let task_ref = self.tasks_state.task(task_id).unwrap();
+            let task_ref = task_ref.upgrade().unwrap();
```
hawkw, line 235: i think if we get an update for a task whose ID we've never seen before, we _probably_ shouldn't panic --- it means the server almost certainly did something wrong, but it would probably be better for the console CLI to print some kind of error and keep going, skipping that update, instead of crashing...

tokio-console/src/state/mod.rs
```diff
@@ -218,16 +229,41 @@ impl State {
...
 
-            *self.current_task_details.borrow_mut() = Some(details);
+            if let None = fresh.causality {
+                fresh.causality = Some(tracing_causality::Trace::with_root(
+                    tracing_causality::Span {
+                        id: tracing::Id::from_u64(span_id),
+                        metadata: meta_id,
+                    },
+                ));
+            }
```
hawkw, line 256: style nit, take it or leave it: this is essentially a reimplementation of [`Option::get_or_insert_with`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.get_or_insert_with):
```suggestion
            fresh.causality.get_or_insert_with(|| tracing_causality::Trace::with_root(
                tracing_causality::Span {
                    id: tracing::Id::from_u64(span_id),
                    metadata: meta_id,
                },
            ));
```

tokio-console/src/state/mod.rs
```diff
@@ -218,16 +229,41 @@ impl State {
...
+
+            for causal_update in update.causality {
+                for update in trace::updates_from_pb(causal_update.update.unwrap()) {
+                    if let Some(trace) = fresh.causality {
+                        fresh.causality = trace.apply(update);
+                    }
```
hawkw, line 262: nit, take it or leave it: this could be:
```suggestion
                    fresh.causality = fresh.causality.and_then(|trace| trace.apply(update));
```

tokio-console/src/state/mod.rs
```diff
@@ -264,12 +300,50 @@ impl Metadata {
...
+            line: pb.line,
+            column: pb.column,
+        }
+    }
+}
+
+impl Display for Location {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Some(file) = &self.file {
+            write!(f, "{file}")?;
```
hawkw, line 329: the use of this syntax increases the `tokio-console` crate's minimum Rust dependency, so we'll either have to update the MSRV in Cargo.toml and on CI, or use the old syntax.

tokio-console/src/view/task.rs
```diff
@@ -332,6 +353,21 @@ impl Details {
         text.extend(percentiles);
         text
     }
+
+    /// Render the consequences as an ascii tree.
+    fn make_stack_widget(&self, _styles: &view::Styles, state: &State) -> Text<'static> {
+        // XXX(jswrenn): also implement an ASCII-only rendering mode
```
hawkw, line 359: i think it's fine to just not display the stack in ASCII-only mode, if it's difficult to implement an ASCII-only rendering for this. we already don't display the poll duration histogram in ASCII-only mode...

tokio-console/src/state/mod.rs
```diff
@@ -497,7 +571,7 @@ impl Attribute {
     }
 }
 
-// === impl Ids ===
+// === impl Ids ===a
```
hawkw, line 574: ?
```suggestion
// === impl Ids ===
```

tokio-console/src/view/task.rs
```diff
@@ -341,3 +377,53 @@ fn dur(styles: &view::Styles, dur: std::time::Duration) -> Span<'static> {
     // have to be given a string in order to do layout stuff?
     styles.time_units(format!("{:.prec$?}", dur, prec = DUR_PRECISION))
 }
+
+fn display<W: fmt::Write>(
```
hawkw, line 381: can we maybe give this a more descriptive name? this is specifically for displaying stack elements, right? so maybe something that indicates that?

alternatively, we could make it an inner function of the `make_stack_widget` function, i suppose.

tokio-console/src/view/task.rs
```diff
@@ -332,6 +353,21 @@ impl Details {
...
+    /// Render the consequences as an ascii tree.
+    fn make_stack_widget(&self, _styles: &view::Styles, state: &State) -> Text<'static> {
+        // XXX(jswrenn): also implement an ASCII-only rendering mode
+        let mut buf = String::new();
+        if let Some(causality) = self.causality.as_ref() {
+            let root = causality.root().clone();
+            display(&mut buf, state, causality, &root, true, "").unwrap();
+        } else {
+            buf.push_str(
+                "An unexpected error prevents us from showing consequences for this task.",
```
hawkw, line 366: under what circumstances would this be the case?

tokio-console/src/state/trace.rs
```diff
@@ -0,0 +1,79 @@
...
+                .into_iter()
+                .zip(repeat(cause.clone()))
+                .map(
+                    move |(consequence, cause)| tracing_causality::Update::OpenDirect {
+                        cause: cause,
+                        consequence: span_from_pb(consequence),
+                    },
+                );
+            let indirect_consequences = indirect_consequences.into_iter().zip(repeat(cause)).map(
+                move |(consequence, cause)| tracing_causality::Update::NewIndirect {
```
hawkw, line 33: what is moved into these closures? AFAICT the `move` keyword is unneeded:
```suggestion
                    |(consequence, cause)| tracing_causality::Update::OpenDirect {
                        cause: cause,
                        consequence: span_from_pb(consequence),
                    },
                );
            let indirect_consequences = indirect_consequences.into_iter().zip(repeat(cause)).map(
                |(consequence, cause)| tracing_causality::Update::NewIndirect {
```
i'm assuming this was because `cause` was captured into the closures at one point before the `zip(repeat(cause))` thing?

console-subscriber/src/aggregator/mod.rs
hawkw: this is quite unfortunate, i'd like to figure out a better solution for this. at the very least, we could probably do this a _single_ time when the aggregator task is spawned, and clone the `Dispatch`, so that we don't have to do this every subsequent time we want to get details for a task...

tokio-console/Cargo.toml
```diff
@@ -34,6 +34,7 @@ tonic = { version = "0.7", features = ["transport"] }
 futures = "0.3"
 tui = { version = "0.16.0", default-features = false, features = ["crossterm"] }
 tracing = "0.1"
+tracing-causality = { path = "../../tracing-causality" }
```
hawkw, line 37: can this be changed to a Git dependency rather than a path dependency, at least? i'd really like to be able to actually try out this branch without having to check that out.

console-subscriber/src/lib.rs
```diff
@@ -494,6 +500,17 @@ impl<S> Layer<S> for ConsoleLayer
...
+    fn on_register_dispatch(&self, dispatch: &Dispatch) {
+        let tracer = consequences::Tracer::from_dispatch::<S>(dispatch);
+        if let Err(_) = self.shared.tracer.set(tracer) {
+            // XXX(jswrenn): This layer has seemingly been supplied *multiple*
+            // subscribers. Is that possible? I guess you can call
+            // `Layer::with_subscriber` multiple times; what happens if you
+            // install the resulting `Layered` as a subscriber? Is `Arc`ing a
+            // `Layered<ConsoleLayer, S>` also a recipe for weirdness?
+        }
+    }
```
jswrenn, line 512: The `on_register_dispatch` method defined by https://github.com/tokio-rs/tracing/pull/2269 is implemented here, to construct a `Tracer` (i.e., a type that defines a `trace` method that consumes a span ID and produces a trace of its consequences), which is then made available to the `Aggregator` thread (the part of console_subscriber that actually handles the rpcs for reporting traces).

console-subscriber/src/lib.rs
```diff
@@ -494,6 +500,17 @@ impl<S> Layer<S> for ConsoleLayer
...
     S: Subscriber + for<'a> LookupSpan<'a>,
 {
+    fn on_register_dispatch(&self, dispatch: &Dispatch) {
+        let tracer = consequences::Tracer::from_dispatch::<S>(dispatch);
+        if let Err(_) = self.shared.tracer.set(tracer) {
+            // XXX(jswrenn): This layer has seemingly been supplied *multiple*
+            // subscribers. Is that possible? I guess you can call
+            // `Layer::with_subscriber` multiple times; what happens if you
+            // install the resulting `Layered` as a subscriber? Is `Arc`ing a
+            // `Layered<ConsoleLayer, S>` also a recipe for weirdness?
```
hawkw, line 510: I think a `Layer` implementation for `Arc<L: Layer<_>>` is the only way this could be called twice. Fortunately, [there is no such impl in the current release version](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/trait.Layer.html#foreign-impls)

hawkw:
@jswrenn i believe the requisite upstream changes in `tracing` have all been released now; any interest in picking this back up?

PR #364, opened 2022-08-02
chore: update Tonic and Prost dependencies

hawkw:
This branch updates the console crates' dependency on `tonic` to v0.8.0
and `prost` to 0.11.0.

In addition, I've added a [`cargo xtask`][xtask] command for manually
regenerating the generated protobuf bindings.

This is necessary as the current approach, regenerating the bindings in
an integration test, does not work when the protos fail to compile
(which they do after the Tonic update). Since running the crate's tests
requires compiling the crate, if the proto bindings don't compile, we
can't re-run the test.

[xtask]: https://github.com/matklad/cargo-xtask

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -59,6 +59,5 @@ fn gen_proto() -> Result<(), Box<dyn std::error::Error>> {
         .compile(iface_files, &[proto_dir])?;
 
     eprintln!("protos regenerated!");
-
     Ok(())
 }
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -1,5 +1,6 @@
 use clap::Parser;
-use std::path::PathBuf;
+use std::{fs, path::PathBuf};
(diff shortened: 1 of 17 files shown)


hawkw:
cc @LucioFranco what do you think of the current approach to generating the bindings here?

xtask/src/main.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    tonic_build::configure()
+        .build_client(true)
+        .build_server(true)
+        .protoc_arg("--experimental_allow_proto3_optional")
+        .out_dir(format!("{}", out_dir.display()))
+        .compile(iface_files, &[proto_dir])?;
+
+    eprintln!("protos regenerated!");
+
```
LucioFranco:
```suggestion
```

LucioFranco (commented):
Does CI check that the generated protofiles were updated?

hawkw:
> Does CI check that the generated protofiles were updated?

yes, it runs `cargo test`, which does that

PR #374, opened 2022-09-03
chore: fix console-subscriber on tokio 1.21.0

Noah-Kennedy:
Fixes #373.

Due to a change in the unstable task builder APIs, this no longer compiles with the latest version of Tokio.

Fortunately, it's a simple fix.

PR #375, opened 2022-09-04
chore: fix clippy errors

Noah-Kennedy:
Needed for #374.

### DIFF
--- a/console-api/src/async_ops.rs
+++ b/console-api/src/async_ops.rs
@@ -1,1 +1,3 @@
+#![allow(warnings)]
+
 include!("generated/rs.tokio.console.async_ops.rs");
--- a/console-api/src/common.rs
+++ b/console-api/src/common.rs
@@ -1,7 +1,12 @@
 use std::fmt;
 use std::hash::{Hash, Hasher};
 
-include!("generated/rs.tokio.console.common.rs");
+pub use generated::*;
+
+mod generated {
+    #![allow(warnings)]
+    include!("generated/rs.tokio.console.common.rs");
+}
 
 impl From<tracing_core::Level> for metadata::Level {
     fn from(level: tracing_core::Level) -> Self {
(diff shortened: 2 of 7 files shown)


Noah-Kennedy:
Oh these issues are with generated files!

Noah-Kennedy:
@hawkw should I have clippy ignore the generated files?

hawkw:
> @hawkw should I have clippy ignore the generated files?

yeah, probably...that's annoying. :|

Noah-Kennedy:
Done.

Noah-Kennedy:
Alright, it looks to be passing CI. Is squashing enabled for this repo?

console-api/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![doc = include_str!("../README.md")]
+#![allow(warnings)]
```
hawkw: hmm, could we, instead, add the attribute only to the modules that are generated by `prost`? the `common` module also contains some handwritten code: https://github.com/tokio-rs/console/blob/ff6d048e0464f55fc9d8202b05c7762bc99f9924/console-api/src/common.rs#L5-L240 and it would be nice if clippy could lint that. we might have to put the generated code in a submodule of `common`, like this:

```rust
// in `common.rs`

mod generated {
    #![allow(warnings)]
    include!("generated/rs.tokio.console.common.rs");
}

pub use self::generated::*;

// ...
```
for the other modules, we can just put the `#[allow(warnings)]` attribute on the declaration of those modules.

hawkw (approved):
i'm fine with merging this as-is in order to fix CI, but if you don't mind, it might be nice to change the blanket `allow` for the whole `console-api` crate so that it doesn't include the human-written code in `console_api::common`.

console-api/src/lib.rs
Noah-Kennedy: Will do!

Noah-Kennedy: Done.

Noah-Kennedy:
squashed

Noah-Kennedy:
@hawkw I don't have merge permissions for this repo, so you'll have to do it.

Noah-Kennedy:
Thanks!

PR #377, opened 2022-09-06
fix(console): fix ascii-only flipped input

### LINKED ISSUE #372: --ascii-only is flipped

tokio-console

### What is the issue?

See repro

`tokio-console --ascii-only true` shows emojis
`tokio-console --ascii-only false` shows ascii

### Logs, error output, etc

_No response_

### Versions

0.1.7

### Possible solution

`clap` crate has a `default_value` feature, makes the code cleaner and documents the defaults in the help message

### Additional context

_No response_

no

michealkeines:
Fixes: #372

`--ascii-only true` shows ascii
`--ascii-only false` shows emojis

if `--ascii-only` is not passed, default value will be true, shows emojis

### DIFF
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -411,7 +411,7 @@ fn default_log_directory() -> PathBuf {
 
 impl ViewOptions {
     pub fn is_utf8(&self) -> bool {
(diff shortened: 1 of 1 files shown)


hawkw (approved):

PR #379, opened 2022-09-21
Documented cargo-run in README no longer works, as now there are two binaries

dcaba:
Suggesting to declare `tokio-console` as the [default-run](https://doc.rust-lang.org/cargo/reference/manifest.html#the-default-run-field), as the current `cargo run` command in the README fails with:

```bash
$ cargo run                         
error: `cargo run` could not determine which binary to run. Use the `--bin` option t
o specify a binary, or the `default-run` manifest key.                              
available binaries: tokio-console, xtask         
```

Alternatively, we can just update the README line with `cargo run --bin tokio-console`


hawkw (approved):
Thanks, today I learnt about `default-run`, that's great! Thanks for the fix!

PR #383, opened 2022-10-07
fix(tokio-console): Make `retain_for` default to 6s if not specfied

### LINKED ISSUE #382: `--retain-for` default value not used

tokio-console

### What is the issue?

The help text for the `--retain-for` option suggests that the default value is 6 seconds, however this seems to be ignored. Done tasks seem to hang around forever in the UI.

Explicitly specifying a duration with `--retain-for` does work though.

Run `tokio-console` on a project, and observe "done" tasks staying around in the UI.

### Logs, error output, etc

Output of `tokio-console --help` (notice the last line):

```
        --retain-for <RETAIN_FOR>
            How long to continue displaying completed tasks and dropped resources after they have
            been closed.

            This accepts either a duration, parsed as a combination of time spans (such as `5days
            2min 2s`), or `none` to disable removing completed tasks and dropped resources.

            Each time span is an integer number followed by a suffix. Supported suffixes are:

            * `nsec`, `ns` -- nanoseconds

            * `usec`, `us` -- microseconds

            * `msec`, `ms` -- milliseconds

            * `seconds`, `second`, `sec`, `s`

            * `minutes`, `minute`, `min`, `m`

            * `hours`, `hour`, `hr`, `h`

            * `days`, `day`, `d`

            * `weeks`, `week`, `w`

            * `months`, `month`, `M` -- defined as 30.44 days

            * `years`, `year`, `y` -- defined as 365.25 days

            [default: 6s]
```

### Versions

```
❯ tokio-console --version
tokio-console 0.1.7
```

### Possible solution

_No response_

### Additional context

_No response_

maybe

abusch:
closes #382

### DIFF
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -326,9 +326,7 @@ impl Config {
     }
 
     pub(crate) fn retain_for(&self) -> Option<Duration> {
-        self.retain_for
-            .unwrap_or_default()
-            .0
+        self.retain_for.unwrap_or_default().0
     }
 
     pub(crate) fn target_addr(&self) -> Uri {
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -121,6 +121,12 @@ pub enum OptionalCmd {
 #[derive(Debug, Clone, Copy, Deserialize)]
 struct RetainFor(Option<Duration>);
 
+impl Default for RetainFor {
+    fn default() -> Self {
+        Self(Some(Duration::from_secs(6)))
+    }
+}
+
 impl fmt::Display for RetainFor {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self.0 {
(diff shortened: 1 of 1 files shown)


hawkw (approved):

It occurs to me that, rather than using `unwrap_or_default`, we could alternatively change the  `clap` arguments struct so that, rather than an `Option<RetainFor>`, the args struct has a `RetainFor` with the `#[clap(default_value_t = RetainFor::default())]` attribute. That way, the default in the help text would be determined from the actual value. On the other hand, I think that may not play as nicely with the config file overriding, as we would need to determine whether a 6s retention from the CLI args was actually provided explicitly, or is just the default value. So, I think the current approach may actually be the better one...

It might be worth thinking about how to make usage of `clap` default values play nicer with the config file code in a separate branch, though...

tokio-console/src/config.rs
```diff
@@ -320,7 +326,9 @@ impl Config {
     }
 
     pub(crate) fn retain_for(&self) -> Option<Duration> {
-        self.retain_for.as_ref().and_then(|value| value.0)
+        self.retain_for
+            .unwrap_or_default()
+            .0
```
hawkw: hmm, looks like `rustfmt` would prefer to see
```suggestion
        self.retain_for.unwrap_or_default().0
```
here: https://github.com/tokio-rs/console/actions/runs/3202161732/jobs/5240217564

PR #387, opened 2022-11-16
enable the view switching keystrokes on the details views

arifd:
enables the view switching keystrokes on the details views

### DIFF
--- a/tokio-console/src/view/mod.rs
+++ b/tokio-console/src/view/mod.rs
@@ -101,10 +101,12 @@ impl View {
             self.state = TasksList;
             return update_kind;
         }
+        
         if matches!(event, key!(Char('r'))) {
             self.state = ResourcesList;
             return update_kind;
         }
+        
         match self.state {
             TasksList => {
                 // The enter key changes views, so handle here since we can
--- a/tokio-console/src/view/mod.rs
+++ b/tokio-console/src/view/mod.rs
@@ -96,6 +96,15 @@ impl View {
     pub(crate) fn update_input(&mut self, event: input::Event, state: &State) -> UpdateKind {
         use ViewState::*;
         let mut update_kind = UpdateKind::Other;
+
+        if matches!(event, key!(Char('t'))) {
+            self.state = TasksList;
+            return update_kind;
+        }
+        if matches!(event, key!(Char('r'))) {
+            self.state = ResourcesList;
+            return update_kind;
+        }
         match self.state {
             TasksList => {
                 // The enter key changes views, so handle here since we can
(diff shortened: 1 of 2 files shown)


tokio-console/src/view/mod.rs
```diff
@@ -144,6 +144,12 @@ impl View {
                         self.state = ResourcesList;
                         update_kind = UpdateKind::Other;
                     }
+                    key!(Char('r')) => {
+                        self.state = ResourcesList;
+                    }
+                    key!(Char('t')) => {
+                        self.state = TasksList;
+                    }
```
hawkw: take it or leave it: it occurs to me that we _could_ potentially reduce duplication a bit by pulling this code out to _before_ the match here: https://github.com/tokio-rs/console/blob/5621004bf5fa620f945f559daea3c9aa2d4cedab/tokio-console/src/view/mod.rs#L99 so that we have an `if` that handles the `r` and `t` keys regardless of what view we're in, rather than duplicating it for each potential view.

currently, these keys are handled inside the match because the Tasks view doesn't switch back to itself when the `t` key is pressed, and the Resources view similarly skips the `r` key. however, this probably isn't necessary --- changing `self.state` won't result in redrawing the view at a different position or anything.

so, we could probably simplify this code a bit by handling the `r` and `t` keys outside of the `match`, if you think that would be nicer? not a blocker though.

hawkw (approved):
this looks good to me, thanks!

i had one suggestion on how we could potentially make this code a bit simpler, but it's up to you whether you want to make that change or not.

tokio-console/src/view/mod.rs
arifd: Ah, do you mean, if `r` or `t`; do the right thing; else do the match on line 99?

hawkw: yeah, the pseudocode would be 
```rust
if 't' {
    self.state = TasksList;
} else if 'r' {
    self.state = ResourcesList;
} else {
    match self.state {
        // ...
   }
}
```

or, something like:
```rust
if 't' {
   self.state = TasksList;
   return;
}

if 'r' {
   // ...
}

// ...
```

tokio-console/src/view/mod.rs
```diff
@@ -96,6 +96,15 @@ impl View {
...
         let mut update_kind = UpdateKind::Other;
+
+        if matches!(event, key!(Char('t'))) {
+            self.state = TasksList;
+            return update_kind;
+        }
+        if matches!(event, key!(Char('r'))) {
+            self.state = ResourcesList;
+            return update_kind;
+        }
```
hawkw: very minor style nit: i would add a line of whitespace here:
```suggestion
        }
        
        if matches!(event, key!(Char('r'))) {
            self.state = ResourcesList;
            return update_kind;
        }
        
```

hawkw (approved):
looks great, thank you!

PR #388, opened 2022-12-04
feat(console): add support for Unix domain sockets

### LINKED ISSUE #296: Currently, the console only supports TCP, is there a plan to support Unix Domain Socket?

benesch:
Add support for console connections that use Unix domain sockets rather than TCP.

Fix #296.

### DIFF
--- a/console-subscriber/examples/uds.rs
+++ b/console-subscriber/examples/uds.rs
@@ -1,0 +1,40 @@
+//! Demonstrates serving the console API over a [Unix domain socket] (UDS)
+//! connection, rather than over TCP.
+//!
+//! Note that this example only works on Unix operating systems that
+//! support UDS, such as Linux, BSDs, and macOS.
+//!
+//! [Unix domain socket]: https://en.wikipedia.org/wiki/Unix_domain_socket
+
+#[cfg(unix)]
+use {
+    std::time::Duration,
+    tokio::{fs, task, time},
+    tracing::info,
+};
+
+#[cfg(unix)]
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    let cwd = fs::canonicalize(".").await?;
+    let addr = cwd.join("console-server");
+    console_subscriber::ConsoleLayer::builder()
+        .server_addr(&*addr)
+        .init();
+    info!(
+        "listening for console connections at file://localhost{}",
+        addr.display()
+    );
+    task::Builder::default()
+        .name("sleepy")
+        .spawn(async move { time::sleep(Duration::from_secs(90)).await })
+        .unwrap()
+        .await?;
+
+    Ok(())
+}
+
+#[cfg(not(unix))]
+fn main() {
+    panic!("only supported on Unix platforms")
+}
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -1,6 +1,8 @@
 use super::{ConsoleLayer, Server};
+#[cfg(unix)]
+use std::path::Path;
 use std::{
-    net::{SocketAddr, ToSocketAddrs},
+    net::{IpAddr, SocketAddr, SocketAddrV4, SocketAddrV6, ToSocketAddrs},
     path::PathBuf,
     thread,
     time::Duration,
@@ -32,7 +34,7 @@ pub struct Builder {
     pub(crate) retention: Duration,
 
     /// The address on which to serve the RPC server.
-    pub(super) server_addr: SocketAddr,
+    pub(super) server_addr: ServerAddr,
 
     /// If and where to save a recording of the events.
     pub(super) recording_path: Option<PathBuf>,
@@ -58,7 +60,7 @@ impl Default for Builder {
             publish_interval: ConsoleLayer::DEFAULT_PUBLISH_INTERVAL,
             retention: ConsoleLayer::DEFAULT_RETENTION,
             poll_duration_max: ConsoleLayer::DEFAULT_POLL_DURATION_MAX,
-            server_addr: SocketAddr::new(Server::DEFAULT_IP, Server::DEFAULT_PORT),
+            server_addr: ServerAddr::Tcp(SocketAddr::new(Server::DEFAULT_IP, Server::DEFAULT_PORT)),
             recording_path: None,
             filter_env_var: "RUST_LOG".to_string(),
             self_trace: false,
@@ -137,8 +139,34 @@ impl Builder {
     /// before falling back on constructing a socket address from those
     /// defaults.
     ///
+    /// The socket address can be either a TCP socket address or a
+    /// [Unix domain socket] (UDS) address. Unix domain sockets are only
+    /// supported on Unix-compatible operating systems, such as Linux, BSDs,
+    /// and macOS.
+    ///
+    /// Each call to this method will overwrite the previously set value.
+    ///
+    /// # Examples
+    ///
+    /// Connect to the TCP address `localhost:1234`:
+    ///
+    /// ```
+    /// # use console_subscriber::Builder;
+    /// use std::net::Ipv4Addr;
+    /// let builder = Builder::default().server_addr((Ipv4Addr::LOCALHOST, 1234));
+    /// ```
+    ///
+    /// Connect to the UDS address `/tmp/tokio-console`:
+    ///
+    /// ```
+    /// # use console_subscriber::Builder;
+    /// use std::path::Path;
+    /// let builder = Builder::default().server_addr(Path::new("/tmp/tokio-console"));
+    /// ```
+    ///
     /// [environment variable]: `Builder::with_default_env`
-    pub fn server_addr(self, server_addr: impl Into<SocketAddr>) -> Self {
+    /// [Unix domain socket]: https://en.wikipedia.org/wiki/Unix_domain_socket
+    pub fn server_addr(self, server_addr: impl Into<ServerAddr>) -> Self {
         Self {
             server_addr: server_addr.into(),
             ..self
@@ -231,11 +259,14 @@ impl Builder {
         }
 
         if let Ok(bind) = std::env::var("TOKIO_CONSOLE_BIND") {
-            self.server_addr = bind
-                .to_socket_addrs()
-                .expect("TOKIO_CONSOLE_BIND must be formatted as HOST:PORT, such as localhost:4321")
-                .next()
-                .expect("tokio console could not resolve TOKIO_CONSOLE_BIND");
+            self.server_addr = ServerAddr::Tcp(
+                bind.to_socket_addrs()
+                    .expect(
+                        "TOKIO_CONSOLE_BIND must be formatted as HOST:PORT, such as localhost:4321",
+                    )
+                    .next()
+                    .expect("tokio console could not resolve TOKIO_CONSOLE_BIND"),
+            );
         }
 
         if let Some(interval) = duration_from_env("TOKIO_CONSOLE_PUBLISH_INTERVAL") {
@@ -453,6 +484,66 @@ impl Builder {
             .expect("console subscriber could not spawn thread");
 
         layer
+    }
+}
+
+/// Specifies the address on which a [`Server`] should listen.
+///
+/// This type is passed as an argument to the [`Builder::server_addr`]
+/// method, and may be either a TCP socket address, or a [Unix domain socket]
+/// (UDS) address. Unix domain sockets are only supported on Unix-compatible
+/// operating systems, such as Linux, BSDs, and macOS.
+///
+/// [`Server`]: crate::Server
+/// [Unix domain socket]: https://en.wikipedia.org/wiki/Unix_domain_socket
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub enum ServerAddr {
+    /// A TCP address.
+    Tcp(SocketAddr),
+    /// A Unix socket address.
+    #[cfg(unix)]
+    Unix(PathBuf),
+}
+
+impl From<SocketAddr> for ServerAddr {
+    fn from(addr: SocketAddr) -> ServerAddr {
+        ServerAddr::Tcp(addr)
+    }
+}
+
+impl From<SocketAddrV4> for ServerAddr {
+    fn from(addr: SocketAddrV4) -> ServerAddr {
+        ServerAddr::Tcp(addr.into())
+    }
+}
+
+impl From<SocketAddrV6> for ServerAddr {
+    fn from(addr: SocketAddrV6) -> ServerAddr {
+        ServerAddr::Tcp(addr.into())
+    }
+}
+
+impl<I> From<(I, u16)> for ServerAddr
+where
+    I: Into<IpAddr>,
+{
+    fn from(pieces: (I, u16)) -> ServerAddr {
+        ServerAddr::Tcp(pieces.into())
+    }
+}
+
+#[cfg(unix)]
+impl From<PathBuf> for ServerAddr {
+    fn from(path: PathBuf) -> ServerAddr {
+        ServerAddr::Unix(path)
+    }
+}
+
+#[cfg(unix)]
+impl<'a> From<&'a Path> for ServerAddr {
+    fn from(path: &'a Path) -> ServerAddr {
+        ServerAddr::Unix(path.to_path_buf())
     }
 }
 
--- a/console-subscriber/src/builder.rs
+++ b/console-subscriber/src/builder.rs
@@ -160,7 +160,11 @@ impl Builder {
     ///
     /// ```
     /// # use console_subscriber::Builder;
+    /// # #[cfg(unix)]
     /// use std::path::Path;
+    ///
+    /// // Unix domain sockets are only available on Unix-compatible operating systems.
+    /// #[cfg(unix)]
     /// let builder = Builder::default().server_addr(Path::new("/tmp/tokio-console"));
     /// ```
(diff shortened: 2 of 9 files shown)


console-subscriber/src/builder.rs
```diff
@@ -231,11 +231,14 @@ impl Builder {
...
-                .next()
-                .expect("tokio console could not resolve TOKIO_CONSOLE_BIND");
+            self.server_addr = ServerAddr::Tcp(
+                bind.to_socket_addrs()
+                    .expect(
+                        "TOKIO_CONSOLE_BIND must be formatted as HOST:PORT, such as localhost:4321",
+                    )
+                    .next()
+                    .expect("tokio console could not resolve TOKIO_CONSOLE_BIND"),
+            );
```
benesch, line 273: @hawkw do you think this should support UDS addresses too? E.g., by detecting the presence of a leading /, or the absence of a `:port`?

console-subscriber/src/builder.rs
```diff
@@ -138,7 +138,7 @@ impl Builder {
     /// defaults.
     ///
     /// [environment variable]: `Builder::with_default_env`
-    pub fn server_addr(self, server_addr: impl Into<SocketAddr>) -> Self {
+    pub fn server_addr(self, server_addr: impl Into<ServerAddr>) -> Self {
```
benesch, line 173: Open to other API design here. E.g., two separate, mutually-exclusive methods, one for TCP and one for Unix sockets. I went with the new `ServerAddr` type to avoid breaking existing users.

hawkw: Hmm, this approach seems fine to me if we want to make UDS and TCP mutually exclusive. 

However, it occurs to me that we _might_ want to add a separate method for setting a UDS server address, but _not_ making them mutually exclusive. That way, we would allow the user to configure serving the console API on both a TCP port _and_ a Unix socket, if desired. Do you think this is something people will want to be able to do?

If we did have separate methods, it might be worth adding a new method `tcp_server_addr` (or something), and deprecating the existing `server_addr` in favor of that, to make it clearer when called what type of socket is being configured.

On the other hand, the current approach is probably a nicer API surface if we don't want to allow serving over both TCP and UDS...

console-subscriber/src/builder.rs
hawkw: Hmm, my preference is to have this env var only configure a TCP socket, because it seems unfortunate to have a potential typo in the environment variable silently switch from TCP to UDS when that might not actually be desired.

Instead, I think we might want to consider adding a separate environment variable to configure serving over Unix sockets. It would be fine to add that env var in this branch, or in a follow-up PR.

console-subscriber/src/builder.rs
```diff
@@ -456,6 +459,56 @@ impl Builder {
     }
 }
 
+/// Specifies the address on which a [`Server`] should listen.
+///
+/// [`Server`]: crate::Server
+#[derive(Clone, Debug)]
```
hawkw, line 503: It might be desirable to mark this as 
```suggestion
#[derive(Clone, Debug)]
#[non_exhaustive]
```
on the off chance that we add more variants without breaking existing code. Although...it doesn't seem terribly likely that we will (unless we want to support Windows named pipes, or something?)

console-subscriber/src/lib.rs
```diff
@@ -5,15 +5,19 @@ use serde::Serialize;
...
     },
     time::{Duration, Instant},
 };
 use thread_local::ThreadLocal;
-use tokio::sync::{mpsc, oneshot};
+use tokio::{
+    net::UnixListener,
+    sync::{mpsc, oneshot},
+};
+use tokio_stream::wrappers::UnixListenerStream;
```
hawkw: i believe the `UnixListener` imports won't compile on operating systems other than Unix (e.g., on Windows). we probably need to add a `cfg` attribute so that this code builds on Windows.
```suggestion
use tokio::sync::{mpsc, oneshot};
#[cfg(unix)]
use tokio::net::UnixListener;
#[cfg(unix)]
use tokio_stream::wrappers::UnixListenerStream;
```

console-subscriber/src/builder.rs
```diff
@@ -456,6 +459,56 @@ impl Builder {
...
+pub enum ServerAddr {
+    /// A TCP address.
+    Tcp(SocketAddr),
+    /// A Unix socket address.
+    Unix(PathBuf),
```
hawkw: This should probably have `#[cfg(unix)]` so that the API is only available when building for Unix.
```suggestion
    /// A Unix socket address.
    #[cfg(unix)]
    Unix(PathBuf),
```

console-subscriber/src/builder.rs
```diff
@@ -456,6 +459,56 @@ impl Builder {
...
+    fn from(path: PathBuf) -> ServerAddr {
+        ServerAddr::Unix(path)
+    }
+}
+
+impl<'a> From<&'a Path> for ServerAddr {
+    fn from(path: &'a Path) -> ServerAddr {
+        ServerAddr::Unix(path.to_path_buf())
+    }
+}
```
hawkw: similarly, these probably need to be `cfg`-flagged:
```suggestion
#[cfg(unix)]
impl From<PathBuf> for ServerAddr {
    fn from(path: PathBuf) -> ServerAddr {
        ServerAddr::Unix(path)
    }
}

#[cfg(unix)]
impl<'a> From<&'a Path> for ServerAddr {
    fn from(path: &'a Path) -> ServerAddr {
        ServerAddr::Unix(path.to_path_buf())
    }
}
```

console-subscriber/src/lib.rs
```diff
@@ -945,13 +950,21 @@ impl Server {
...
+        let res = match addr {
+            ServerAddr::Tcp(addr) => {
+                let serve = router.serve(addr);
+                spawn_named(serve, "console::serve").await
+            }
+            ServerAddr::Unix(path) => {
+                let incoming = UnixListener::bind(path)?;
+                let serve = router.serve_with_incoming(UnixListenerStream::new(incoming));
+                spawn_named(serve, "console::serve").await
+            }
```
hawkw, line 966: this will need to be `cfg`-flagged so that it's only available on Unix platforms
```suggestion
            #[cfg(unix)]
            ServerAddr::Unix(path) => {
                let incoming = UnixListener::bind(path)?;
                let serve = router.serve_with_incoming(UnixListenerStream::new(incoming));
                spawn_named(serve, "console::serve").await
            }
```

tokio-console/src/conn.rs
```diff
@@ -5,7 +5,11 @@ use console_api::instrument::{
 use console_api::tasks::TaskDetails;
 use futures::stream::StreamExt;
 use std::{error::Error, pin::Pin, time::Duration};
-use tonic::{transport::Channel, transport::Uri, Streaming};
+use tokio::net::UnixStream;
```
hawkw, line 9: this import will need to be `cfg`-flagged:
```suggestion
#[cfg(unix)]
use tokio::net::UnixStream;
```

tokio-console/src/conn.rs
```diff
@@ -78,7 +82,23 @@ impl Connection {
...
+                    None | Some("file") => {
+                        // Dummy endpoint is ignored by the connector.
+                        let endpoint = Endpoint::from_static("http://localhost");
+                        let path = self.target.path().to_owned();
+                        endpoint
+                            .connect_with_connector(tower::service_fn(move |_| {
+                                UnixStream::connect(path.clone())
+                            }))
+                            .await?
+                    }
```
hawkw: this match arm will need to be `#[cfg(unix)]`, because it uses the Unix-only `UnixStream` type:
```suggestion
                    #[cfg(unix)]
                    None | Some("file") => {
                        // Dummy endpoint is ignored by the connector.
                        let endpoint = Endpoint::from_static("http://localhost");
                        let path = self.target.path().to_owned();
                        endpoint
                            .connect_with_connector(tower::service_fn(move |_| {
                                UnixStream::connect(path.clone())
                            }))
                            .await?
                    }
```

also, we may want to add a separate `#[cfg(not(unix))]` match arm that prints an error message when trying to connect to a Unix domain socket on non-unix platforms?

tokio-console/src/conn.rs
```diff
@@ -78,7 +82,23 @@ impl Connection {
                 tokio::time::sleep(backoff).await;
             }
             let try_connect = async {
-                let mut client = InstrumentClient::connect(self.target.clone()).await?;
+                let channel = match self.target.scheme_str() {
+                    None | Some("file") => {
```
hawkw: i'm not sure how i feel about treating URLs without a scheme as UDS...is that the generally accepted convention?

hawkw (changes requested):
Thanks for working on this! This branch looks like a great start!

 I had a few suggestions, most of which I also elaborated on in inline comments:
1. Unix-only APIs need `#[cfg(unix)]` attributes so that the console still compiles on Windows.
2. We should determine whether we want to allow serving the API on both a Unix domain socket _and_ a TCP socket simultaneously, which might be desirable in some cases.
3. We probably want a separate env-var-based configuration for serving on UDS, if we even want to allow configuring that with the environment variable API.
4. Let's update this comment (which generates the CLI help text for `tokio-console`): https://github.com/tokio-rs/console/blob/c7ce40fcdc46c9a8f74c2442b3d87452f5515e6c/tokio-console/src/config.rs#L25-L29 so that it also states that the URL can be a Unix Domain Socket if it starts with `file://`, on unix platforms.

console-subscriber/src/builder.rs
benesch: > Do you think this is something people will want to be able to do?

I think it's fairly unlikely, to be honest! I think it's just as likely to want to serve on multiple TCP addresses as it is to want to serve on exactly one TCP and exactly one UDP address. For example, if you've got two IPs attached to your machine, and you want to serve on both IPs, but at different ports.

If that ever becomes important, I think there's a pretty nice API to be had:

```rust
/// Sets multiple addresses on which to serve the RPC server.
///
/// Overrides any address previously installed by `Self::server_addr`.
fn server_addrs<I, A>(self, server_addrs: I) -> Self
where
    I: IntoIterator<Item = A>,
    A: Into<ServerAddr>,
```

console-subscriber/src/builder.rs
benesch: > Instead, I think we might want to consider adding a separate environment variable to configure serving over Unix sockets. It would be fine to add that env var in this branch, or in a follow-up PR.

I'll leave that to a different PR, then!

console-subscriber/src/builder.rs
benesch: Good idea. Done.

console-subscriber/src/builder.rs
benesch: Done.

console-subscriber/src/builder.rs
benesch: Done.

console-subscriber/src/lib.rs
benesch: Done.

console-subscriber/src/lib.rs
benesch: Done.

tokio-console/src/conn.rs
benesch: Done.

tokio-console/src/conn.rs
benesch: I don't think there _is_ a generally accepted convention. Some other software uses `unix:` as the scheme, as in `unix://path/to/socket`. For now I just removed the treatment of `None` as UDS; let me know if you'd prefer some other scheme instead.

benesch (commented):
Thanks for the comments! I've addressed everything, I think!

benesch:
@hawkw anything you're waiting on me on?

hawkw:
> @hawkw anything you're waiting on me on?

Nope, sorry, I just hadn't gotten the chance to give this a review yet. I'll try to do one today!

benesch:
No problem! Just wanted to make sure I hadn't missed something. Feel free to just push directly at the branch (or open a new PR, whatever) if easier.

console-subscriber/examples/uds.rs
```diff
@@ -0,0 +1,23 @@
+use std::time::Duration;
```
hawkw: what does this example do if it's run on a non-Unix OS? if the `ConsoleLayer` builder emits a nice error here, that's sufficient, but it's definitely important that something in the example makes it clear that this won't work on Windows etc.

in any case, it might be nice to add a comment here like
```suggestion
//! Demonstrates serving the console API over a [Unix Domain Socket] (UDS)
//! connection, rather than over TCP.
//!
//! Note that this example only works on Unix operating systems that 
//! support UDS, such as Linux, BSDs, and macOS.
//!
//! [Unix Domain Socket]: https://en.wikipedia.org/wiki/Unix_domain_socket
use std::time::Duration;
```

console-subscriber/src/builder.rs
hawkw: alternatively, if we wanted to support this, we could just make `server_addr` called multiple times append addresses to a list.

In the meantime, though, I think we should update the documentation for `server_addr` to make sure that it's very clear that multiple calls will overwrite any previously set values.

console-subscriber/src/builder.rs
```diff
@@ -138,7 +140,7 @@ impl Builder {
     /// defaults.
```
hawkw, line 140: this documentation should be updated to state that the provided address can be a TCP address, or (on Unix systems) a UDS address. It would be nice to show examples of both usages in this method's documentation.

console-subscriber/src/builder.rs
```diff
@@ -456,6 +461,60 @@ impl Builder {
     }
 }
 
+/// Specifies the address on which a [`Server`] should listen.
+///
```
hawkw, line 495: This type's documentation should probably include something like:
```suggestion
///
/// This type is passed as an argument to the [`Builder::server_addr`]
/// method, and may be either a TCP socket address, or a [Unix Domain Socket]
/// (UDS) address. Unix Domain Sockets are only supported on Unix-compatible
/// operating systems, such as Linux, BSDs, and macOS.
///
/// [Unix Domain Socket]: https://en.wikipedia.org/wiki/Unix_domain_socket 
```

hawkw (approved):
Overall, this looks good to me! I had some last suggestions for improving the documentation. Once those are addressed, though, I'll be very happy to merge this!

hawkw:
It looks like, because additional documentation was added to the CLI help text, the documentation needs to be regenerated (which is why CI is failing): https://github.com/tokio-rs/console/actions/runs/3715257647/jobs/6300201144#step:8:150

console-subscriber/src/builder.rs
benesch: Done.

console-subscriber/src/builder.rs
benesch: Makes sense! Done.

console-subscriber/src/builder.rs
benesch: Done.

console-subscriber/examples/uds.rs
benesch: Added the comment! I don't have a Windows machine so I can't easily test what happens if you try this on Windows.

benesch (commented):
Awesome, thanks for the second pass! Should be ready to go.

benesch:
Sorry to be a bother, @hawkw, but could you trigger a CI run?

hawkw:
> Sorry to be a bother, @hawkw, but could you trigger a CI run?

Yup, sorry, I hadn't realized you pushed more commits. Thanks for the reminder!

hawkw (approved):
This looks great! Thanks again for working on it!

I'm going to merge this as soon as CI passes. :)

benesch:
Thank you! I fixed the UDS example not compiling on Windows with some conditional compilation. Didn't see a better way. There's no equivalent of `required-features` for an example (e.g., `required-platform`) in Cargo.toml.

Also looks like there's a clippy failure that's unrelated to this PR. Want me to push up a fix in this PR, or would you rather address separately?

benesch:
> Also looks like there's a clippy failure that's unrelated to this PR. Want me to push up a fix in this PR, or would you rather address separately?

Gentle ping on this question, @hawkw!

console-subscriber/src/builder.rs
```diff
@@ -137,8 +139,34 @@ impl Builder {
...
+    /// let builder = Builder::default().server_addr((Ipv4Addr::LOCALHOST, 1234));
+    /// ```
+    ///
+    /// Connect to the UDS address `/tmp/tokio-console`:
+    ///
+    /// ```
+    /// # use console_subscriber::Builder;
+    /// use std::path::Path;
+    /// let builder = Builder::default().server_addr(Path::new("/tmp/tokio-console"));
+    /// ```
```
hawkw: hmm, this doctest appears to be failing on Windows: https://github.com/tokio-rs/console/actions/runs/4545469942/jobs/8012835749?pr=388

let's maybe do something like this:
```suggestion
    /// ```
    /// # use console_subscriber::Builder;
    /// # #[cfg(unix)]
    /// use std::path::Path;
    ///
    /// // Unix domain sockets are only available on Unix-compatible operating systems.
    /// #[cfg(unix)]
    /// let builder = Builder::default().server_addr(Path::new("/tmp/tokio-console"));
    /// ```
```

benesch:
Thanks, Eliza! Excited to have this merged.

PR #391, opened 2022-12-24
Fix off-by-one indexing for `callsites`

danielhenrymantilla:
Screenshot 2022-12-24 at 12 56 55

🙃


hawkw (approved):
Thank you, good catch! My bad, lol.

If you have the time to add a regression test for this, that would be really nice, but I'm happy to merge it regardless. Just let me know?

PR #394, opened 2023-01-01
fix(console): fix `ViewOptions` default lang

### LINKED ISSUE #393: `tokio_console::config::ViewOptions::default()` set "en_us.UTF8" to lang

`tokio-console`

### What is the issue?

```rust
impl Default for ViewOptions {
    fn default() -> Self {
        Self {
            no_colors: false,
            lang: Some("en_us.UTF8".to_string()),  // 👈
            ascii_only: Some(false),
            truecolor: Some(true),
            palette: Some(Palette::All),
            toggles: ColorToggles {
                color_durations: Some(true),
                color_terminated: Some(true),
            },
        }
    }
}
```
https://github.com/tokio-rs/console/blob/70fc2c5d694b76499a68eca271ce4e337a313cfd/tokio-console/src/config.rs#L501

`"en_us.UTF-8"` is probably correct ? (with `-` between UTF and 8)

then

```rust
impl ViewOptions {
    pub fn is_utf8(&self) -> bool {
        if self.ascii_only.unwrap_or(false) {
            return false;
        }
        self.lang.as_deref().unwrap_or_default().ends_with("UTF-8") // 👈
    }
    //...
}
```
https://github.com/tokio-rs/console/blob/70fc2c5d694b76499a68eca271ce4e337a313cfd/tokio-console/src/config.rs#L423

in `ViewOptions::is_utf8()`, lang is assumed to be `UTF-8`

```
tokio-console gen-config > ./console.toml

bat console.toml | rg lang
lang = 'en_us.UTF8'
```

### Logs, error output, etc

_No response_

### Versions

```text
tokio-console 0.1.7
```

### Possible solution

set `"en_us.UTF-8"` to lang in `Default::default` in `ViewOptions`

### Additional context

_No response_

yes

ymgyt:
Fixes https://github.com/tokio-rs/console/issues/393

### DIFF
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -498,7 +498,7 @@ impl Default for ViewOptions {
     fn default() -> Self {
         Self {
             no_colors: false,
-            lang: Some("en_us.UTF8".to_string()),
+            lang: Some("en_us.UTF-8".to_string()),
             ascii_only: Some(false),
             truecolor: Some(true),
             palette: Some(Palette::All),


0xPoe (approved):
Looks good to me!

hawkw (approved):
thank you!

PR #402, opened 2023-03-02
fix(console): reduce decimal digits in UI

### LINKED ISSUE #224: Reduce the number of digits shown in UI
It probably doesn't make much sense to include fractions in the numbers displayed in the UI.

ref: https://news.ycombinator.com/item?id=29596537

hds:
The durations in the tokio-console UI are shown with a unit, so the
digits after the decimal separator are generally not relevant. Since
space is at a premium in a terminal UI, it makes sense to cut down where
possible.

This change removes all digits after the decimal separator in the tasks
table view. In the task detail view, 2 decimal places are kept.

Additionally, 4 new duration formats are added to represent
minutes-with-secondsi, hours-with-minutes, days-with-hours and days.
These are only applied once the leading unit is greater than zero.

For example, 59 seconds will be shown as seconds: `99s` and then 60
seconds will be shown as minutes-with-seconds: `1m00s`. New colors have
been chosen for each format.

NOTE: I had to make a small unrelated change in
`task::Details::make_percentiles_widget` to make clippy happy.

Fixes: #224

## PR Notes (not to be commited)

Here are some screenshots showing the reduction in decimal digits. It also
shows the new time formats and colors.

Tasks table (all colors):

palette_all-tasks

Task detail (all colors):

palette_all-task_detail

Tasks table (8 colors):

palette_8-tasks

### DIFF
--- a/tokio-console/src/view/mod.rs
+++ b/tokio-console/src/view/mod.rs
@@ -18,11 +18,12 @@ mod tasks;
 pub(crate) use self::styles::{Palette, Styles};
 pub(crate) use self::table::SortBy;
 
-const DUR_LEN: usize = 10;
+const DUR_LEN: usize = 6;
 // This data is only updated every second, so it doesn't make a ton of
 // sense to have a lot of precision in timestamps (and this makes sure
 // there's room for the unit!)
-const DUR_PRECISION: usize = 4;
+const DUR_LIST_PRECISION: usize = 2;
+const DUR_TABLE_PRECISION: usize = 0;
 const TABLE_HIGHLIGHT_SYMBOL: &str = ">> ";
 
 pub struct View {
--- a/tokio-console/src/view/mod.rs
+++ b/tokio-console/src/view/mod.rs
@@ -1,6 +1,5 @@
 use crate::view::{resources::ResourcesTable, table::TableListState, tasks::TasksTable};
 use crate::{input, state::State};
-use std::num::NonZeroUsize;
 use std::{borrow::Cow, cmp};
 use tui::{
     layout,
@@ -19,8 +18,7 @@ mod tasks;
 pub(crate) use self::styles::{Palette, Styles};
 pub(crate) use self::table::SortBy;
 
-// SAFETY: 6 is non-zero.
-const DUR_LEN: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(6) };
+const DUR_LEN: usize = 6;
 // This data is only updated every second, so it doesn't make a ton of
 // sense to have a lot of precision in timestamps (and this makes sure
 // there's room for the unit!)
--- a/tokio-console/src/view/mod.rs
+++ b/tokio-console/src/view/mod.rs
@@ -1,5 +1,6 @@
 use crate::view::{resources::ResourcesTable, table::TableListState, tasks::TasksTable};
 use crate::{input, state::State};
+use std::num::NonZeroUsize;
 use std::{borrow::Cow, cmp};
 use tui::{
     layout,
@@ -18,7 +19,8 @@ mod tasks;
 pub(crate) use self::styles::{Palette, Styles};
 pub(crate) use self::table::SortBy;
 
-const DUR_LEN: usize = 6;
+// SAFETY: 6 is non-zero.
+const DUR_LEN: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(6) };
 // This data is only updated every second, so it doesn't make a ton of
 // sense to have a lot of precision in timestamps (and this makes sure
 // there's room for the unit!)
--- a/tokio-console/src/view/styles.rs
+++ b/tokio-console/src/view/styles.rs
@@ -1,6 +1,6 @@
 use crate::config;
 use serde::{Deserialize, Serialize};
-use std::{borrow::Cow, str::FromStr};
+use std::{str::FromStr, time::Duration};
 use tui::{
     style::{Color, Modifier, Style},
     text::Span,
@@ -32,11 +32,31 @@ pub enum Palette {
     All,
 }
 
+enum FormattedDuration {
+    Days(String),
+    DaysHours(String),
+    HoursMinutes(String),
+    MinutesSeconds(String),
+    Debug(String),
+}
+
+impl FormattedDuration {
+    fn into_inner(self) -> String {
+        match self {
+            Self::Days(inner) => inner,
+            Self::DaysHours(inner) => inner,
+            Self::HoursMinutes(inner) => inner,
+            Self::MinutesSeconds(inner) => inner,
+            Self::Debug(inner) => inner,
+        }
+    }
+}
+
 fn fg_style(color: Color) -> Style {
     Style::default().fg(color)
 }
 
-// === impl Config ===
+// === impl Styles ===
 
 impl Styles {
     pub fn from_config(config: config::ViewOptions) -> Self {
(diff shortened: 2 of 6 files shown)


tokio-console/src/view/styles.rs
```diff
@@ -32,11 +32,31 @@ pub enum Palette {
     All,
 }
 
+enum FormattedDuration {
```
hawkw, line 38: it wasn't immediately obvious to me what the purpose of the `FormattedDuration` enum is, but then I see it's intended to indicate how the duration should be colored. It might be worth including a comment explaining what this is for?

tokio-console/src/view/styles.rs
```diff
@@ -126,39 +146,97 @@ impl Styles {
...
+        } else if secs >= 60 {
+            FormattedDuration::MinutesSeconds(format!(
+                "{:>width$}",
+                format!(
+                    "{minutes}m{seconds:02.0}s",
+                    minutes = secs / 60,
+                    seconds = secs % 60,
+                ),
+                width = width,
+            ))
```
hawkw: it feels a bit unfortunate that these all have a nested invocation of the `format!` macro, which allocates a new `String`, since the `String` produced by the inner `format!` only lives long enough to be formatted by the outer `format!` and then is immediately dropped. can we use `format_args!` here instead? `format_args!` creates a borrowed `fmt::Arguments` on the stack, rather than allocating a `String`, and it can immediately be formatted in the outer `format!` invocation.
```suggestion
            FormattedDuration::DaysHours(format!(
                "{:>width$}",
                format_args!(
                    "{days}d{hours:02.0}h",
                    days = hours / 24,
                    hours = hours % 24,
                ),
                width = width
            ))
        } else if secs >= 60 * 60 {
            let mins = secs / 60;
            FormattedDuration::HoursMinutes(format!(
                "{:>width$}",
                format_args!(
                    "{hours}h{minutes:02.0}m",
                    hours = mins / 60,
                    minutes = mins % 60,
                ),
                width = width
            ))
        } else if secs >= 60 {
            FormattedDuration::MinutesSeconds(format!(
                "{:>width$}",
                format_args!(
                    "{minutes}m{seconds:02.0}s",
                    minutes = secs / 60,
                    seconds = secs % 60,
                ),
                width = width,
            ))
```

tokio-console/src/view/styles.rs
```diff
@@ -126,39 +146,97 @@ impl Styles {
         }
     }
 
-    pub fn time_units<'a>(&self, text: impl Into<Cow<'a, str>>) -> Span<'a> {
-        let mut text = text.into();
-        if !self.toggles.color_durations() {
-            return Span::raw(text);
-        }
+    fn duration_text(&self, dur: Duration, width: usize, prec: usize) -> FormattedDuration {
```
hawkw: tiny style nit, take it or leave it (really not a big deal): in general, I like to put the public API functions above the inner implementation functions that they call, so I would have put `duration_text` after `time_units`. i find that this makes it easier to read the code in a top-down manner. 

of course, i'm not going to block merging the PR based on this --- it's just a minor personal preference.

tokio-console/src/view/styles.rs
```diff
@@ -126,39 +146,97 @@ impl Styles {
...
+                FormattedDuration::DaysHours(_) => fg_style(Color::Indexed(38)), // deep sky blue 2
+                FormattedDuration::HoursMinutes(_) => fg_style(Color::Indexed(39)), // deep sky blue 1
+                FormattedDuration::MinutesSeconds(_) => fg_style(Color::Indexed(45)), // turquoise 2
+                FormattedDuration::Debug(s) if s.ends_with("ps") => fg_style(Color::Indexed(40)), // green 3
+                FormattedDuration::Debug(s) if s.ends_with("ns") => fg_style(Color::Indexed(41)), // spring green 3
+                FormattedDuration::Debug(s) if s.ends_with("µs") || s.ends_with("us") => {
+                    fg_style(Color::Indexed(42))
+                } // spring green 2
+                FormattedDuration::Debug(s) if s.ends_with("ms") => fg_style(Color::Indexed(43)), // cyan 3
+                FormattedDuration::Debug(s) if s.ends_with('s') => fg_style(Color::Indexed(44)), // dark turquoise,
```
hawkw: i'm not sure if the coloring scheme here really works the way i'd like it to --- the intention behind the colors for durations was to form a gradient from the largest to the smallest durations. the idea being that it would be easy to visually distinguish the largest and smallest durations based on colors.

so, i would expect `Days` and `DaysHours` to be the same color, because they're in the same general order of magnitude. and, in the case of the `Ansi8`/`Ansi16` palette, i would expect `Days` and `DaysHours` to be the most blue, and picoseconds to be the most red (or in the case of the 256-color palette), the same gradient but from "most green" to "most blue".

tokio-console/src/view/styles.rs
```diff
@@ -126,39 +146,97 @@ impl Styles {
...
+                if let Some(mu_offset) = text.find("µs") {
+                    text.replace_range(mu_offset.., "us");
+                }
             }
+
+            FormattedDuration::Debug(text)
+        }
+    }
+
+    pub fn time_units<'a>(&self, dur: Duration, prec: usize, width: Option<usize>) -> Span<'a> {
```
hawkw, line 168: it's probably worth documenting what the `width` value does here, and what the difference between passing `None` and `Some` (and `None` versus `Some(0)`) is?

hawkw (changes requested):
I think this change makes sense overall --- I'm a fan of saving horizontal space in the tasks/resources list views, and i don't think that we need to display these values with as much precision as we do currently on those views.

I had some minor feedback on the implementation, most of which isn't a hard blocker. However, I would really prefer it if we could find a way to preserve the color gradients from smallest to longest durations --- I think that's a useful way of communicating this information visually. I'll be happy to merge this when that's addressed.

tokio-console/src/view/styles.rs
hds: Style question here. An `Option<usize>` seemed like a good idea at the time, but `None` and `Some(0)` are equivalent. Do you have any opinion on the merits of just passing a bare `usize` instead?

hawkw: You could also use an `Option<NonZeroUsize>` i suppose (which has the same representation as a single `usize`, with the actual zero value representing `None`)...but this is an internal API, so misuse-resistent API design is less important. I definitely prefer `None` to indicate "no width limit" than 0, though --- when reading the code you might expect "0 width" to mean "no characters", i guess?

hds: Okay. That sounds like a better option (using NonZero). Internal or no, as you said, None to mean no width restriction makes more sense than 0. I'll document the function too, so that it's clear. Thanks!

hawkw: sounds good to me!

tokio-console/src/view/styles.rs
hds: I've tried to make the 256-color once agian go from "most green" to "most blue", keeping the 2 units in days equal.

To fit all the units into a 7-color palette (becuase black can't be used), I've had to also double up at the bottom end, where I've set both picoseconds and nanoseconds to gray, assuming that picoseconds probably won't appear in the UI anyway.

How does this look?

Screenshot 2023-03-07 at 16 32 58

tokio-console/src/view/styles.rs
hds: Done.

tokio-console/src/view/styles.rs
hds: Easy, done.

tokio-console/src/view/styles.rs
```diff
@@ -126,39 +146,97 @@ impl Styles {
...
+                    "{hours}h{minutes:02.0}m",
+                    hours = mins / 60,
+                    minutes = mins % 60,
+                ),
+                width = width
+            ))
+        } else if secs >= 60 {
+            FormattedDuration::MinutesSeconds(format!(
+                "{:>width$}",
+                format_args!(
```
hawkw: hmm, looks like Clippy is saying that I was wrong about this: https://github.com/tokio-rs/console/actions/runs/4348338005/jobs/7596743879

tokio-console/src/view/styles.rs
hawkw: these palettes look good to me! thank you!

tokio-console/src/view/styles.rs
hds: Yeah, it doesn't work.

Screenshot 2023-03-07 at 18 18 37

I've been fighting with this for the last 20 minutes trying to work out a way to avoid that extra allocation. Right now it doesn't seem possible.

I have one last idea...

hds: Fixed it, we can just subtract the 4 known characters and calculate what the width **should** be, then a single `format!` can be used.

tokio-console/src/view/styles.rs
hds: So, this is a bit ugly because I now need this line when declaring `DUR_LEN` or I need to be `unwrap`ing in a bunch of places because `new` returns a `Option<NonZeroUsize>`.

```rust
// SAFETY: 6 is non-zero.
const DUR_LEN: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(6) };
```

https://github.com/tokio-rs/console/pull/402/files#diff-295fe639b204aaddc6b81e4c68fbb9bc7f2fb3aef43fbca622e99620dfc7debaR22-R23

Thoughts? Should I just go back to the `Option<usize>` and explain that `Some(0)` and `None` are equivalent?

hawkw: ugh, that's annoying. i would prefer not to put unsafe code in the const. at the callsite, we could call `time_units(..., NonZeroUsize::new(DUR_LEN))` but that's more verbose. i think it would be fine to allow `Some(0)` and `None` to mean the same thing, or just get rid of the option...

tokio-console/src/view/styles.rs
hawkw: oh, nice! that's great. i would have been fine with just doing the second allocation if we couldn't figure out something nicer, but that's a cool solution!

tokio-console/src/view/mod.rs
```diff
@@ -18,11 +19,13 @@ mod tasks;
 pub(crate) use self::styles::{Palette, Styles};
 pub(crate) use self::table::SortBy;
 
-const DUR_LEN: usize = 10;
+// SAFETY: 6 is non-zero.
+const DUR_LEN: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(6) };
```
hawkw: i don't love that this is `unsafe`. note that `NonZeroUsize::new` is `const fn` in our MSRV, so I believe this could just be 
```suggestion
const DUR_LEN: NonZeroUsize = NonZeroUsize::new(6).expect("6 is non-zero");
```

alternatively, we could just get rid of the `NonZeroUsize`, which i would also be fine about.

tokio-console/src/view/styles.rs
```diff
@@ -126,39 +156,104 @@ impl Styles {
...
+            let days = secs / (60 * 60 * 24);
+            FormattedDuration::Days(format!("{days:>width$}d", days = days, width = width))
+        } else if secs >= 60 * 60 * 24 {
+            let hours = secs / (60 * 60);
+            FormattedDuration::DaysHours(format!(
+                "{days:>leading_width$}d{hours:02.0}h",
+                days = hours / 24,
+                hours = hours % 24,
+                // Subtract the known 4 characters that trail the days value.
+                leading_width = if width > 4 { width - 4 } else { 0 }
```
hawkw: nit, take it or leave it: this looks like a manual implementation of `saturating_sub`:
```suggestion
                leading_width = width.saturating_sub(4),
```

tokio-console/src/view/styles.rs
```diff
@@ -126,39 +156,104 @@ impl Styles {
...
+            ))
+        } else if secs >= 60 * 60 {
+            let mins = secs / 60;
+            FormattedDuration::HoursMinutes(format!(
+                "{hours:>leading_width$}h{minutes:02.0}m",
+                hours = mins / 60,
+                minutes = mins % 60,
+                // Subtract the known 4 characters that trail the hours value.
+                leading_width = if width > 4 { width - 4 } else { 0 }
```
hawkw: similarly:
```suggestion
                leading_width = width.saturating_sub(4),
```

tokio-console/src/view/styles.rs
```diff
@@ -126,39 +156,104 @@ impl Styles {
...
+            ))
+        } else if secs >= 60 {
+            FormattedDuration::MinutesSeconds(format!(
+                "{minutes:>leading_width$}m{seconds:02.0}s",
+                minutes = secs / 60,
+                seconds = secs % 60,
+                // Subtract the known 4 characters that trail the minutes value.
+                leading_width = if width > 4 { width - 4 } else { 0 },
```
hawkw:
```suggestion
                leading_width = width.saturating_sub(4),
```

tokio-console/src/view/mod.rs
hds: The unwrap wasn't const in our MSRV I believe.

I changed this back to `Option<usize>`.

tokio-console/src/view/styles.rs
hds: Changed it back to `Option<usize>`.

tokio-console/src/view/styles.rs
hds: Nice, thanks for the tip!

tokio-console/src/view/styles.rs
hds: Nice, thanks for the tip!

tokio-console/src/view/styles.rs
hds: Nice, thanks for the tip!

tokio-console/src/view/mod.rs
hds: Yeah, I get the following when trying that code:

```
error: `std::option::Option::<T>::expect` is not yet stable as a const fn
  --> tokio-console/src/view/mod.rs:22:31
   |
22 | const DUR_LEN: NonZeroUsize = NonZeroUsize::new(6).expect("6 is non-zero");
   |        
```

So it seems like while `NonZeroUsize::new()` is stable, we would also need `const Option::unwrap()`, which isn't yet stable.

Another thing I've learnt today.

hawkw (approved):
looks good, thank you!

PR #403, opened 2023-03-13
feat(console): use tokio task ids in task views

### LINKED ISSUE #385: Task IDs not stable

tokio-console, console-subscriber

### What is the issue?

I have a service which creates two long-running tasks as part of its initialization. They end up with task ids 1 & 2 as expected. The service creates a task per request, and I see those show up as expected.

image
*initial state*

image
*after running some requests*

However, when I reconnect, I see the task ids have scrambled - the long running tasks no longer have ids 1 & 2 (in this case 2 is preserved, but associated with the other long-running task).
image
*immediately disconnect and reconnect*

I'd really like task ids to be stable so that I can use them for reference. Ideally they'd be monotonically increasing in order of task spawns, so they can also be used to understand when things are spawned in relation to each other.

1. Have service which creates some long-running tasks. Start it.
2. Attach tokio-console
3. Run some activity, including transient tasks. Note long-running task ids
4. Disconnect and reconnect tokio-console. Expect to see task ids unchanged, but they won't be.

### Logs, error output, etc

_No response_

### Versions

```
tokio-console 0.1.7
console-subscriber 0.1.8
```

### Possible solution

_No response_

### Additional context

_No response_

maybe

hds:
Tokio Console generates its own sequential Id for internal tracking and
indexing of objects (tasks, resources, etc.). However, this Id will be
recreated if Console is restarted.

In order to provide more useful information to the user, the task Id
generated by Tokio can be used in the task list and task details screens
instead. If used in this way, the ID field in the task list and task
detail views will be stable across restarts of Console (assuming the
monitored application is not restarted).

This also frees up horizontal space, as the `task.id` attribute
doesn't need to be displayed separately.

The disadvantage of using Tokio's task Id is that it is not guaranteed
to be present by the type system.

To avoid problems with missing task Ids, the Tokio task Id is store in
addition to the `span::Id` (used to communicate with the
`console-subscriber`) and the sequential console `Id` (used in the
`store`). If a task is missing the `task.id` field for whatever reason
it will still appear, but with an empty ID. If multiple runtimes are
active, then duplicate ID values will appear.

Fixes: #385

### DIFF
--- a/tokio-console/src/state/tasks.rs
+++ b/tokio-console/src/state/tasks.rs
@@ -5,7 +5,7 @@ use crate::{
         histogram::DurationHistogram,
         pb_duration,
         store::{self, Id, SpanId, Store},
-        Field, Metadata, Visibility,
+        Field, FieldValue, Metadata, Visibility,
     },
     util::Percentage,
     view,
@@ -58,6 +58,17 @@ pub(crate) enum TaskState {
 
 pub(crate) type TaskRef = store::Ref<Task>;
 
+/// The Id for a Tokio task.
+///
+/// This should be equivalent to [`tokio::task::Id`], which can't be
+/// used because it's not possible to construct outside the `tokio`
+/// crate.
+///
+/// Within the context of `tokio-console`, we don't depend on it
+/// being the same as Tokio's own type, as the task id is recorded
+/// as a `u64` in tracing and then sent via the wire protocol as such.
+pub(crate) type TaskId = u64;
+
 #[derive(Debug)]
 pub(crate) struct Task {
     /// The task's pretty (console-generated, sequential) task ID.
@@ -65,6 +76,11 @@ pub(crate) struct Task {
     /// This is NOT the `tracing::span::Id` for the task's tracing span on the
     /// remote.
     id: Id<Task>,
+    /// The `tokio::task::Id` in the remote tokio runtime.
+    ///
+    /// This Id may not be unique if there are multiple runtimes in the
+    /// remote process.
+    task_id: Option<TaskId>,
     /// The `tracing::span::Id` on the remote process for this task's span.
     ///
     /// This is used when requesting a task details stream.
@@ -147,6 +163,7 @@ impl TasksState {
                     }
                 };
                 let mut name = None;
+                let mut task_id = None;
                 let mut fields = task
                     .fields
                     .drain(..)
@@ -155,6 +172,13 @@ impl TasksState {
                         // the `task.name` field gets its own column, if it's present.
                         if &*field.name == Field::NAME {
                             name = Some(strings.string(field.value.to_string()));
+                            return None;
+                        }
+                        if &*field.name == Field::TASK_ID {
+                            task_id = match field.value {
+                                FieldValue::U64(id) => Some(id as TaskId),
+                                _ => None,
+                            };
                             return None;
                         }
                         Some(field)
@@ -178,6 +202,7 @@ impl TasksState {
                 let mut task = Task {
                     name,
                     id,
+                    task_id,
                     span_id,
                     short_desc,
                     formatted_fields,
@@ -239,6 +264,10 @@ impl Details {
 impl Task {
     pub(crate) fn id(&self) -> Id<Task> {
         self.id
+    }
+
+    pub(crate) fn task_id(&self) -> Option<TaskId> {
+        self.task_id
     }
 
     pub(crate) fn span_id(&self) -> SpanId {
@@ -426,7 +455,9 @@ impl Default for SortBy {
 impl SortBy {
     pub fn sort(&self, now: SystemTime, tasks: &mut [Weak<RefCell<Task>>]) {
         match self {
-            Self::Tid => tasks.sort_unstable_by_key(|task| task.upgrade().map(|t| t.borrow().id)),
+            Self::Tid => tasks.sort_unstable_by_key(|task| {
+                task.upgrade().map(|t| t.borrow().task_id.map(|id| id))
+            }),
             Self::Name => {
                 tasks.sort_unstable_by_key(|task| task.upgrade().map(|t| t.borrow().name.clone()))
             }
--- a/tokio-console/src/state/tasks.rs
+++ b/tokio-console/src/state/tasks.rs
@@ -194,9 +194,11 @@ impl TasksState {
                 // remap the server's ID to a pretty, sequential task ID
                 let id = ids.id_for(span_id);
 
-                let short_desc = strings.string(match name.as_ref() {
-                    Some(name) => format!("{} ({})", id, name),
-                    None => format!("{}", id),
+                let short_desc = strings.string(match (task_id, name.as_ref()) {
+                    (Some(task_id), Some(name)) => format!("{} ({})", task_id, name),
+                    (Some(task_id), None) => format!("{}", task_id),
+                    (None, Some(name)) => format!("({})", name),
+                    (None, None) => "".to_owned(),
                 });
 
                 let mut task = Task {
--- a/tokio-console/src/state/tasks.rs
+++ b/tokio-console/src/state/tasks.rs
@@ -77,14 +77,13 @@ pub(crate) struct Task {
     /// remote.
     id: Id<Task>,
     /// The `tokio::task::Id` in the remote tokio runtime.
-    ///
-    /// This Id may not be unique if there are multiple runtimes in the
-    /// remote process.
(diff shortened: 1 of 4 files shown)


hds:
@hawkw I'm a bit torn about this approach. Adding "yet another Id" seems a bit hacky, but replacing either the `span::Id` or the sequential console `Id` which are both shared amongst all object types also seemed like a bad idea.

We could skip tasks with no tokio task `Id` (which I think shouldn't happen), the same way we skip tasks without a `span::Id` or a metadata `Id`, which would at least guarantee that some Id is present.

As a follow up to this change, I think that it's worth trying to get some per runtime identifier in (unstable) tokio so that we can distinguish tasks from different runtimes, an optional user supplied name would also be nice, but it's neither necessary nor sufficient to properly distinguish tasks. I think there's already an Issue for this, I'll see if I can find it.

hawkw:
> @hawkw I'm a bit torn about this approach. Adding "yet another Id" seems a bit hacky, but replacing either the `span::Id` or the sequential console `Id` which are both shared amongst all object types also seemed like a bad idea.

Hmm, I'm not sure if I love having three separate concepts of IDs...the primary reason we added the sequential console-generated IDs was because the raw span IDs are quite long regardless of the number of tasks, because span IDs are generated based on which thread a span was created on (among other things), and are non-sequential and able to be reused by `tracing`. 

The tokio task IDs and console generated sequential IDs have more or less the same characteristics (they are sequential and not reused when a task completes). However, because the console-generated sequential IDs are generated in the `tokio-console` TUI rather than in the instrumented application, they have some additional issues: they are not persistent across `tokio-console` sessions (see https://github.com/tokio-rs/console/issues/385). 

Since the tokio-assigned task IDs and the console-assigned task IDs have the same general characteristics, my preference would be to remove the console IDs entirely. This would simplify the implementation significantly, as we would no longer have to track ID remappings in the console, and would solve #385. We would preferentially display the tokio task IDs in the console UI, and fall back to the raw span ID if a task ID is unknown. We may want to have some way of visually distinguishing tokio IDs and span IDs in the UI...

> We could skip tasks with no tokio task `Id` (which I think shouldn't happen), the same way we skip tasks without a `span::Id` or a metadata `Id`, which would at least guarantee that some Id is present.

I think we should probably use the span ID if no task ID is present, but clearly indicate that these are different.

> As a follow up to this change, I think that it's worth trying to get some per runtime identifier in (unstable) tokio so that we can distinguish tasks from different runtimes, an optional user supplied name would also be nice, but it's neither necessary nor sufficient to properly distinguish tasks. I think there's already an Issue for this, I'll see if I can find it.

+1 for this --- we should definitely add a runtime ID/name notion to Tokio. But for now, I think it's acceptable to permit collisions in the case of multiple runtimes, since they can be resolved by comparing the task's span IDs.

tokio-console/src/state/tasks.rs
```diff
@@ -170,14 +194,17 @@ impl TasksState {
                 // remap the server's ID to a pretty, sequential task ID
                 let id = ids.id_for(span_id);
 
-                let short_desc = strings.string(match name.as_ref() {
-                    Some(name) => format!("{} ({})", id, name),
-                    None => format!("{}", id),
+                let short_desc = strings.string(match (task_id, name.as_ref()) {
+                    (Some(task_id), Some(name)) => format!("{} ({})", task_id, name),
+                    (Some(task_id), None) => format!("{}", task_id),
+                    (None, Some(name)) => format!("({})", name),
```
hawkw: i think this should probably just be the name, without parens --- the parens were just intended to separate the ID from the name
```suggestion
                    (None, Some(name)) => name.to_owned()

File: tokio-console/src/state/tasks.rs [hunk 2]
```diff
@@ -170,14 +194,17 @@ impl TasksState {
                 // remap the server's ID to a pretty, sequential task ID
                 let id = ids.id_for(span_id);
 
-                let short_desc = strings.string(match name.as_ref() {
-                    Some(name) => format!("{} ({})", id, name),
-                    None => format!("{}", id),
+                let short_desc = strings.string(match (task_id, name.as_ref()) {
+                    (Some(task_id), Some(name)) => format!("{} ({})", task_id, name),
```
hawkw: style nit, take it or leave it: could use inlined format args here:
```suggestion
                    (Some(task_id), Some(name)) => format!("{task_id} ({name})"),
```

tokio-console/src/state/tasks.rs
```diff
@@ -170,14 +194,17 @@ impl TasksState {
...
+                    (Some(task_id), None) => format!("{}", task_id),
```
hawkw: style nit: since we're only formatting a single value, we could avoid the use of the format macro:
```suggestion
                    (Some(task_id), None) => task_id.to_string(),
```

tokio-console/src/view/tasks.rs
```diff
@@ -122,7 +122,7 @@ impl TableList<11> for TasksTable {
                         warnings,
                         Cell::from(id_width.update_str(format!(
                             "{:>width$}",
-                            task.id(),
+                            task.task_id().map(|id| id.to_string()).unwrap_or_default(),
```
hawkw: since the task ID is not going to change over the lifetime of the task, and we have to format an ID for each task every time we draw the task list, i wonder if we want to cache a string representation of the task's ID as soon as we receive a task --- we could have an `id_str` field on the `Task` struct, which would allow us to avoid calling `to_string()` on the ID for every task every time we render the tasks list?

hawkw (commented):
Since this is a relatively small change, I would be open to merging it as-is and then removing the console-generated IDs in a follow-up, or something. Removing the console's sequential IDs would be a larger change that might require some discussion, while this branch is basically a quick fix that will improve UX for now regardless of what we decide to do long term.

hds:
> +1 for this --- we should definitely add a runtime ID/name notion to Tokio. But for now, I think it's acceptable to permit collisions in the case of multiple runtimes, since they can be resolved by comparing the task's span IDs.

Actually it turns out that tokio task ids are globally unique. I had thought (assumed) that they would depend on the runtime, but there's a once_cell atomic counter (or similar) used to generate them. Means that there aren't collisions, but the runtime ids would still be useful.

tokio-console/src/state/tasks.rs
hds: I really thought that in-line format args was a newer feature, but I looked it up (actually, I kind of just asked someone) and realised it's from 1.58. I'll apply this.

tokio-console/src/view/tasks.rs
hds: I need to check the docs, but do we really have to format all rows in a table every time it's opened? Do we not only format the visible rows?

hawkw: er, that's what i meant. the point is that this work is performed every time we redraw the table.

hds: OK, got it. I understand now.

tokio-console/src/state/tasks.rs
```diff
@@ -195,9 +195,9 @@ impl TasksState {
                 let id = ids.id_for(span_id);
 
                 let short_desc = strings.string(match (task_id, name.as_ref()) {
-                    (Some(task_id), Some(name)) => format!("{} ({})", task_id, name),
-                    (Some(task_id), None) => format!("{}", task_id),
-                    (None, Some(name)) => format!("({})", name),
+                    (Some(task_id), Some(name)) => format!("{task_id} ({name})"),
+                    (Some(task_id), None) => task_id.to_string(),
+                    (None, Some(name)) => name.to_owned()
```
hawkw: missing `,` here (i think that was my fault, sorry!)
```suggestion
                    (None, Some(name)) => name.to_owned(),
```

tokio-console/src/view/tasks.rs
hds: @hawkw  I've made some changes to store a string of the id. Please let me know if this is what you had in mind.

tokio-console/src/state/tasks.rs
hds: Yep, rustc caught that one for me, I didn't notice it either. (-;

I needed to use `name.as_ref().to_owned()` here because `name` is an `InternedStr`. Not sure if there's a better pattern for this.

hawkw: Oh, whoops, yeah, that makes sense. I assumed it was just a string.

hawkw (approved):

PR #406, opened 2023-04-03
feat(console): add scheduled time per task

hds:
Each task displays the sum of the time it has been idle and busy, as
well as the total. The idle time includes the time between when a task
is woken, and when the runtime actually polls that task.

There are cases where a task may be scheduled for a long time after
being woken, before it is polled. Especially if many tasks are woken at
the same time and don't yield back to the runtime quickly.

To add visibility to this, the total time that a task is scheduled
before being polled has been added. Additionally, a new task state
`Scheduled` has been added. This is displayed in both the tasks table
and in the task detail view.

In the `console-api`, the `last_wake` time has been moved from
`tasks::Stats` to `common::PollStats`, where the total `scheduled_time`
for the task has also been added.

This is the first of two parts. In the second part (#409), a histogram for
scheduled time will be added, the equivalent of the poll time histogram
which is already available on the task detail screen.

To show a pathological case which may lead to needing to see the
scheduled time, a new example has been added to the `console-subscriber`
examples.

## Breaking wire format change

Just to be super clear, by moving `last_wake` from `tasks::Stats` to
`common::PollStats`, this is a breaking change to the wire format (at least
functionally). It's not stricly necessary, but leaving it as it is will make some
other things messier.

Of course, mess can be cleaned up later if we want a more polished upgrade
path than requiring that `console-subscriber` and `tokio-console` are
updated at the same time.

## PR Notes

This PR does something adjacent to what is described in #50, but not quite.

The unicode character used for a `SCHED` task is ⏫.

The second PR (#409) will record a scheduled time histogram the same as it
recorded for poll times. These two changes should go in together (and certainly
shouldn't be released separately). However, this PR is already quite big, so they'll
be separated out.

The idea is that this PR isn't merged until the next one has also been reviewed
and approved. It would be good to get some feedback at this stage though.

The task list view with the new column for `Sched` time:

a tasks table view for the long-scheduled example

The `Task` block in the task detail view showing the new `Scheduled` time entry.

The task block on the task detail view for the rx task in the long-scheduled example

### DIFF
--- a/console-api/proto/common.proto
+++ b/console-api/proto/common.proto
@@ -161,6 +161,10 @@ message PollStats {
     // Subtracting this timestamp from `created_at` can be used to calculate the
     // time to first poll for this object, a measurement of executor latency.
     optional google.protobuf.Timestamp first_poll = 3;
+    // The timestamp of the most recent time this object was woken.
+    //
+    // If this is `None`, the object has not yet been woken.
+    optional google.protobuf.Timestamp last_wake = 7;
     // The timestamp of the most recent time this objects's poll method was invoked.
     //
     // If this is `None`, the object has not yet been polled.
@@ -180,8 +184,15 @@ message PollStats {
     // all polls. Note that this includes only polls that have completed and is
     // not reflecting any inprogress polls. Subtracting `busy_time` from the
     // total lifetime of the polled object results in the amount of time it
-    // has spent *waiting* to be polled.
+    // has spent *waiting* to be polled (including `scheduled_time`).
     google.protobuf.Duration busy_time = 6;
+    // The total duration this object was scheduled prior to being polled, summed
+    // across all poll cycles. Note that this includes only polls that have
+    // started and is not reflecting any scheduled state where the polling hasn't
+    // yet finished. Subtracting both `busy_time` and `scheduled_time` from the
+    // total lifetime of the polled object results in the amount of time it spent
+    // unable to progress because it was waiting on some resource.
+    google.protobuf.Duration scheduled_time = 8;
 }
 
 // State attributes of an entity. These are dependent on the type of the entity.
--- a/console-api/proto/common.proto
+++ b/console-api/proto/common.proto
@@ -161,10 +161,6 @@ message PollStats {
     // Subtracting this timestamp from `created_at` can be used to calculate the
     // time to first poll for this object, a measurement of executor latency.
     optional google.protobuf.Timestamp first_poll = 3;
-    // The timestamp of the most recent time this object was woken.
-    //
-    // If this is `None`, the object has not yet been woken.
-    optional google.protobuf.Timestamp last_wake = 7;
     // The timestamp of the most recent time this objects's poll method was invoked.
     //
     // If this is `None`, the object has not yet been polled.
@@ -186,13 +182,6 @@ message PollStats {
     // total lifetime of the polled object results in the amount of time it
     // has spent *waiting* to be polled (including `scheduled_time`).
     google.protobuf.Duration busy_time = 6;
-    // The total duration this object was scheduled prior to being polled, summed
-    // across all poll cycles. Note that this includes only polls that have
-    // started and is not reflecting any scheduled state where the polling hasn't
-    // yet finished. Subtracting both `busy_time` and `scheduled_time` from the
-    // total lifetime of the polled object results in the amount of time it spent
-    // unable to progress because it was waiting on some resource.
-    google.protobuf.Duration scheduled_time = 8;
 }
 
 // State attributes of an entity. These are dependent on the type of the entity.
--- a/console-api/proto/common.proto
+++ b/console-api/proto/common.proto
@@ -180,7 +180,8 @@ message PollStats {
     // all polls. Note that this includes only polls that have completed and is
     // not reflecting any inprogress polls. Subtracting `busy_time` from the
     // total lifetime of the polled object results in the amount of time it
-    // has spent *waiting* to be polled (including `scheduled_time`).
+    // has spent *waiting* to be polled (including `scheduled_time`, if
+    // this object is a task).
     google.protobuf.Duration busy_time = 6;
 }
 
--- a/console-api/proto/common.proto
+++ b/console-api/proto/common.proto
@@ -177,11 +177,13 @@ message PollStats {
     // its poll method has completed.
     optional google.protobuf.Timestamp last_poll_ended = 5;
     // The total duration this object was being *actively polled*, summed across
-    // all polls. Note that this includes only polls that have completed and is
-    // not reflecting any inprogress polls. Subtracting `busy_time` from the
+    // all polls.
+    //
+    // Note that this includes only polls that have completed, and does not
+    // reflect any in-progress polls. Subtracting `busy_time` from the
     // total lifetime of the polled object results in the amount of time it
-    // has spent *waiting* to be polled (including `scheduled_time`, if
-    // this object is a task).
+    // has spent *waiting* to be polled (including the `scheduled_time` value
+    // from `TaskStats`, if this is a task).
     google.protobuf.Duration busy_time = 6;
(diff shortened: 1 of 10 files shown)


console-subscriber/src/stats.rs
```diff
@@ -466,18 +462,49 @@ impl ToProto for ResourceStats {
...
+        let scheduled = match (timestamps.last_wake, timestamps.last_poll_ended) {
+            // If the last poll ended after the last wake then it was likely
+            // a self-wake, so we measure from the end of the last poll instead.
+            // This also ensures that `busy_time` and `scheduled_time` don't overlap.
+            (Some(last_wake), Some(last_poll_ended)) if last_poll_ended > last_wake => {
+                last_poll_ended
+            }
+            (Some(last_wake), _) => last_wake,
+            (None, _) => return, // Async operations record polls, but not wakes
+        };
```
hawkw: nit, take it or leave it: I believe this is equivalent to just using the `cmp::PartialOrd` implementation for `Option`:
```suggestion
        // If the last poll ended after the last wake then it was likely
        // a self-wake, so we measure from the end of the last poll instead.
        // This also ensures that `busy_time` and `scheduled_time` don't overlap.
        let scheduled = match std::cmp::max(timestamps.last_wake, timestamps.last_poll_ended) {
            Some(scheduled) => scheduled,
            None => return, // Async operations record polls, but not wakes
        }
```

hawkw:
> Just to be super clear, by moving `last_wake` from `tasks::Stats` to
> `common::PollStats`, this is a breaking change to the wire format (at least
> functionally). It's not stricly necessary, but leaving it as it is will make some
> other things messier.

I'm not totally sure if I understand the motivation for this change. Only tasks can be woken, so why do we now include a wake timestamp on non-task objects?

console-subscriber/examples/long_scheduled.rs
```diff
@@ -0,0 +1,73 @@
...
+//! time between being woken and being polled.
+//!
+//! It consists of a channel where a sender task sends a message
+//! through the channel and then immediately does a lot of work
+//! (simulated in this case by a call to `std::thread::sleep`).
+//!
+//! As soon as the sender task calls `send()` the receiver task gets
+//! woken, but because there's only a single worker thread, it doesn't
+//! get polled until after the sender task has finished "working" and
+//! yields (via `tokio::time::sleep`).
```
hawkw, line 13: it might be worth describing how this behavior pattern is visible in the console in this comment as well?

tokio-console/src/state/tasks.rs
```diff
@@ -429,11 +454,17 @@ impl From<proto::tasks::Stats> for TaskStats {
 
         let poll_stats = pb.poll_stats.expect("task should have poll stats");
         let busy = poll_stats.busy_time.map(pb_duration).unwrap_or_default();
-        let idle = total.map(|total| total.checked_sub(busy).unwrap_or_default());
+        let scheduled = poll_stats
+            .scheduled_time
+            .map(pb_duration)
+            .unwrap_or_default();
+        let idle = total.map(|total| total.checked_sub(busy + scheduled).unwrap_or_default());
```
hawkw, line 458: just taking a moment to note that we've decided to change the task's idle time to be the time the task was idle and _not_ scheduled. i think this makes sense but is worth noting.

alternatively, it occurs to me that we could leave "idle" as the total live-but-not-running time, and introduce a new notion of "waiting" time for tasks that are not running but have not been scheduled yet. but, i'm not sure if it makes sense to add so many different time metrics, and i think this change is fairly reasonable...

console-subscriber/examples/long_scheduled.rs
hds: Good idea. Done.

tokio-console/src/state/tasks.rs
hds: I think that it might be more confusing to have times that don't add up to the total time (or add up in pairs). Are you happy to leave this as is?

console-subscriber/src/stats.rs
```diff
@@ -482,16 +482,13 @@ impl<H: RecordPoll> PollStats<H> {
...
-            (Some(last_wake), _) => last_wake,
-            (None, _) => return, // Async operations record polls, but not wakes
-        };
+        // If the last poll ended after the last wake then it was likely
+        // a self-wake, so we measure from the end of the last poll instead.
+        // This also ensures that `busy_time` and `scheduled_time` don't overlap.
+        let scheduled = match std::cmp::max(timestamps.last_wake, timestamps.last_poll_ended) {
+            Some(scheduled) => scheduled,
+            None => return, // Async operations record polls, but not wakes
+        }
```
hawkw: agh, sorry, my suggestion was missing a semicolon
```suggestion
        };
```

hds:
> > Just to be super clear, by moving `last_wake` from `tasks::Stats` to
> > `common::PollStats`, this is a breaking change to the wire format (at least
> > functionally). It's not stricly necessary, but leaving it as it is will make some
> > other things messier.
> 
> I'm not totally sure if I understand the motivation for this change. Only tasks can be woken, so why do we now include a wake timestamp on non-task objects?

I think that here I was basically commiting the sin of exposing implementation details in the API. I needed to move the `last_wake` time in the `console-subscriber` in order to calculate the scheduled time, and so I moved it in the API too.

You're right, it doesn't make sense. And I think that by the same reasoning, it also doesn't make sense to put the `scheduled_time` inside `common::PollStats` either is it isn't used by async ops. I'm going to move it into `tasks::Stats`.

tokio-console/src/state/tasks.rs
hawkw: yes, i think the current definitions of "idle time" works for me...we should probably document what exactly these various time metrics mean, though...

hds: I'll add some more info to the docs. Would you like that in this PR, or is a follow up okay?

console-api/proto/common.proto
```diff
@@ -180,7 +180,7 @@ message PollStats {
     // all polls. Note that this includes only polls that have completed and is
     // not reflecting any inprogress polls. Subtracting `busy_time` from the
     // total lifetime of the polled object results in the amount of time it
-    // has spent *waiting* to be polled.
+    // has spent *waiting* to be polled (including `scheduled_time`).
```
hawkw: since scheduled time is only defined for tasks, we should maybe add
```suggestion
    // has spent *waiting* to be polled (including `scheduled_time`, if
    // this object is a task).
```

console-api/proto/tasks.proto
```diff
@@ -130,6 +130,13 @@ message Stats {
     common.PollStats poll_stats = 7;
     // The total number of times this task has woken itself.
     uint64 self_wakes = 8;
+    // The total duration this task was scheduled prior to being polled, summed
+    // across all poll cycles. Note that this includes only polls that have
+    // started and is not reflecting any scheduled state where the task hasn't
+    // yet been polled. Subtracting both `busy_time` and `scheduled_time` from the
```
hawkw: should this maybe note that `busy_time` is defined on `PollStats`?

console-api/proto/tasks.proto
```diff
@@ -130,6 +130,13 @@ message Stats {
...
+    // total lifetime of the polled object results in the amount of time it spent
```
hawkw: nit: since this is now only defined on tasks, we can just say "task" here:
```suggestion
    // total lifetime of the polled task results in the amount of time it spent
```

hawkw (approved):
some tiny docs nits, but beyond that, looks good to me!

hawkw:
Agh, sorry, we raced --- I also pushed some docs fixes because I was about to merge this.

hds:
No worries, I just added what I think was a missing backtick too.

hds:
@hawkw I think I disabled your auto merge with that last commit.

PR #408, opened 2023-04-11
refac(console): factor out `Durations` widget from task view

hds:
There are 2 widgets which display the poll times for a task in the
detail view. The poll times percentiles are always displayed and if
UTF-8 is available, then a sparkline histogram is also shown to the
right.

The logic for displaying these two widgets is quite long and is
currently interspersed within the `render` function for the task detail
view plus helper functions. Additionally, it is not easy to add a second
set of widgets showing the time between waking and being polled for a
task which is planned for #409.

This change factors out that logic into separate widgets.

There was already a separate widget `MiniHistogram`. Some of the logic
that was previously in the task detail view has been moved here.

A new widget `Percentiles` has been added to encapsulate the logic for
preparing and displaying the percentiles.

A top level `Durations` widget occupies the entire width of the task
detail view and control the horizontal layout of the `Percentiles` and
`MiniHistogram` widgets. The new widget will also supress the histogram
if there isn't at least enough room to display the legend at the bottom.

## PR Notes

This refactoring in this PR was heavily inspired by #50, just that the requirements
were a bit different.

### DIFF
--- a/tokio-console/src/view/durations.rs
+++ b/tokio-console/src/view/durations.rs
@@ -12,6 +12,7 @@ use crate::{
 
 // This is calculated so that a legend like the below generally fits:
 // │0647.17µs  909.31µs │
+// This also gives at characters for the sparkline itself.
 const MIN_HISTOGRAM_BLOCK_WIDTH: u16 = 22;
 
 /// This is a tui-rs widget to visualize durations as a list of percentiles
@@ -20,7 +21,10 @@ const MIN_HISTOGRAM_BLOCK_WIDTH: u16 = 22;
 /// This widget wraps the [`Percentiles`] and [`MiniHistogram`] widgets which
 /// are displayed side by side. The mini-histogram will only be displayed if
 ///   a) UTF-8 support is enabled via [`Styles`]
-///   b) There is at least a minimum width left after drawing the percentiles
+///   b) There is at least a minimum width (22 characters to display the full
+///      bottom legend) left after drawing the percentiles
+///
+/// This
 ///
 /// [`Styles`]: crate::view::Styles
 pub(crate) struct Durations<'a> {
@@ -51,7 +55,7 @@ impl<'a> Widget for Durations<'a> {
                     .constraints(
                         [
                             layout::Constraint::Length(percentiles_width),
-                            layout::Constraint::Min(50),
+                            layout::Constraint::Min(MIN_HISTOGRAM_BLOCK_WIDTH),
                         ]
                         .as_ref(),
                     )
--- a/tokio-console/src/view/durations.rs
+++ b/tokio-console/src/view/durations.rs
@@ -1,0 +1,104 @@
+use std::cmp;
+
+use tui::{
+    layout::{self},
+    widgets::Widget,
+};
+
+use crate::{
+    state::histogram::DurationHistogram,
+    view::{self, mini_histogram::MiniHistogram, percentiles::Percentiles},
+};
+
+// This is calculated so that a legend like the below generally fits:
+// │0647.17µs  909.31µs │
+const MIN_HISTOGRAM_BLOCK_WIDTH: u16 = 22;
+
+/// This is a tui-rs widget to visualize durations as a list of percentiles
+/// and if possible, a mini-histogram too.
+///
+/// This widget wraps the [`Percentiles`] and [`MiniHistogram`] widgets which
+/// are displayed side by side. The mini-histogram will only be displayed if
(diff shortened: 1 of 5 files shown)


tokio-console/src/view/durations.rs
```diff
@@ -0,0 +1,104 @@
...
+// │0647.17µs  909.31µs │
+const MIN_HISTOGRAM_BLOCK_WIDTH: u16 = 22;
+
+/// This is a tui-rs widget to visualize durations as a list of percentiles
+/// and if possible, a mini-histogram too.
+///
+/// This widget wraps the [`Percentiles`] and [`MiniHistogram`] widgets which
+/// are displayed side by side. The mini-histogram will only be displayed if
+///   a) UTF-8 support is enabled via [`Styles`]
+///   b) There is at least a minimum width left after drawing the percentiles
```
hawkw: nit, take it or leave it: maybe we should mention what the minimum width is?

tokio-console/src/view/durations.rs
```diff
@@ -0,0 +1,104 @@
...
+            // then we won't draw the sparkline at all.
+            if area.width < percentiles_width + MIN_HISTOGRAM_BLOCK_WIDTH {
+                (area, None)
+            } else {
+                let areas = layout::Layout::default()
+                    .direction(layout::Direction::Horizontal)
+                    .constraints(
+                        [
+                            layout::Constraint::Length(percentiles_width),
+                            layout::Constraint::Min(50),
```
hawkw: why is the constraint here `Min(50)` instead of `Min(MIN_HISTOGRAM_BLOCK_WIDTH)`?

tokio-console/src/view/task.rs
```diff
@@ -224,133 +196,23 @@ impl TaskView {
...
             let warnings = List::new(warnings).block(styles.border_block().title("Warnings"));
             frame.render_widget(warnings, warnings_area);
         }
 
         let task_widget = Paragraph::new(overview).block(styles.border_block().title("Task"));
         let wakers_widget = Paragraph::new(waker_stats).block(styles.border_block().title("Waker"));
+        let poll_durations_widget = Durations::new(styles)
+            .histogram(details.and_then(|d| d.poll_times_histogram()))
+            .percentiles_title("Poll Times Percentiles")
+            .histogram_title("Poll Times Histogram");
```
hawkw, line 212: TIOLI: it occurs to me that the `Durations` widget _could_ just take a single title ("Poll Times") and append "Percentiles"/"Histogram" to that title value, unless we're ever planning on having a durations widget where the percentiles and the graphical histogram have totally different titles (which I don't think we're going to need any time soon...)

tokio-console/src/view/mod.rs
```diff
@@ -207,6 +209,13 @@ impl View {
     }
 }
 
+pub(crate) fn dur(styles: &Styles, dur: std::time::Duration) -> Span<'static> {
```
hawkw: i don't love making this internal helper public when it's just a small wrapper around `styles.time_units()` and there's no doc comment explaining what this does and how it's distinct from just calling `styles.time_units()`. i think this was less of a problem when it was an internal helper in one module where it was only used in that module.

could we add a doc comment or a clearer name to this so that its use is more obvious?

tokio-console/src/view/durations.rs
hds: Added.

tokio-console/src/view/durations.rs
hds: Missed that one. Thanks, I've fixed it now.

tokio-console/src/view/mod.rs
hds: Since this is only used in 3 places, I inlined the call to `time_units` and removed this function completely.

tokio-console/src/view/task.rs
hds: The reason I did it like this was to avoid having to construct (with an extra allocation) the title on each draw.

I don't know if tui's stateful widgets could help here, but having to call `format!("{title} Percentiles")` on each draw seemed unnecessary when I could specify the titles separately and only use static strings.

hawkw: Yeah, that's reasonable. Perhaps the `Durations` widget could own those allocated strings and format them when it's constructed, but I don't think it's worth blocking this PR on figuring that out. Current solution is fine!

hawkw (approved):
looks good to me, thank you!

PR #409, opened 2023-04-11
feat(console): add task scheduled times histogram

hds:
Display the scheduled time percentiles and sparklines for the histogram
of scheduled times. The schduled time is the time between when a task is
woken and when it is next polled.

The scheduled time, which was already calculated, is now stored in a
histogram and sent over the wire in together with the task details.

This is used to draw percentiles and sparklines on the task details
view, in the same way that is done for the poll times histogram.

The refactoring done in #408 has been used to more easily display two
sets of durations (percentiles and histogram where possible).

## PR Notes

The PR depends on both #406, which adds initial support for recording the
scheduled (wake-to-poll) time, and #408, which refactors the percentile and
histogram widgets to make them easier to reuse.

It shouldn't really be reviewed in depth until those two have been merged as
it contains a lot of duplication and will need to be rebased.

Here are some examples of the scheduled times durations on the task detail view:

task detail view for the sender task in the long-scheduled example

task detail view for the burn task in the app example

### DIFF
--- a/console-api/proto/common.proto
+++ b/console-api/proto/common.proto
@@ -161,6 +161,10 @@ message PollStats {
     // Subtracting this timestamp from `created_at` can be used to calculate the
     // time to first poll for this object, a measurement of executor latency.
     optional google.protobuf.Timestamp first_poll = 3;
+    // The timestamp of the most recent time this object was woken.
+    //
+    // If this is `None`, the object has not yet been woken.
+    optional google.protobuf.Timestamp last_wake = 7;
     // The timestamp of the most recent time this objects's poll method was invoked.
     //
     // If this is `None`, the object has not yet been polled.
@@ -180,8 +184,15 @@ message PollStats {
     // all polls. Note that this includes only polls that have completed and is
     // not reflecting any inprogress polls. Subtracting `busy_time` from the
     // total lifetime of the polled object results in the amount of time it
-    // has spent *waiting* to be polled.
+    // has spent *waiting* to be polled (including `scheduled_time`).
     google.protobuf.Duration busy_time = 6;
+    // The total duration this object was scheduled prior to being polled, summed
+    // across all poll cycles. Note that this includes only polls that have
+    // started and is not reflecting any scheduled state where the polling hasn't
+    // yet finished. Subtracting both `busy_time` and `scheduled_time` from the
+    // total lifetime of the polled object results in the amount of time it spent
+    // unable to progress because it was waiting on some resource.
+    google.protobuf.Duration scheduled_time = 8;
 }
 
 // State attributes of an entity. These are dependent on the type of the entity.
--- a/console-api/proto/common.proto
+++ b/console-api/proto/common.proto
@@ -161,10 +161,6 @@ message PollStats {
     // Subtracting this timestamp from `created_at` can be used to calculate the
     // time to first poll for this object, a measurement of executor latency.
     optional google.protobuf.Timestamp first_poll = 3;
-    // The timestamp of the most recent time this object was woken.
-    //
-    // If this is `None`, the object has not yet been woken.
-    optional google.protobuf.Timestamp last_wake = 7;
     // The timestamp of the most recent time this objects's poll method was invoked.
     //
     // If this is `None`, the object has not yet been polled.
@@ -189,13 +185,6 @@ message PollStats {
     // has spent *waiting* to be polled (including the `scheduled_time` value
     // from `TaskStats`, if this is a task).
     google.protobuf.Duration busy_time = 6;
-    // The total duration this object was scheduled prior to being polled, summed
-    // across all poll cycles. Note that this includes only polls that have
-    // started and is not reflecting any scheduled state where the polling hasn't
-    // yet finished. Subtracting both `busy_time` and `scheduled_time` from the
-    // total lifetime of the polled object results in the amount of time it spent
-    // unable to progress because it was waiting on some resource.
-    google.protobuf.Duration scheduled_time = 8;
 }
 
 // State attributes of an entity. These are dependent on the type of the entity.
--- a/console-api/proto/common.proto
+++ b/console-api/proto/common.proto
@@ -181,10 +181,13 @@ message PollStats {
     // its poll method has completed.
     optional google.protobuf.Timestamp last_poll_ended = 5;
     // The total duration this object was being *actively polled*, summed across
-    // all polls. Note that this includes only polls that have completed and is
-    // not reflecting any inprogress polls. Subtracting `busy_time` from the
+    // all polls.
+    //
+    // Note that this includes only polls that have completed, and does not
+    // reflect any in-progress polls. Subtracting `busy_time` from the
     // total lifetime of the polled object results in the amount of time it
-    // has spent *waiting* to be polled (including `scheduled_time`).
+    // has spent *waiting* to be polled (including the `scheduled_time` value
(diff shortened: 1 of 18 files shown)


hawkw:
This is kind of a silly naming bikeshed, but I'm not sure if I love the term "woken duration" for this measurement. It's not super clear to me whether this refers to the "duration from wake to poll" (which is what it's used to refer to here), the "duration from wake to yield, including the poll" (arguably, the task is still "woken" while it's being polled), or "the duration from yield _to_ wake" (which is what we refer to as "idle duration" but could be referred to as "duration until woken").

As an alternative naming suggestion, I'd propose "queued duration", since what we are measuring here is the amount of time that a woken task is held in a queue by the runtime before being polled.

hawkw:
Actually, I just took a look at `tokio-metrics`, and it looks like it [also exposes a similar metric, which they call `scheduled_duration`](https://docs.rs/tokio-metrics/0.2.1/tokio_metrics/struct.TaskMetrics.html#structfield.total_scheduled_duration). I think it's probably best to make our naming consistent with `tokio-metrics`, so let's call this "scheduled duration" instead of "woken duration"?

hds:
I agree on both counts, scheduled is a better name and if it's already used in tokio-metrics then best to align anyway.

I'll rework both these PRs and let you know when it's done.

If you haven't looked at the code too much, then I'll force push the update with a new commit message to keep the PR a bit neater.

hawkw:
I haven't looked super closely at this change yet, so feel free to force push! :)

hds:
**edit**: Not only did I name it wrong, but I had the wrong description in as well. *facepalm*
**edit edit**: Both #406 (which should be reviewed first) and this PR are ready now.

@hawkw I only realized later that this is the second PR in the set (I named it wrong, so both PRs had the same name).

I've now updated the 1st PR (#406) to the new naming (scheduled), please review that one first. In the meantime, I'll update the naming on this one too. I've marked it as a draft in the meantime.

hawkw:
Okay, #406 has merged, and I'll be excited to review this as soon as it's rebased! :)

hds:
Great, thanks! I've also got to adapt the wire format to the changes we made in #406, hopefully sone tomorrow!

hawkw:
CI lints failure seems unrelated (and looks wacky): https://github.com/tokio-rs/console/actions/runs/4756530428/jobs/8452202802?pr=409

console-api/proto/common.proto
```diff
@@ -161,6 +161,10 @@ message PollStats {
     // Subtracting this timestamp from `created_at` can be used to calculate the
     // time to first poll for this object, a measurement of executor latency.
     optional google.protobuf.Timestamp first_poll = 3;
+    // The timestamp of the most recent time this object was woken.
+    //
+    // If this is `None`, the object has not yet been woken.
+    optional google.protobuf.Timestamp last_wake = 7;
```
hawkw: looks like this needs to be removed, since we changed #406 to add this to `TaskStats` instead?

console-api/proto/common.proto
```diff
@@ -185,6 +189,13 @@ message PollStats {
     // has spent *waiting* to be polled (including the `scheduled_time` value
     // from `TaskStats`, if this is a task).
     google.protobuf.Duration busy_time = 6;
+    // The total duration this object was scheduled prior to being polled, summed
+    // across all poll cycles. Note that this includes only polls that have
+    // started and is not reflecting any scheduled state where the polling hasn't
+    // yet finished. Subtracting both `busy_time` and `scheduled_time` from the
+    // total lifetime of the polled object results in the amount of time it spent
+    // unable to progress because it was waiting on some resource.
+    google.protobuf.Duration scheduled_time = 8;
```
hawkw: similarly, looks like this needs to be removed, since we changed #406 to add this to `TaskStats` instead?

hds:
@hawkw this is ready for review now.

The CI failure is a new clippy lint in 1.69 affecting `clap::Subcommand`. I'll try and fix that in a separate PR.

tokio-console/src/view/task.rs
```diff
@@ -207,16 +225,32 @@ impl TaskView {
...
         let task_widget = Paragraph::new(overview).block(styles.border_block().title("Task"));
         let wakers_widget = Paragraph::new(waker_stats).block(styles.border_block().title("Waker"));
+
+        let poll_percentiles_title = "Poll Times Percentiles";
+        let scheduled_percentiles_title = "Sched Times Percentiles";
+        let percentiles_width = cmp::max(
+            poll_percentiles_title.len(),
+            scheduled_percentiles_title.len(),
+        ) as u16
+            + 2_u16;
```
hawkw: nit: it would be nice if there was a comment here explaining what the extra 2 characters of width is for? i'm assuming it's for borders?

tokio-console/src/view/durations.rs
```diff
@@ -36,14 +36,22 @@ pub(crate) struct Durations<'a> {
...
         // sparkline. We require UTF-8 to draw the sparkline and also enough width.
         let (percentiles_area, histogram_area) = if self.styles.utf8 {
-            let percentiles_width = cmp::max(self.percentiles_title.len() as u16, 13_u16) + 2;
+            let percentiles_width = match self.percentiles_width {
+                // Fixed width
+                width if width > 0 => width,
+                // Long enough for the title or for a single line
+                // like "p99: 544.77µs" (13) (and borders on the sides).
+                _ => cmp::max(self.percentiles_title.len() as u16, 13_u16) + 2,
+            };
```
hawkw, line 54: Where is the Durations widget currently constructed with a width of 0? Is that used for the resource details view, where we only have one durations widget?

hawkw (changes requested):
Overall, this looks great. I think an accidental protobuf change may have slipped in while rebasing after #406 merged, so we'll need to undo that. Other than that, I had a few small questions, but no other blockers.

console-api/proto/common.proto
hds: I seem to have ended up with `last_wake` and `scheduled_time` twice somehow. Sorry.

console-api/proto/common.proto
hds: I seem to have ended up with `last_wake` and `scheduled_time` twice somehow. Sorry.

tokio-console/src/view/durations.rs
hds: Actually (and I had to go and check this to be sure), the resource details view doesn't have a Durations widget currently (although it will be much easier to add after the refactoring). The `width=0` value here is set as a sensible default if a fixed width isn't set.

The fixed width was introduced in this PR so that the percentiles and sparklines widgets are the same width for both the Durations widgets irrespective of the title length (so that it looks nicer).

tokio-console/src/view/task.rs
hds: Done.

tokio-console/src/view/durations.rs
hawkw: It seems like we could change the constructor to require that a non-zero width is always provided, and remove this behavior, if we intend to always construct the widget with a fixed width? This might be a bit simpler. Not a blocker though...

hawkw (approved):

hawkw:
I'm going to go ahead and merge this, we can address potential refactors to the way the `Durations` widget's width is configured in a follow-up (if you want to --- it's also fine to just ignore that suggestion!).

PR #414, opened 2023-04-26
fix: work around Rust 1.69 `almost_swapped` lint 

hds:
The enhanced `clippy::almost_swapped` lint which shipped in Rust 1.69.0
was triggering in some cases where we derive `clap::Subcommand`. It
wasn't possible to allow it, because a deny on `clippy::correctness`
(which implies `clippy::almost_swapped`) was present in the generated
code.

As a result, we had to fix the version of Rust used for Clippy to
1.68.2.

In clap_derive (included in clap) 3.2.25, the deny on `clippy::correctness`
has been removed and an allow for `clippy::almost_swapped` has beed
added. Fixing the forced clippy error.

This change updates clap to the latest 3.x version, including
clap_derive and clap_lex.

It also reverts the change made in #413, so we will go back to using
the latest version of Rust for `clippy` and `fmt`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,10 +25,6 @@ env:
   RUSTUP_MAX_RETRIES: 10
   # Don't emit giant backtraces in the CI logs.
   RUST_BACKTRACE: short
-  # Use Rust 1.68.2 for Clippy (and fmt) because of an enhanced `almost_swapped`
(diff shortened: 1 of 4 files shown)


hds:
@hawkw This change is ready to go in. It updates clap_derive in the Cargo.lock file and fully reverts #413 so we're back to using the latest Rust version for clippy and fmt.

hawkw (approved):

PR #419, opened 2023-05-09
docs(console): update screenshots in README

hds:
The screenshots in the README, which are also used in in the
`tokio-console` page on docs.rs were a bit outdated. The UI has changed
a bit since they were created.

This change updates the screenshots to be based on the latest UI from
the `main` branch.


tokio-console/README.md
```diff
@@ -122,13 +122,13 @@ Like the task list view, the resource list view can be navigated using the
 <kbd>&#8593;</kbd> and <kbd>&#8595;</kbd> arrow keys. Pressing <kbd>enter</kbd>
 while a resource is highlighted displays details about that resource:
 
-![resource details --- oneshot](https://raw.githubusercontent.com/tokio-rs/console/main/assets/resource_details1.png)
+![resource details --- oneshot](https://raw.githubusercontent.com/tokio-rs/console/main/assets/resource_details_sleep.png)
```
hawkw: this now shows a sleep, should we change this?
```suggestion
![resource details --- sleep](https://raw.githubusercontent.com/tokio-rs/console/main/assets/resource_details_sleep.png)
```

hawkw (commented):
one potential note about this: your new screenshots are substantially wider than the previous ones (2560px vs 1250px). when i was taking the original screenshots, i tried to keep the size of the terminal relatively small, so that the docs would still look nice in smaller browser windows.

we should probably make sure the new screenshots still look okay when the browser window is smaller.i was going to suggest checking out the [netlify deploy preview](https://deploy-preview-419--tokio-console.netlify.app/tokio_console/index.html) for the new docs, but it doesn't actually show the new screenshots, since they are referenced from the main branch and we haven't merged this PR yet...hmm...

hawkw (approved):
i'm gonna go ahead and get this merged so we can see how it looks!

PR #421, opened 2023-05-09
docs(console): update screenshots in README

hds:
The screenshots in the README, which are also used in in the
`tokio-console` page on docs.rs were a bit outdated. The UI has changed
a bit since they were created.

This change updates the screenshots to be based on the latest UI from
the `main` branch.

All previous versions of `tokio-console` on docs.rs reference the images
on the main branch, so it is important that they remain unchanged. The
new screenshots have been added to a sub-directory of the `assets`
directory named after the next version: `tokio-console-0.1.8`.


hawkw (approved):
these look great! thank you!!

PR #424, opened 2023-05-17
fix(console): remove histogram minimum count

hds:
The sparklines histogram used to visualize poll times and scheduled
times renders a legend for both axes. On the horizontal axis, it
displays the durations for the minimum and maximum buckets. On the
vertical axis, it displays the bucket count minimum and maximums.

Unfortunately, the minimum values for each axis are places together,
which can be confusing:

```text
╭Poll Times Histogram─────────────────────────────────╮
│46█                                                  │
│  █                                                  │
│  █                                                  │
│  █                                                  │
│  █ ▂                                                │
│  █▁█▇▃▁▁           ▁     ▁▁                       ▁ │
│ 031.23µs                                      7.11ms│
╰─────────────────────────────────────────────────────╯
  ^|----|
  │  ╰─ This is the minimum bucket duration value
  ╰── This is the minimum count
```

This can be confusing because it looks like there is an erroneous
leading zero on the minimum bucket duration.

If the minimum value is not zero, then the vertical axis minimum is
updated, but the bottom of the sparkline still represents zero. This
sitaution is unlikely, I had to add a fixed value to the bucket counts
to generate the histogram below.

```text
╭Poll Times Histogram─────────────────────────────────╮
│32██                                                 │
│  ██▇    ▁               ▁   ▁  ▁ ▁                  │
│  ███▇▆▆▆█▇▆▇▆▆▆▆▆▆▆▇▆▆▆▆█▆▇▆█▆▇█▆█▇▇▇▆▇▇▆▇▆▆▇▆▆▆▆▆▇ │
│  ██████████████████████████████████████████████████ │
│  ██████████████████████████████████████████████████ │
│  ██████████████████████████████████████████████████ │
│2033.79µs                                   536.58µs │
╰─────────────────────────────────────────────────────╯
 ^^|----|
  │  ╰─ This is the minimum bucket duration value
  ╰── This is the minimum count
```

This is probably misleading.

This change removes the minimum label on the vertical axis entirely.
This removes the confusion around the "erroneous" leading zero and since
the bottom of the graph is always zero anyway, it should be clear to
users without having to add the label.

The resulting histogram looks like:

```text
╭Poll Times Histogram─────────────────────────────────╮
│27 █▄                                                │
│   ██                                                │
│   ██                                                │
│   ██                                                │
│   ██▂                                               │
│  ▃███▃▁ ▁      ▁▃ ▃▃▃▁▃▅▁▃ ▃▃▃▃▃▁▁▁   ▁  ▁         ▁│
│  28.93µs                                   811.01µs │
╰─────────────────────────────────────────────────────╯
```

### DIFF
--- a/tokio-console/src/view/mini_histogram.rs
+++ b/tokio-console/src/view/mini_histogram.rs
@@ -39,8 +39,6 @@ pub(crate) struct HistogramMetadata {
     pub(crate) min_value: u64,
     /// The value of the bucket with the greatest quantity
     pub(crate) max_bucket: u64,
-    /// The value of the bucket with the smallest quantity
-    pub(crate) min_bucket: u64,
     /// Number of high outliers, if any
     pub(crate) high_outliers: u64,
     pub(crate) highest_outlier: Option<Duration>,
@@ -87,7 +85,6 @@ impl<'a> Widget for MiniHistogram<'a> {
         };
 
         let max_qty_label = metadata.max_bucket.to_string();
-        let min_qty_label = metadata.min_bucket.to_string();
         let max_record_label = format!(
             "{:.prec$?}",
             Duration::from_nanos(metadata.max_value),
@@ -107,7 +104,6 @@ impl<'a> Widget for MiniHistogram<'a> {
             max_record_label,
             min_record_label,
             max_qty_label,
-            min_qty_label,
         );
 
         let legend_height = if metadata.high_outliers > 0 { 2 } else { 1 };
@@ -229,7 +225,6 @@ fn render_legend(
     max_record_label: String,
     min_record_label: String,
     max_qty_label: String,
-    min_qty_label: String,
 ) {
     // If there are outliers, display a note
     let labels_pos = if metadata.high_outliers > 0 {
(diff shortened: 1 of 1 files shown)


hawkw (approved):
Thanks, this definitely looks less confusing!

Another potential solution, if we did want to display the minimum value, would be to move the axis labels _outside_ of the bounding box we draw around the histogram, so that it looks like this:
```text
  ╭Poll Times Histogram───────────────────────────────╮
46│█                                                  │
  │█                                                  │
  │█                                                  │
  │█ ▂                                                │
  │█▁█▇▃▁▁           ▁     ▁▁                       ▁ │
 0╰───────────────────────────────────────────────────╯
   31.23µs                                          7.11ms
```
I think this would probably be substantially more work, though, and the current solution already solves a lot of the potential ambiguity, so I'm happy to merge this as-is and if we want to sink more time into making it fancier, we can do that later!

hds:
I think that experimenting with where we place the labels would be interesting. We'd have to make sure that they don't end up confusing things with the other widgets' bounding boxes.

PR #425, opened 2023-05-21
chore: migrate to ratatui and update crossterm

joshka:
See https://github.com/tui-rs-revival/ratatui/wiki/Migrating-from-TUI

[asciicast](https://asciinema.org/a/ZUC4m05kaKJu7aSgdQb4gv4F9)

### DIFF
--- a/tokio-console/Cargo.toml
+++ b/tokio-console/Cargo.toml
@@ -33,13 +33,13 @@ clap = { version = "3", features = ["cargo", "derive", "env"] }
 tokio = { version = "1", features = ["full", "rt-multi-thread"] }
 tonic = { version = "0.9", features = ["transport"] }
 futures = "0.3"
-tui = { version = "0.16.0", default-features = false, features = ["crossterm"] }
+ratatui = { version = "0.20.1", default-features = false, features = ["crossterm"] }
 tower = "0.4.12"
 tracing = "0.1"
 tracing-subscriber = { version = "0.3.0", features = ["env-filter"] }
 tracing-journald = { version = "0.2", optional = true }
 prost-types = "0.11"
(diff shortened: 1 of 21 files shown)


tokio-console/Cargo.toml
```diff
@@ -33,13 +33,13 @@ clap = { version = "3", features = ["cargo", "derive", "env"] }
 tokio = { version = "1", features = ["full", "rt-multi-thread"] }
 tonic = { version = "0.9", features = ["transport"] }
 futures = "0.3"
-tui = { version = "0.16.0", default-features = false, features = ["crossterm"] }
+tui = { package = "ratatui", version = "0.20.1", default-features = false, features = ["crossterm"] }
```
hds: We'd prefer to fully switch to `ratatui`, rather than using it under the name of `tui`.

```suggestion
ratatui = { version = "0.20.1", default-features = false, features = ["crossterm"] }
```

hds (changes requested):
Thank you very much for your PR!

If we're going to switch to `ratatui`, which seems to be the right direction to move in. Then we'd prefer to perform a full switch and import the `ratatui` package directly, rather than under the `tui` alias.

If you don't mind making the changes to do so, that would be awesome! Otherwise, we'll take over the change. Thanks again for the interest!

tokio-console/Cargo.toml
joshka: Done

joshka:
Renamed all tui refereces to ratatui in latest push

[asciicast](https://asciinema.org/a/9Ul5gkS1Ob8dkUrByhhGdKaRd)

hds (approved):
Looks good. Thank you for making the changes!

PR #427, opened 2023-05-25
refac(console): generalize controls widget

hds:
Each view in tokio-console has a widget up the top that lists the
available controls for that view. There was a common implementation of
this for table based views (tasks, resources, and async_ops) and
separate implementations for the task and resource views. The resource
view included two controls widgets, one for the Resource details at the
top of the view, and another above the table of Async Ops at the bottom
of the view.

This change centralises the logic for the creation of this controls
widget. This change is mostly a precursor to also displaying the
controls in the help view (at which point we can revisit whether the
entire list needs to be shown at the top of the screen).

Controls (an action and the key or keys used to invoke it) are defined
in structs so that their definition can be separated from the display
logic (which includes whether or not UTF-8 is supported).

This allows the problem of the text in the controls widget wrapping in
the middle of a control definition to be fixed. Previously a subset of
the controls would have wrapped like this:

```text
controls: select column (sort) = ←→ or h, l, scroll = ↑↓ or k, j, view details
= ↵, invert sort (highest/lowest) = i,
```

Notice how "view details = ↵," was split across multiple lines. The same
list of controls will now wrap at a full control definition.

```text
controls: select column (sort) = ←→ or h, l, scroll = ↑↓ or k, j,
view details = ↵, invert sort (highest/lowest) = i,
```

Additionally, the list of controls on the Resource view has been
consolidated up the top of the screen.

Universal controls, those that are available in all views, are also
defined centrally. As well as the quit action, using the space bar to
pause has been added to that list. This was previously somewhat of an
undocumented feature.

### DIFF
--- a/tokio-console/src/view/controls.rs
+++ b/tokio-console/src/view/controls.rs
@@ -1,6 +1,5 @@
 use crate::view::{self, bold};
 
-use once_cell::sync::OnceCell;
 use tui::{
     layout,
     text::{Span, Spans, Text},
@@ -16,7 +15,7 @@ pub(crate) struct Controls {
 
 impl Controls {
     pub(in crate::view) fn new(
-        view_controls: &Vec<ControlDisplay>,
+        view_controls: &'static [ControlDisplay],
         area: &layout::Rect,
         styles: &view::Styles,
     ) -> Self {
@@ -32,28 +31,39 @@ impl Controls {
 
         let controls_count: usize = spans_controls.len();
         for (idx, spans) in spans_controls.into_iter().enumerate() {
+            // If this is the first item on this line - or first item on the
+            // first line, then always include it - even if it goes beyond the
+            // line width, not much we can do anyway.
             if idx == 0 || current_line.width() == 0 {
                 current_line.0.extend(spans.0);
-            } else {
-                let needed_trailing_separator_width = if idx == controls_count + 1 {
-                    separator.width()
-                } else {
-                    0
-                };
+                continue;
+            }
 
-                let total_width = current_line.width()
-                    + separator.width()
-                    + spans.width()
-                    + needed_trailing_separator_width;
-                    
-                if total_width <= area.width as usize {
-                    current_line.0.push(separator.clone());
-                    current_line.0.extend(spans.0);
-                } else {
-                    current_line.0.push(separator.clone());
-                    lines.push(spans);
-                    current_line = lines.last_mut().expect("This vector is never empty");
-                }
+            // Include the width of our separator in the current item if we
+            // aren't placing the last item. This is the separator after the
+            // new element.
+            let needed_trailing_separator_width = if idx == controls_count + 1 {
+                separator.width()
+            } else {
+                0
+            };
+
+            let total_width = current_line.width()
+                + separator.width()
+                + spans.width()
+                + needed_trailing_separator_width;
+
+            // If the current item fits on this line, append it.
+            // Otherwise, append only the separator - we accounted for its
+            // width in the previous loop iteration - and then create a new
+            // line for the current item.
+            if total_width <= area.width as usize {
+                current_line.0.push(separator.clone());
+                current_line.0.extend(spans.0);
+            } else {
+                current_line.0.push(separator.clone());
+                lines.push(spans);
+                current_line = lines.last_mut().expect("This vector is never empty");
             }
         }
 
@@ -99,17 +109,7 @@ pub(crate) struct KeyDisplay {
 }
 
 impl ControlDisplay {
-    pub(crate) fn new_simple(action: &'static str, key: &'static str) -> Self {
-        ControlDisplay {
-            action,
-            keys: vec![KeyDisplay {
-                base: key,
-                utf8: None,
-            }],
-        }
-    }
-
-    pub fn to_spans(&self, styles: &view::Styles) -> Spans<'static> {
+    pub(crate) fn to_spans(&self, styles: &view::Styles) -> Spans<'static> {
         let mut spans = Vec::new();
 
         spans.push(Span::from(self.action));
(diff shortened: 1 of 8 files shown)


tokio-console/src/view/controls.rs
```diff
@@ -0,0 +1,141 @@
...
+        let mut spans_controls = Vec::with_capacity(view_controls.len() + universal_controls.len());
+        spans_controls.extend(view_controls.iter().map(|c| c.to_spans(styles)));
+        spans_controls.extend(universal_controls.iter().map(|c| c.to_spans(styles)));
+
+        let mut lines = vec![Spans::from(vec![Span::from("controls: ")])];
+        let mut current_line = lines.last_mut().expect("This vector is never empty");
+        let separator = Span::from(", ");
+
+        let controls_count: usize = spans_controls.len();
+        for (idx, spans) in spans_controls.into_iter().enumerate() {
```
hawkw, line 49: nit, take it or leave it: i might like to see some documentation explaining this wrapping logic?

tokio-console/src/view/controls.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// Construct span to display a control.
+///
+/// A control is made up of an action and one or more keys that will trigger
+/// that action.
+#[derive(Clone)]
+pub(crate) struct ControlDisplay {
+    pub(crate) action: &'static str,
+    pub(crate) keys: Vec<KeyDisplay>,
+}
```
hawkw: take it or leave it: it seems like we could allow these to be statically constructed with a `const fn` constructor if we changed this to
```suggestion
pub(crate) struct ControlDisplay {
    pub(crate) action: &'static str,
    pub(crate) keys: &'static [KeyDisplay],
}
```
since there's no actual need to construct them dynamically?

tokio-console/src/view/controls.rs
```diff
@@ -0,0 +1,141 @@
...
+pub(crate) fn universal_controls() -> &'static Vec<ControlDisplay> {
+    static UNIVERSAL_CONTROLS: OnceCell<Vec<ControlDisplay>> = OnceCell::new();
+
+    UNIVERSAL_CONTROLS.get_or_init(|| {
+        vec![
+            ControlDisplay::new_simple("toggle pause", "space"),
+            ControlDisplay::new_simple("quit", "q"),
+        ]
+    })
+}
```
hawkw: if we changed the `ControlDisplay` type to own an `&'static [KeyDisplay]` rather than a `Vec`, like I described above, `UNIVERSAL_CONTROLS` could just be a constant, rather than needing to be lazily initialized.

tokio-console/src/view/controls.rs
```diff
@@ -0,0 +1,141 @@
...
+                } else {
+                    0
+                };
+
+                if current_line.width()
+                    + separator.width()
+                    + spans.width()
+                    + needed_trailing_separator_width
+                    <= area.width as usize
+                {
```
hawkw: style nit, take it or leave it: i might consider
```suggestion
                let total_width = current_line.width()
                    + separator.width()
                    + spans.width()
                    + needed_trailing_separator_width;
                    
                if total_width <= area.width as usize {
```
but, not a big deal.

tokio-console/src/view/controls.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        let mut lines = vec![Spans::from(vec![Span::from("controls: ")])];
+        let mut current_line = lines.last_mut().expect("This vector is never empty");
+        let separator = Span::from(", ");
+
+        let controls_count: usize = spans_controls.len();
+        for (idx, spans) in spans_controls.into_iter().enumerate() {
+            if idx == 0 || current_line.width() == 0 {
+                current_line.0.extend(spans.0);
+            } else {
```
hawkw, line 79: style nit, take it or leave it: i might prefer
```suggestion
                current_line.0.extend(spans.0);
                continue;
            }
```
here, so that the majority of the code in this loop, which is currently an `else` block, doesn't need to be indented? but, this is up to you.

hawkw (approved):
thanks, this looks good to me! i had some very minor suggestions, but they're not blockers for merging this!

hawkw (approved):
thanks, this looks good to me! i had some very minor suggestions, but they're not blockers for merging this!

tokio-console/src/view/controls.rs
hds: Done.

tokio-console/src/view/controls.rs
hds: Done (I didn't apply the suggestion because I had to re-indent everything anyway).

tokio-console/src/view/controls.rs
hds: Done. Although in resource.rs I had to keep the `OnceCell` because I'm concatenating 2 arrays which it seems can't be dont in `const`.

If you know a way of doing this, that would be really cool.

tokio-console/src/view/controls.rs
```diff
@@ -129,13 +129,21 @@ impl ControlDisplay {
...
-pub(crate) fn universal_controls() -> &'static Vec<ControlDisplay> {
-    static UNIVERSAL_CONTROLS: OnceCell<Vec<ControlDisplay>> = OnceCell::new();
-
-    UNIVERSAL_CONTROLS.get_or_init(|| {
-        vec![
-            ControlDisplay::new_simple("toggle pause", "space"),
-            ControlDisplay::new_simple("quit", "q"),
-        ]
-    })
+const fn universal_controls() -> &'static [ControlDisplay] {
```
hawkw: nit, not a big deal: can't this just be 
`const UNIVERSAL_CONTROLS: &'static [ControlDisplay]`?

hawkw (approved):
looks good to me!

tokio-console/src/view/controls.rs
hds: Yes it can! Changed.

PR #430, opened 2023-06-01
fix(subscriber): fix self wakes count

### LINKED ISSUE #423: Self wakes are not detected

_No response_

### What is the issue?

Self wakes are not being detected by tokio console.

The `burn` task in the `app` example should result in self wakes, because it calls `tokio::task::yield_now()` from within the task.

```rust
#[tracing::instrument]
async fn burn(min: u64, max: u64) {
    loop {
        for i in min..max {
            for _ in 0..i {
            }
            tokio::time::sleep(Duration::from_secs(i - min)).await;
        }
    }
}
```

Looking at the logic of `yield_now()`, it seems that it may not actually result in a self wake (wake being called from within the span of the task being woken). So to check, I replaced the call to `yield_now()` with a call to a `simple_yield_now()` which **should** always wake within the span of the task:

```rust
async fn simple_yield_now() {
    pub struct YieldNow {
        yielded: bool,
    }

    impl Future for YieldNow {
        type Output = ();

        fn poll(
            mut self: std::pin::Pin<&mut Self>,
            cx: &mut std::task::Context<'_>,
        ) -> Poll<Self::Output> {
            if self.yielded == true {
                return Poll::Ready(());
            }

            self.yielded = true;
            cx.waker().wake_by_ref();

            Poll::Pending
        }
    }

    YieldNow { yielded: false }.await
}
```

However, tokio-console still fails to report a self wake.

### Logs, error output, etc

We the following in the traces:

```
2023-05-15T14:07:54.387864Z TRACE runtime.spawn{kind=task task.name=burn task.id=19 loc.file="console-subscriber/examples/self-wake.rs" loc.line=11 loc.col=10}:burn{min=1 max=10}: tokio::task::waker: op="waker.wake_by_ref" task.id=1
```

Unfortunately, the `task.id` field in `runtime.spawn` span refers to the Tokio Task ID of the task, whereas the `task.id` in the `waker` event is the Span ID of the `runtime.spawn` span. So it's not immediately clear why

### Versions

* `tokio-console` v0.1.8
* `console-subscriber` v0.1.9

It's also occuring on `main` at the time of writing [cbf6f5](https://github.com/tokio-rs/console/commit/cbf6f56a16036ecf13548c4209fcc62f8a84bae2).

### Possible solution

_No response_

### Additional context

_No response_

yes

hds:
Self-wakes were not being detected and displayed in the console. The
`burn` task in the `app` example - which deliberately has many
self-wakes - was not registering any.

It appears that a logic error was present in the self-wake counting in
`console-subscriber` since it was added in #238. When a self wake was
detected, the `wakes` count was incremented a second time (the `wakes`
count is incremented for all wakes before checking for a self wake),
instead of increamenting the `self_wakes` count.

This PR fixes the logic so that when a self wake is detected, the
`self_wakes` count is incremented.

Fixes: #423

### DIFF
--- a/console-subscriber/src/stats.rs
+++ b/console-subscriber/src/stats.rs
@@ -217,7 +217,7 @@ impl TaskStats {
 
         self.wakes.fetch_add(1, Release);
         if self_wake {
-            self.wakes.fetch_add(1, Release);
+            self.self_wakes.fetch_add(1, Release);
         }
 
         self.make_dirty();


hawkw (approved):
good catch, looks good to me!

I wonder if it would be possible to have a test for this...

hds:
Testing is one thing that a lot of console lacks at the moment I think. It should be possible to test with synthetic logs for very specific cases (so as to make the tests fairly easy to read and write).

hawkw:
Yeah...I'm not going to block this PR on testing, but it could be nice to figure out eventually.

PR #434, opened 2023-06-08
console_subscriber: remove clock skew warning in start_poll

### LINKED ISSUE #433: console_subscriber: "poll's start timestamp was before the wake time/last poll timestamp"

console-subscriber

### What is the issue?

console subscriber spams "possible Instant skew" message.

Running a relatively complex project on a multithreaded runtime. I suspect the error happens when a task is moved between threads.

This is happening on an intel platform with big and little cores.

### Logs, error output, etc

```
possible Instant clock skew detected: a poll's start timestamp was before the wake time/last poll end timestamp                                                                                                                                                                                                               
wake = Instant { tv_sec: 68792, tv_nsec: 171614373 }                                                                                                                                                                                                                                                                          
  start = Instant { tv_sec: 68792, tv_nsec: 171610785 }
```

### Versions

Current git version (ea006945b790ed01b1bda54068a0200333495d54). v0.1.9 as well. Anything past https://github.com/tokio-rs/console/commit/ac20daaf301f80e87002593813965d11d11371e4

### Possible solution

Remove symmetric `eprintln!` from [here](https://github.com/tokio-rs/console/commit/ac20daaf301f80e87002593813965d11d11371e4#diff-0d2dd8f5e2f895fa5c5a9f2083c93d0ce352cf0cbee76b62ccf7a218ad3c5f42R507-R517). Unlike the warning in `end_poll`, this warning can be triggered in regular situations on multithreaded runtime, because tasks may be moved across threads in-between the polls.

One alternative we may want to consider is using saturating subtraction, which would lead to slightly inaccurate results (lower runtime?), but would not drop the sample.

### Additional context

This is not the same as https://github.com/tokio-rs/console/issues/299, this is coming from a different line.

This bug was introduced in [here](https://github.com/tokio-rs/console/commit/ac20daaf301f80e87002593813965d11d11371e4#diff-0d2dd8f5e2f895fa5c5a9f2083c93d0ce352cf0cbee76b62ccf7a218ad3c5f42R507-R517)

yes

h33p:
Remove "possible Instant clock skew" message in start_poll. This warning could occur when a task switches threads between 2 polls. Replaced the code to use saturating_duration_since, which clamps the elapsed time to 0, if the starting poll time is less than the scheduled time.

Fixes #433

### DIFF
--- a/console-subscriber/src/stats.rs
+++ b/console-subscriber/src/stats.rs
@@ -518,17 +518,8 @@ impl<H: RecordDuration> PollStats<H> {
             None => return, // Async operations record polls, but not wakes
         };
 
-        let elapsed = match at.checked_duration_since(scheduled) {
-            Some(elapsed) => elapsed,
-            None => {
-                eprintln!(
-                    "possible Instant clock skew detected: a poll's start timestamp \
-                    was before the wake time/last poll end timestamp\nwake = {:?}\n  start = {:?}",
-                    scheduled, at
-                );
-                return;
-            }
-        };
+        // `at < scheduled` is possible when a task switches threads between polls.
+        let elapsed = at.saturating_duration_since(scheduled);
 
         // if we have a scheduled time histogram, add the timestamp
         timestamps.scheduled_histogram.record_duration(elapsed);


hawkw (approved):
looks good to me, thank you!

Issue #203, opened 2021-12-15
Report when a task doesn't / hasn't yielded in a while.


tobz:
While using `tokio-console` for debugging some weird behavior -- by the way, `tokio-console` has been great so far! -- I ended up discovering that my original issue was related to a task that never yielded.  While `tokio-console` helped me figure this out by displaying that a companion task hadn't yet been woken up (because of the other task not yielding), it would have saved me a step or two if instead I was able to see that the non-yielding task hadn't yielded in a while/at all.

Not sure if this is simple to surface or not, but figured I'd file an issue for it regardless. :)

hawkw:
This should be pretty easy. I think the only difficult part is that we'll have to determine how long a task should have run before yielding before adding a warning.

hawkw:
@tobz if you're interested in working on this, the way we'd do it is adding a new type implementing the `Warn` trait: https://github.com/tokio-rs/console/blob/7d16eadc5c254f21b0f4fba31f2fdf758808a8f4/console/src/warnings.rs#L4-L47

This would be pretty similar to existing warnings, like `SelfWakePercent`: https://github.com/tokio-rs/console/blob/7d16eadc5c254f21b0f4fba31f2fdf758808a8f4/console/src/warnings.rs#L118-L135

I think that in the check method for this, we'd want to flag a task for the warning if *all* of the following are true:

1. it is currently active (so the warning doesn't fire for tasks that are waiting to run)
2. it has never yielded
3. it has been running for a period of time >= some threshold (configurable?)

hawkw:
we probably want to be _fairly_ generous with the threshold, to avoid false positives...it could be something like 5-10 ms? I don't think we want the lint to fire on particularly short-lived tasks; if a task only has two or three await points in it, it's entirely possible that one instance of that task might just be lucky enough that they're all ready, and that's _fine_ if it runs for like, 1ms or less!

hawkw:
@tobz still planning on working on this?

tobz:
As interesting as it may be, I think my spare time is a bit too small at
the moment to take this. :(

fbrv:
Hi there,
I'm learning Tokio and Rust and I think this issue could be interesting for me, if no one else is working on that I could give it a try

hawkw:
@fbrv sorry i didn't reply sooner --- are you still interested in working on this issue? If so, please feel free to open a PR, or let me know if you have any questions!

rybertm:
[...]

PR #439, opened 2023-06-22
feat(console): add warning for tasks that never yield

jefftt:
This PR only addresses the _never yielding_ part of https://github.com/tokio-rs/console/issues/203 wasn't sure if the _hasn't yielded in a long time_ part should be in the same warning or not.


tokio-console/src/warnings.rs
```diff
@@ -150,3 +154,56 @@ impl Warn<Task> for LostWaker {
...
+
+/// Warning for if a task has never yielded
+#[derive(Clone, Debug)]
+pub(crate) struct NeverYielded {
+    min_duration: Duration,
+    description: String,
+}
+
+impl NeverYielded {
+    pub(crate) const DEFAULT_DURATION: Duration = Duration::from_millis(10);
```
hds: I think that there are probably many users who have tasks which take more than 10 milliseconds to yield the first time. I would suggest that we are conservative here and not trigger this lint until a task has been running for 1 second and not yielded.

This also matches the default for considering a poll time as an outlier.

```suggestion
    pub(crate) const DEFAULT_DURATION: Duration = Duration::from_secs(1);
```

console-subscriber/examples/busy_loop.rs
```diff
@@ -0,0 +1,31 @@
+use std::sync::atomic::Ordering;
```
hds: I propose that this example gets merged into the `app` example as an option.

It's possible to work together with the existing examples.

Instead of using an actual busy loop, I suggest we use `std::time::sleep` in a loop. In async code, the std library sleep effectively emulates a task being busy, but it will do so without overloading the machine and allows fine grained control for how long the "work" lasts.

For example, the following code could be used with the `app` example to periodically spawn a task that takes `seconds` to yield (perhaps try 20) the first time (and then the task ends and another one is spawned).

```rust
#[tracing::instrument]
async fn no_yield(seconds: u64) {
    loop {
        let handle = tokio::task::Builder::new()
            .name("greedy")
            .spawn(async move {
                std::thread::sleep(Duration::from_secs(seconds));
            })
            .expect("Couldn't spawn greedy task");

        _ = handle.await;
    }
}
```

tokio-console/src/warnings.rs
```diff
@@ -150,3 +154,56 @@ impl Warn<Task> for LostWaker {
...
+        self.description.as_str()
+    }
+
+    fn check(&self, task: &Task) -> bool {
+        // Don't fire warning for tasks that are waiting to run
+        if task.state() != TaskState::Running {
+            return false;
+        }
+
+        if task.yielded() {
```
hds: I would inline the logic from `task.yielded()` here and then we don't need that method at all.

```suggestion
        if task.total_polls() > 1 {
```

hds (changes requested):
Thanks for your PR!

I agree with the approach to splitting #203 into 2 separate lints. One for tasks that have never yielded (this PR) and another for tasks that have long poll times (but have yielded at least once). That's a good idea!

I've got a few suggestions for changes here. Would you be able to have a look at them?

tokio-console/src/warnings.rs
hawkw: eventually we should really figure out a nice way to add configuration values for lints to the config file...

jefftt:
Thanks for the helpful review! Changes have been applied

hds (changes requested):
Thanks for the changes!

I tested this locally and found an issue. The lint isn't always being run at the right time to catch the "never yielded" state. This is because lints only get run when a task receives an update.

See this video of my test: 

https://github.com/tokio-rs/console/assets/89589/283426c8-313a-4005-bad8-c07605dcbfd5

The first time you start tokio-console (using the `app` example), the lint will be run (because all tasks are received and this counts as an update). Since 1 second has probably already passed,`check()` will return `true` and we see the warning.

However, when the greedy task gets spawned again (which is what would happen if a task gets spawned after tokio-console has already started running), it only gets checked on a state transition. So you would see a check when it transitions to `Running` and another when it transitions to state `Completed`. Since there are no further updates for this task, the lint will never be run again. And since it doesn't yield, there won't be further updates - that is of course the whole point of the lint! (-;

Have a look at the tracing events I produced from within the `NeverYielded::check` method:

```text
2023-07-12T14:45:02.401182Z  INFO tokio_console::warnings: checking for NeverYielded: 'Some("greedy")' state: Running, total_polls: 1, busy: 9.778227771s
2023-07-12T14:45:02.626523Z  INFO tokio_console::warnings: checking for NeverYielded: 'Some("greedy")' state: Running, total_polls: 1, busy: 10.003567771s
2023-07-12T14:45:13.625785Z  INFO tokio_console::warnings: checking for NeverYielded: 'Some("greedy")' state: Running, total_polls: 1, busy: 998.316004ms
2023-07-12T14:45:13.626350Z  INFO tokio_console::warnings: checking for NeverYielded: 'Some("greedy")' state: Completed, total_polls: 1, busy: 20.003362491s
2023-07-12T14:45:33.625470Z  INFO tokio_console::warnings: checking for NeverYielded: 'Some("greedy")' state: Running, total_polls: 1, busy: 995.971509ms
2023-07-12T14:45:33.625605Z  INFO tokio_console::warnings: checking for NeverYielded: 'Some("greedy")' state: Completed, total_polls: 1, busy: 20.001640062s
2023-07-12T14:45:53.625837Z  INFO tokio_console::warnings: checking for NeverYielded: 'Some("greedy")' state: Running, total_polls: 1, busy: 991.334708ms
2023-07-12T14:45:53.626094Z  INFO tokio_console::warnings: checking for NeverYielded: 'Some("greedy")' state: Completed, total_polls: 1, busy: 20.004502567s
```

See how the first check happens when the task has already been running for 9 seconds. Then the subsequent checks (when the task is in state `Running`) occur when the task has been running for just under a second, so the warning check returns `false`.

If we got a bit luckier, that check might have happened 10 milliseconds later and it would have returned `true`. However, we probably need to find a better solution than being lucky.

It may be necessary for a lint to indicate that it should be checked again after some interval. This would allow the `NeverYielded` lint to ask to be checked again, for a specific task, at some time after the 1 second would have passed (or a different period if we make that default value configurable).

The `TasksState::update_tasks()` would then need a way to keep track of which lints need to be rechecked and then recheck them during the next cycle, **even if the task itself hasn't been updated**.

Sorry for the big dump of information. If you're still interested in working on this topic and need some help, let me know. You can comment here in the PR, or if you're on Discord, you can find me in the Tokio server as `haydenstainsby` (just post in the #console channel, I'll see it).

jefftt:
ah i see yeah that makes sense thank you for the detailed explaination. would still be interested in working on this - just need to find the time :) happy to pass this off if someone else gets to it before me but will take a stab otherwise

jefftt:
@hds finally got around to this and found an easy change to make to fix this issue which is to lint every task on each update. not sure if the performance hit that it takes is worth it though. thoughts?

hds:
@jefftt thanks for looking into this further! I just wanted to let you know that it will be a few days before I look at this PR as I'm on holiday at the moment. I should get to it in the middle of next week.

tokio-console/src/state/tasks.rs
```diff
@@ -217,14 +217,16 @@ impl TasksState {
...
                 Some((id, task))
             });
 
         for (stats, mut task) in self.tasks.updated(stats_update) {
             tracing::trace!(?task, ?stats, "processing stats update for");
             task.stats = stats.into();
-            task.lint(linters);
+        }
+
+        for (_, task) in self.tasks.iter() {
```
hds: I think that this strategy probably isn't ideal. I'm concerned that we will be adding too much additional processing on every update.

Especially as more lints are added, a situation in which there are lots and lots of tasks. In the case that most of these don't change in every update (which is probably a common situation), this will introduce a large amount of unnecessary overhead. Especially given that most of the lints won't change result unless there is a change in the task itself.

hds (changes requested):
Thanks for your work on this PR!

I think we need to find a more light weight solution. This approach will potentially add a lot of unnecessary computational overhead in every update cycle.

Would you like to discuss some different options?

jefftt:
Thanks for your help, as discussed offline, I updated the PR with the suggestions

tokio-console/src/state/tasks.rs
```diff
@@ -430,15 +453,21 @@ impl Task {
         &self.warnings[..]
     }
 
-    fn lint(&mut self, linters: &[Linter<Task>]) {
+    fn lint(&mut self, linters: &[Linter<Task>]) -> bool {
```
hds: I find that "raw" `bool` return values can be confusing, because it's not always clear what `true` and `false` mean from the function signature.

How about if we use a small enum here instead:
```rust
enum TaskLintResult {
    Linted,
    RequiresRecheck,
}
```

This also has the advantage that it isn't necessary to set the output of `lint()` to an intermediate variable like you've done above in order to document the meaning of the return value (I like the way you did that, if we weren't able to change the return type, then that would be a great improvement to readability).

So instead of:
```rust
 let recheck = task.lint(linters);
    if recheck {
        self.pending_lint.insert(task.id);
    }
```

you could write:
```rust
match task.lint(linters) {
    TaskLintResult::RequiresRecheck => next_pending_lint.insert(task.id),
    _ => {}
}
```

tokio-console/src/warnings.rs
```diff
@@ -46,6 +50,31 @@ pub trait Warn<T>: Debug {
...
+
+    /// The lint should be rechecked as the conditions to allow for checking are
+    /// not satisfied yet
+    Recheck,
+}
+
+/// A result for a warning check
+pub enum Warning {
+    /// Set `true` if the warning applies or `false` otherwise
+    Check(bool),
```
hds: I feel it might be simpler to have two separate enum variants, instead of keeping a bool in a single variant.

```suggestion
    /// No warning for this entity.
    Ok,
    
    /// A warning has been detected for this entity.
    Warn,
```

tokio-console/src/warnings.rs
```diff
@@ -46,6 +50,31 @@ pub trait Warn<T>: Debug {
     fn summary(&self) -> &str;
 }
 
+/// A result for a linter check
```
hds: Let's move this enum to below the `Linter<T>` struct where it's used.

tokio-console/src/state/tasks.rs
```diff
@@ -217,16 +218,38 @@ impl TasksState {
                     warnings: Vec::new(),
                     location,
                 };
-                task.lint(linters);
+                let recheck = task.lint(linters);
+                if recheck {
+                    self.pending_lint.insert(task.id);
```
hds: Instead add these to another `HashSet` that we won't check until next update (part of **PendingLintLogic**, see below).
```suggestion
                    next_pending_lint.insert(task.id);
```

tokio-console/src/state/tasks.rs
```diff
@@ -217,16 +218,38 @@ impl TasksState {
...
+        let mut checked = HashSet::new();
         for (stats, mut task) in self.tasks.updated(stats_update) {
             tracing::trace!(?task, ?stats, "processing stats update for");
             task.stats = stats.into();
-            task.lint(linters);
+            let recheck = task.lint(linters);
+            if !recheck {
+                self.pending_lint.remove(&task.id);
+            }
+            checked.insert(task.id);
```
hds: Switch the logic here (part of **PendingLintLogic**, see below).
```suggestion
            if recheck {
                next_pending_lint.insert(task.id);
            }
            self.pending_lint.remove(&task.id);
```

tokio-console/src/state/tasks.rs
```diff
@@ -217,16 +218,38 @@ impl TasksState {
...
         }
 
+        self.pending_lint.retain(|id| {
```
hds: **PendingLintLogic**

There are a couple of suggestions up above which relate to this topic, the full description is here.

I might be misunderstanding the logic here, but I think that there are some cases where tasks are gettinbg linted twice. Specifically, when a new task gets linted and returns to retain, then the task ID goes into `self.pending_lint`, but not into `checked` and so it will get checked again immediately.

I propose that instead of keeping the `checked` hash set, we instead create a new hash set for the tasks that we want to check **next** update cycle. Then the process would like like this:

1. Create new `HashSet` called `next_pending_lint`
2. New tasks loop
  a. lint task, if retain == true add ID to `next_pending_lint`.
3. Update tasks loop
  a. lint task, if retain == true add ID to `next_pending_lint`, if retain == false, remove ID from `self.pending_lint`
4. `self.pending_lint` loop
  a. get task, if task not available don't panic, just skip it
  b. lint task, if retain == true add ID to `next_pending_lint`
5. set `self.pending_lint` to `next_pending_lint`

This will avoid needing an extra check in (4), because we will already have removed any tasks which have already been linted (irrespective of their lint result).

tokio-console/src/state/tasks.rs
```diff
@@ -217,16 +218,38 @@ impl TasksState {
...
+            if checked.contains(id) {
+                true
+            } else {
+                let recheck = self
+                    .tasks
+                    .get(*id)
+                    .expect("task pending lint is not in task store")
```
hds: I mentioned this in the **PendingLintLogic** comment above, but I wanted to explain here.

A task might get removed from the list of tasks (because it's finished and been removed after whatever retention timeout we have). In that case, it would be normal to not be able to get the task by ID. So instead of `expect`ing that a task is returned, we should skip it.

hds (changes requested):
Great work! I think that the logic here is almost completely fine (one case where I believe we may lint some tasks twice). I've also got some style suggestions which I think will help us reason about the flow more easily in the future.

tokio-console/src/state/tasks.rs
jefftt: this is much simpler, thanks!

hds:
@jefftt Thanks for your changes! I was hoping to get to this review today, but it might be tomorrow. Just wanted to let you know that it will be soon!

tokio-console/src/warnings.rs
```diff
@@ -90,6 +89,34 @@ impl<T> Linter<T> {
...
+    /// No warning for this entity.
+    Ok,
+
+    /// A warning has been detected for this entity.
+    Warn,
+
+    /// The warning should be rechecked as the conditions to allow for checking
+    /// are not satisfied yet
+    Recheck,
+}
```
hds: nit: My bad, I think I didn't explain myself properly. We try to keep response definitions straight after where they're used (as much as possible).

So `Lint<T>` is now in the right place, but `Warning` should be where you had put it originally (right after the definitition of the trait `Warn<T>`).

tokio-console/src/state/tasks.rs
```diff
@@ -217,16 +221,35 @@ impl TasksState {
                     warnings: Vec::new(),
                     location,
                 };
-                task.lint(linters);
+                if matches!(task.lint(linters), TaskLintResult::RequiresRecheck) {
+                    next_pending_lint.insert(task.id);
+                }
```
hds: nit: It's more idiomatic Rust to use `if let` here.
```suggestion
                if let TaskLintResult::RequiresRecheck = task.lint(linters) {
                    next_pending_lint.insert(task.id);
                }
```

tokio-console/src/state/tasks.rs
```diff
@@ -217,16 +221,35 @@ impl TasksState {
...
         }
 
+        for id in &self.pending_lint {
+            if let Some(task) = self.tasks.get(*id) {
+                if matches!(
+                    task.borrow_mut().lint(linters),
+                    TaskLintResult::RequiresRecheck
+                ) {
+                    next_pending_lint.insert(*id);
+                }
```
hds: nit: It's more idiomatic Rust to use `if let` here.
```suggestion
                if let TaskLintResult::RequiresRecheck = task.borrow_mut().lint(linters) {
                    next_pending_lint.insert(*id);
                }
```

tokio-console/src/warnings.rs
```diff
@@ -78,7 +77,7 @@ impl<T> Linter<T> {
 
     pub(crate) fn format(&self, val: &T) -> String {
         debug_assert!(
-            self.0.check(val),
+            matches!(self.0.check(val), Warning::Ok),
```
hds: This should check for `Warning::Warn` instead of `Warning::Ok` (currently `tokio-console` panics every time I try to view a task with a warning on it).
```suggestion
            matches!(self.0.check(val), Warning::Warn),
```

tokio-console/src/state/tasks.rs
```diff
@@ -217,16 +221,35 @@ impl TasksState {
...
         for (stats, mut task) in self.tasks.updated(stats_update) {
             tracing::trace!(?task, ?stats, "processing stats update for");
             task.stats = stats.into();
-            task.lint(linters);
+            if matches!(task.lint(linters), TaskLintResult::RequiresRecheck) {
+                next_pending_lint.insert(task.id);
+            } else {
+                // Avoid linting this task again this cycle
+                self.pending_lint.remove(&task.id);
+            }
```
hds: nit: It's more idiomatic Rust to use a match statement here (even more so than using `if let` in the comment above):
```suggestion
            match task.lint(linters) {
                TaskLintResult::RequiresRecheck => next_pending_lint.insert(task.id),
                // Avoid linting this task again this cycle
                _ => self.pending_lint.remove(&task.id),
            }
```

hds (changes requested):
Thanks for your changes, we're almost there!

The debug assert in `Linter<T>::format` is currently causing `tokio-console` to panic when viewing the details for a task with a warning. Looks like it's just a logic flip.

I've also got some last style nits.

tokio-console/src/state/tasks.rs
jefftt: i was going back and forth on this, thanks for the clarification!

tokio-console/src/warnings.rs
jefftt: whoops, thanks

hds (approved):
This looks all set now. Thanks for all the work that you put into this PR, we really appreciate it!

Congratulations on your first `tokio-console` contribution!

PR #440, opened 2023-06-25
fix(subscriber): do not report excessive polling (#378)

### LINKED ISSUE #378: Console reporting excessive polling, that isn't there

tokio-console, console-subscriber

### What is the issue?

I've observed that `tokio-console` reports excessive polling, to the tune of around 1000 polls per second, for a simple spawned task that is not being woken. 

This is the task spawning

https://github.com/webrtc-rs/webrtc/blob/5d05b88de6d3a3eff3cd4e9b57219a25e01d3edc/webrtc/src/peer_connection/operation/mod.rs#L54-L56

and the task itself

https://github.com/webrtc-rs/webrtc/blob/b43e05b574be52b65b648bc367467bb6f14c5e11/webrtc/src/peer_connection/operation/mod.rs#L114-L129

I've observed the task being polled over 1M times over a 24 minute period without having been woken.

I've documented the details and made a small reproducible example in https://github.com/webrtc-rs/webrtc/pull/284

In that PR I've added a printing wrapper around the futures in question as well as introduced more logging. On my machine the excessive polling that tokio-console reports does not line up with the amount of times the futures are actually polled.

### Logs, error output, etc

_No response_

### Versions

```text
├── console-subscriber v0.1.7
│   ├── console-api v0.4.0
```

### Possible solution

_No response_

### Additional context

_No response_

maybe

MatteoNardi:
When entering and exiting a span the old code was also updating the parent
stats. This was causing excessive polling being reported for the parent tasks.
See issue #378 for more details. The regression was introduced by the refactor
in #238.
This fixes the issue by limiting updates to the current span.

Closes #378

### DIFF
--- a/console-subscriber/src/lib.rs
+++ b/console-subscriber/src/lib.rs
@@ -816,9 +816,6 @@ where
 
         if let Some(span) = cx.span(id) {
             if let Some(now) = update(&span, None) {
-                if let Some(parent) = span.parent() {
-                    update(&parent, Some(now));
-                }
                 self.current_spans
                     .get_or_default()
                     .borrow_mut()
@@ -860,9 +857,6 @@ where
 
         if let Some(span) = cx.span(id) {
             if let Some(now) = update(&span, None) {
-                if let Some(parent) = span.parent() {
-                    update(&parent, Some(now));
-                }
                 self.current_spans.get_or_default().borrow_mut().pop(id);
 
                 self.record(|| record::Event::Exit {
--- a/console-subscriber/src/lib.rs
+++ b/console-subscriber/src/lib.rs
@@ -789,33 +789,27 @@ where
     }
 
     fn on_enter(&self, id: &span::Id, cx: Context<'_, S>) {
-        fn update<S: Subscriber + for<'a> LookupSpan<'a>>(
-            span: &SpanRef<S>,
-            at: Option<Instant>,
-        ) -> Option<Instant> {
+        if let Some(span) = cx.span(id) {
+            let now = Instant::now();
             let exts = span.extensions();
             // if the span we are entering is a task or async op, record the
             // poll stats.
-            if let Some(stats) = exts.get::<Arc<stats::TaskStats>>() {
-                let at = at.unwrap_or_else(Instant::now);
-                stats.start_poll(at);
-                Some(at)
+            let is_relevant = if let Some(stats) = exts.get::<Arc<stats::TaskStats>>() {
+                stats.start_poll(now);
+                true
             } else if let Some(stats) = exts.get::<Arc<stats::AsyncOpStats>>() {
-                let at = at.unwrap_or_else(Instant::now);
-                stats.start_poll(at);
-                Some(at)
+                stats.start_poll(now);
+                true
             // otherwise, is the span a resource? in that case, we also want
             // to enter it, although we don't care about recording poll
             // stats.
             } else if exts.get::<Arc<stats::ResourceStats>>().is_some() {
-                Some(at.unwrap_or_else(Instant::now))
+                true
             } else {
-                None
-            }
-        }
+                false
+            };
 
-        if let Some(span) = cx.span(id) {
-            if let Some(now) = update(&span, None) {
+            if is_relevant {
                 self.current_spans
                     .get_or_default()
                     .borrow_mut()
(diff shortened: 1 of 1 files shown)


MatteoNardi:
I still have some doubts regarding the original need for the parent update. Here is the research I've made, hopefully it could be useful when reviewing this.

Before #238 landed, `on_enter` sent `id` and `parent_id` to the aggregator:
https://github.com/tokio-rs/console/blob/console-subscriber-v0.1.0/console-subscriber/src/lib.rs#L725-L732
The aggregator would then increase the poll stats of  `id` if it was a task and **increase the poll stats of  `parent_id` if it was an async op.**
https://github.com/tokio-rs/console/blob/console-subscriber-v0.1.0/console-subscriber/src/aggregator/mod.rs#L630-L634
Notice how this is different both from the change in #238 (which increases the stats of both spans) and the code after this PR (which updates the task and async op stats of the current span). Before commit 1aa9b594f30e42098c6c6bbf41eb1d2b01dc0426, I think the behavior matched the one in this PR.

If I got it right, I think this is what the various branches were updating:

|branch|current task|parent task|current async op|parent async op| 
|----------|-|-|-|-|
|this PR|yes|no|yes|no|
|main branch|yes|yes|yes|yes|
|before 238|yes|no|no|yes|
|before 1aa9b594f30e42098c6c6bbf41eb1d2b01dc0426|yes|no|yes|no|

I'll need some help understanding the proper behavior :D

console-subscriber/src/lib.rs
```diff
@@ -789,36 +789,27 @@ where
...
             // to enter it, although we don't care about recording poll
             // stats.
             } else if exts.get::<Arc<stats::ResourceStats>>().is_some() {
-                Some(at.unwrap_or_else(Instant::now))
+                true
             } else {
-                None
-            }
-        }
+                false
```
hds: I think that the logic here would be easier to follow if this else case returned early and the other 3 cases (which currently result in `is_relevant = true` fall through to push the span id and record the event.

console-subscriber/src/lib.rs
```diff
@@ -833,36 +824,27 @@ where
...
                 // to enter it, although we don't care about recording poll
                 // stats.
             } else if exts.get::<Arc<stats::ResourceStats>>().is_some() {
-                Some(at.unwrap_or_else(Instant::now))
+                true
             } else {
-                None
-            }
-        }
+                false
```
hds: Same comment here about returning early in this case and otherwise continuing the logic at the end of the function.

hds (changes requested):
I did some testing of this change myself and it seemed to work for all the cases I tried, so I believe your logic is correct.

I've got a couple of style comments on this change though.

console-subscriber/src/lib.rs
```diff
@@ -789,87 +789,59 @@ where
...
+                stats.start_poll(now);
             } else if let Some(stats) = exts.get::<Arc<stats::AsyncOpStats>>() {
-                let at = at.unwrap_or_else(Instant::now);
-                stats.start_poll(at);
-                Some(at)
+                stats.start_poll(now);
             // otherwise, is the span a resource? in that case, we also want
             // to enter it, although we don't care about recording poll
             // stats.
             } else if exts.get::<Arc<stats::ResourceStats>>().is_some() {
```
hds: I think it might make more sense to put this comment inside the (now) empty block following the `else if`

```suggestion
            } else if exts.get::<Arc<stats::ResourceStats>>().is_some() {
                // otherwise, is the span a resource? in that case, we also want
                // to enter it, although we don't care about recording poll
                // stats.
```

console-subscriber/src/lib.rs
```diff
@@ -789,87 +789,59 @@ where
...
-                stats.end_poll(at);
-                Some(at)
-                // otherwise, is the span a resource? in that case, we also want
-                // to enter it, although we don't care about recording poll
-                // stats.
+                stats.end_poll(now);
+            // otherwise, is the span a resource? in that case, we also want
+            // to enter it, although we don't care about recording poll
+            // stats.
             } else if exts.get::<Arc<stats::ResourceStats>>().is_some() {
```
hds: I think it might make more sense to put this comment inside the (now) empty block following the `else if`

```suggestion
            } else if exts.get::<Arc<stats::ResourceStats>>().is_some() {
                // otherwise, is the span a resource? in that case, we also want
                // to enter it, although we don't care about recording poll
                // stats.
```

hds (changes requested):
Just a couple of small changes for clarity.

hds (approved):
Looks great. Thank you very much!

hds:
@hawkw Would you mind taking a look at this one now that CI is green again? I'm pretty sure that this is the correct logic. What I'm not sure about is why the previous logic was there, which makes me think that I'm missing something important.

hawkw:
yeah, i've been meaning to give this one a closer look!

hawkw:
@hds 
> What I'm not sure about is why the previous logic was there, which makes me think that I'm missing something important.

After thinking about this for a bit, I think the previous logic is just...wrong. I believe I wrote it that way because I was blindly copying the behavior a `tracing` subscriber would do for tracking *span* timing, but spawned task spans are not the same as spans that execute inside of other spans. A spawned task shouldn't be incrementing its parent's active times, and I just wasn't thinking when I wrote that code. :)

I'm pretty sure this change is correct. It does remind me that we should really probably just change `tokio` to have `parent: None` for task spans...

hds:
> I'm pretty sure this change is correct. It does remind me that we should really probably just change `tokio` to have `parent: None` for task spans...

Yeah. We should. I have that issue for this, tokio-rs/tokio#5792.

hawkw (approved):

PR #441, opened 2023-07-02
fix(console): make long locations readable

### LINKED ISSUE #411: make long Locations readable

I can't read where my `task`'s `Location` is, because the paths (which I don't control in all environments and cases) are super long. They're private, but you can see the pixelated piece and that's the path going beyond where the bounding box lets me see.

image

most of the time I probably would be fine with the last 3-4 segments of the path rather than the whole path, but it's great to have the whole thing. it may be a bit janky, but a toggle to wordwrap might do the trick?

could allow some type of navigation in this pane, make it full length, make it full length-able

_No response_

### Would you like to work on this feature?

maybe

guerinoni:
This closes #411.

This is how it looks like.

Screenshot 2023-07-03 at 12 22 24 PM

### DIFF
--- a/tokio-console/src/state/mod.rs
+++ b/tokio-console/src/state/mod.rs
@@ -522,7 +522,7 @@ fn format_location(loc: Option<proto::Location>) -> String {
             let truncated = truncate_registry_path(file);
             l.file = Some(truncated);
         }
-        format!("{} ", l)
+        format!("{}", l)
     })
     .unwrap_or_else(|| "<unknown location>".to_string())
 }
--- a/tokio-console/src/view/task.rs
+++ b/tokio-console/src/view/task.rs
@@ -155,10 +155,17 @@ impl TaskView {
             Span::raw(task.target()),
         ]));
 
-        overview.push(Spans::from(vec![
-            bold("Location: "),
-            Span::raw(task.location()),
-        ]));
+        let title = "Location: ";
+        let location_max_width = stats_area[0].width as usize - 2 - title.len(); // NOTE: -2 for the border
+        let location = if task.location().len() > location_max_width {
+            let ellipsis = styles.if_utf8("\u{2026}", "...");
+            let start = task.location().len() - location_max_width + ellipsis.chars().count();
+            format!("{}{}", ellipsis, &task.location()[start..])
+        } else {
+            task.location().to_string()
+        };
+
+        overview.push(Spans::from(vec![bold(title), Span::raw(location)]));
 
         let total = task.total(now);
 


tokio-console/src/view/task.rs
```diff
@@ -155,10 +155,16 @@ impl TaskView {
             Span::raw(task.target()),
         ]));
 
-        overview.push(Spans::from(vec![
-            bold("Location: "),
-            Span::raw(task.location()),
-        ]));
+        let title = "Location: ";
+        let mut m = Span::raw(task.location());
+        let row_len = m.width() + title.len() + 4;
```
hds: It would be great if you could add a comment explaining what the extra 4 characters is for. I'm actually not sure myself.

tokio-console/src/view/task.rs
```diff
@@ -155,10 +155,16 @@ impl TaskView {
             Span::raw(task.target()),
         ]));
 
-        overview.push(Spans::from(vec![
-            bold("Location: "),
-            Span::raw(task.location()),
-        ]));
+        let title = "Location: ";
+        let mut m = Span::raw(task.location());
```
hds: For style reasons, I'd skip the `mut` here and assign the location from an `if .. else` expression. Something like:

```rust
let location = if row_len <= stats_area[0].width.into() {
    Span::raw(task.location())
} else {
   // existing logic to produce: Span::raw(format!("...{}", location));
}

overview.push(Spans::from(vec![bold(title), location]));
```

We have a preference for more descriptive variable names too. So ideally `m` would be `location` or perhaps `value` (as it goes with `title`).

hds (changes requested):
Thank you very much for your PR! I've got a few changes if you don't mind.

tokio-console/src/view/task.rs
guerinoni: Sure, added! Also added a screenshot to show how "is working"

tokio-console/src/view/task.rs
guerinoni: I improved the syntax!

tokio-console/src/state/mod.rs
```diff
@@ -522,7 +522,7 @@ fn format_location(loc: Option<proto::Location>) -> String {
             let truncated = truncate_registry_path(file);
             l.file = Some(truncated);
         }
-        format!("{} ", l)
+        format!("{}", l)
```
hds: We'll have to wait for #443 to go in, then this part of the diff will go away.

hds (approved):
Looking good. Thank you!

We'll have to wait for #443 to get merged, but then it should be good to go. I'm approving now to run the CI pipeline (nightly will fail, but that is fixed in #442).

hawkw (approved):
looks good to me!

PR #444, opened 2023-07-03
chore: remove branch config from release action

hawkw:
This appears to be breaking it.


hds (approved):
Looks good. Let's see if this fixes it!

PR #449, opened 2023-07-11
feat(console): Add way to inspect details of task from resource view

### LINKED ISSUE #448: Add key to jump from an async op to the related task

In the resource detail view, there is a table of async ops associated with the resource. The user can move up and down to highlight a specific resource. However, going straight to the associated task isn't easy.

tokio-console-screenshot-2023-July-7-60416

It would be nice to be able to press a key and jump to that task in the tasks list view, or a different key and jump straight to the task's detail view.

We could also open a modal view with the task details in there.

Perhaps the `return` key could be used to jump to the task in the task list. This would mean that hitting return twice would result in jumping on to the details view. Alternatively (or in addition), a specific key (`d` for details?) could take the user straight to the task details view.

guerinoni:
Now it is possible while navigate in `async_ops_table` to go directly to
details of the task that you are pointing on the list.

This closes #448

### DIFF
--- a/tokio-console/src/view/mod.rs
+++ b/tokio-console/src/view/mod.rs
@@ -167,6 +167,31 @@ impl View {
                         self.state = ResourcesList;
                         update_kind = UpdateKind::Other;
                     }
+                    key!(Enter) => {
+                        if let Some(op) = view.async_ops_table.selected_item().upgrade() {
+                            if let Some(task_id) = op.borrow().task_id() {
+                                let task: u64 = task_id.to_string().parse().unwrap();
+                                let item = self
+                                    .tasks_list
+                                    .sorted_items
+                                    .iter()
+                                    .filter_map(|i| i.upgrade())
+                                    .find(|t| {
+                                        let task_id: u64 =
+                                            t.borrow().id().to_string().parse().unwrap();
+                                        task_id == task
+                                    });
+
+                                if let Some(t) = item {
+                                    update_kind = UpdateKind::SelectTask(t.borrow().span_id());
+                                    self.state = TaskInstance(self::task::TaskView::new(
+                                        t,
+                                        state.task_details_ref(),
+                                    ));
+                                }
+                            }
+                        }
+                    }
                     _ => {
                         // otherwise pass on to view
                         view.update_input(event);
--- a/tokio-console/src/view/mod.rs
+++ b/tokio-console/src/view/mod.rs
@@ -170,22 +170,17 @@ impl View {
                     key!(Enter) => {
                         if let Some(op) = view.async_ops_table.selected_item().upgrade() {
                             if let Some(task_id) = op.borrow().task_id() {
-                                let task: u64 = task_id.to_string().parse().unwrap();
-                                let item = self
+                                let task = self
                                     .tasks_list
                                     .sorted_items
                                     .iter()
                                     .filter_map(|i| i.upgrade())
-                                    .find(|t| {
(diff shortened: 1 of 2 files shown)


tokio-console/src/view/mod.rs
```diff
@@ -167,6 +167,31 @@ impl View {
...
+                                let item = self
+                                    .tasks_list
+                                    .sorted_items
+                                    .iter()
+                                    .filter_map(|i| i.upgrade())
+                                    .find(|t| {
+                                        let task_id: u64 =
+                                            t.borrow().id().to_string().parse().unwrap();
+                                        task_id == task
+                                    });
```
hds: The `Id<Task>` can be compared directly, so it doesn't need to be turned into a string and then parsed to a u64.

Also, let's call the result `task`, as I think it's clearer than `item`.

```suggestion
                                let task = self
                                    .tasks_list
                                    .sorted_items
                                    .iter()
                                    .filter_map(|i| i.upgrade())
                                    .find(|t| task_id == t.borrow().id());
```

tokio-console/src/view/mod.rs
```diff
@@ -167,6 +167,31 @@ impl View {
...
+
+                                if let Some(t) = item {
+                                    update_kind = UpdateKind::SelectTask(t.borrow().span_id());
+                                    self.state = TaskInstance(self::task::TaskView::new(
+                                        t,
+                                        state.task_details_ref(),
+                                    ));
+                                }
```
hds: We've got a preference for longer variable names to make things clearer without an IDE. Also, we can shadow the variable name so we don't need to pick something different.
```suggestion
                                if let Some(task) = task {
                                    update_kind = UpdateKind::SelectTask(task.borrow().span_id());
                                    self.state = TaskInstance(self::task::TaskView::new(
                                        task,
                                        state.task_details_ref(),
                                    ));
                                }
```

tokio-console/src/view/resource.rs
```diff
@@ -21,7 +21,7 @@ use std::{cell::RefCell, rc::Rc};
 
 pub(crate) struct ResourceView {
     resource: Rc<RefCell<Resource>>,
-    async_ops_table: TableListState<AsyncOpsTable, 9>,
+    pub async_ops_table: TableListState<AsyncOpsTable, 9>,
```
hds: Let's make this `pub(crate)` to match the struct it's in.

```suggestion
    pub(crate) async_ops_table: TableListState<AsyncOpsTable, 9>,
```

hds (changes requested):
Thank you for fixing that error!

I've got a few code suggestions to avoid `unwrap()` and make some parts clearer on first read through.

hds (approved):
Looks good to me. Thank you!

hds:
@hawkw when you've got a moment, would you like to have a look at this PR.

hawkw (approved):
this looks good to me! i wonder if we ought to add "enter: select task" to the help text as well?

guerinoni:
> this looks good to me! i wonder if we ought to add "enter: select task" to the help text as well?

Hey, it is already there
Screenshot 2023-07-15 at 2 43 42 PM

It is `view details`, in this context means task's details.

hds:
> this looks good to me! i wonder if we ought to add "enter: select task" to the help text as well?

@hawkw We could probably improve the text on the controls list, but I would do that in a separate change since the enter key was already listed (erroneously) on this view and so it's somewhat unrelated.

Issue #450, opened 2023-07-14
`console-subscriber` crate has low test coverage


hds:
### What crate(s) in this repo are involved in the problem?

console-subscriber

### What is the issue?

The `console-subscriber` crate has no integration tests. There are some unit tests, but the coverage isn’t great.

Recently, we’ve found or fixed a few of issues which probably could have been caught with a bit of testing. For example the issue #378 (fixed in #440), and the fixes #430 and #447.

### Possible solution

Ideally, we would like a way to test the `console-subscriber` end to end, starting with an instrumented tokio runtime and ending with a validation of tasks received by a client to the gRPC server.

Much of this setup is likely to be duplicated between tests, so some of that test framework should be factored out into common code. Although we likely don’t need anything as complete as [`tracing-mock`](https://github.com/tokio-rs/tracing/tree/master/tracing-mock) in the tracing project.

### Would you like to work on fixing this bug?

yes, I already am

PR #451, opened 2023-07-18
feat(subscriber) expose server parts

### LINKED ISSUE #428: Enable adding the gRPC API to tonic without serving

I would like the `Server` to optionally `serve` the `InstrumentServer` after it's been added as a service to the tonic `Builder`. As of now, the `Server` implementation starts the aggregator, add the service to tonic and start the gRPC service. Ideally, this last step could be optional, leaving the user the choice of adding more services or using this tonic service in a larger set of services: in my case, I would like to transform the tonic service to a tower service and multiplex it with an Axum service.

Probably by extending the `Service` API so that the aggregation is started and the `InstrumentService` is properly added to tonic, returning a tonic `Router`. I have a fork that I'm using for production here: https://github.com/tokio-rs/console/commit/700b80a0013f67627045bc310972b2811c653ad9

There can be a few drawbacks with this:
* The builder can be configured with a listening port / Unix socket so it might feel surprising to have an API that completely ignores it
* We lose the fine grained control of the aggregator thread handle (either by dropping it in this new API body or by leaking it to the user)

I did not consider another solution yet.

_No response_

### Would you like to work on this feature?

maybe

hds:
The `ConsoleLayer` builder provides the user with a console layer and
a server, which is used to start the gRPC server.

However, it may be desireable to expose the instrumentation server together
with other services on the same Tonic router. This was requested
explicitly in #449 428.

Additionally, to add tests which make use of the instrumentation server
(as part of improving test coverage for #450), more flexibility is
needed than what is provided by the current API. Specifically we would
like to connect a client and server via an in memory channel, rather
than a TCP connection.

This change adds an additional method to `console_subscriber::Server`
called `into_parts` which allows the user to access the
`InstrumentServer` directly. The `Aggregator` is also returned and
must be set to run for at least as long as the instrument server. This
allows the aggregator to be spawned wherever the user wishes.

To facilitate the addition of functionality which would result in more
"parts" in the future, `into_parts` returns a non-exhaustive struct,
rather than a tuple of parts.

Closes: #428

### DIFF
--- a/console-subscriber/src/lib.rs
+++ b/console-subscriber/src/lib.rs
@@ -42,7 +42,7 @@ mod stats;
 pub(crate) mod sync;
 mod visitors;
 
-use aggregator::Aggregator;
+pub use aggregator::Aggregator;
 pub use builder::{Builder, ServerAddr};
 use callsites::Callsites;
 use record::Recorder;
@@ -941,10 +941,11 @@ impl Server {
     ) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
         let addr = self.addr.clone();
         let ServerParts {
-            instrument_server: service,
-            aggregator_handle: aggregate,
+            instrument_server,
+            aggregator,
         } = self.into_parts();
-        let router = builder.add_service(service);
+        let aggregate = spawn_named(aggregator.run(), "console::aggregate");
+        let router = builder.add_service(instrument_server);
         let res = match addr {
             ServerAddr::Tcp(addr) => {
                 let serve = router.serve(addr);
@@ -957,16 +958,20 @@ impl Server {
                 spawn_named(serve, "console::serve").await
             }
         };
-        drop(aggregate);
+        aggregate.abort();
         res?.map_err(Into::into)
     }
 
-    /// Returns the parts needed to spawn a gRPC server and keep the aggregation
-    /// worker running.
+    /// Returns the parts needed to spawn a gRPC server and the aggregator that
+    /// supplies it.
     ///
     /// Note that a server spawned in this way will disregard any value set by
     /// [`Builder::server_addr`], as the user becomes responsible for defining
     /// the address when calling [`Router::serve`].
+    ///
+    /// Additionally, the user of this API must ensure that the [`Aggregator`]
+    /// is running for as long as the gRPC server is. If the server stops
+    /// running, the aggregator task can be aborted.
     ///
     /// # Examples
     ///
@@ -984,10 +989,11 @@ impl Server {
     /// let (console_layer, server) = ConsoleLayer::builder().build();
     /// let ServerParts {
     ///     instrument_server,
-    ///     aggregator_handle,
+    ///     aggregator,
     ///     ..
     /// } = server.into_parts();
     ///
+    /// let aggregator_handle = tokio::spawn(aggregator.run());
     /// let router = tonic::transport::Server::builder()
     ///     //.add_service(some_other_service)
     ///     .add_service(instrument_server);
@@ -1007,19 +1013,16 @@ impl Server {
     ///
     /// [`Router::serve`]: fn@tonic::transport::server::Router::serve
     pub fn into_parts(mut self) -> ServerParts {
-        let aggregate = self
+        let aggregator = self
             .aggregator
             .take()
             .expect("cannot start server multiple times");
-        let aggregate = spawn_named(aggregate.run(), "console::aggregate");
 
-        let service = proto::instrument::instrument_server::InstrumentServer::new(self);
+        let instrument_server = proto::instrument::instrument_server::InstrumentServer::new(self);
 
         ServerParts {
-            instrument_server: service,
-            aggregator_handle: AggregatorHandle {
-                join_handle: aggregate,
-            },
+            instrument_server,
+            aggregator,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


console-subscriber/src/lib.rs
```diff
@@ -957,9 +957,110 @@ impl Server {
...
+#[non_exhaustive]
+pub struct ServerParts {
+    /// The instrument server.
+    ///
+    /// See the documentation for [`InstrumentServer`] for details.
+    pub instrument_server: InstrumentServer<Server>,
+
+    /// The aggregate handle.
+    ///
+    /// See the documentation for [`AggregatorHandle`] for details.
```
hawkw: Since dropping this shuts down the aggregator task, we should probably add a Big Scary Warning on this field making it clear that it must not be dropped.

console-subscriber/src/lib.rs
```diff
@@ -957,9 +957,110 @@ impl Server {
...
+///
+/// See the [`Server::into_parts`] documentation for usage.
+#[non_exhaustive]
+pub struct ServerParts {
+    /// The instrument server.
+    ///
+    /// See the documentation for [`InstrumentServer`] for details.
+    pub instrument_server: InstrumentServer<Server>,
+
+    /// The aggregate handle.
```
hawkw: nit: how about
```suggestion
    /// A handle to the background worker task responsible for aggregating trace data.
```

console-subscriber/src/lib.rs
```diff
@@ -957,9 +957,110 @@ impl Server {
...
+    ///
+    /// See the documentation for [`AggregatorHandle`] for details.
+    pub aggregator_handle: AggregatorHandle,
+}
+
+/// Aggregator handle.
```
hawkw, line 1063: This documentation doesn't really explain to a user what the heck an "aggregator handle" _is_ or why they should care. Let's make sure this actually states what it does, and what it's for.

console-subscriber/src/lib.rs
```diff
@@ -957,9 +957,110 @@ impl Server {
...
-        aggregate.abort();
+        drop(aggregate);
         res?.map_err(Into::into)
     }
+
+    /// Returns the parts needed to spawn a gRPC server and keep the aggregation
+    /// worker running.
+    ///
+    /// Note that a server spawned in this way will overwrite any value set by
+    /// [`Builder::server_addr`] as the user becomes responsible for defining
```
hawkw:
```suggestion
    /// [`Builder::server_addr`], as the user becomes responsible for defining
```

console-subscriber/src/lib.rs
```diff
@@ -957,9 +957,110 @@ impl Server {
...
         };
-        aggregate.abort();
+        drop(aggregate);
         res?.map_err(Into::into)
     }
+
+    /// Returns the parts needed to spawn a gRPC server and keep the aggregation
+    /// worker running.
+    ///
+    /// Note that a server spawned in this way will overwrite any value set by
```
hawkw:
```suggestion
    /// Note that a server spawned in this way will disregard any value set by
```

console-subscriber/src/lib.rs
```diff
@@ -957,9 +957,110 @@ impl Server {
...
+/// kept as long as the `InstrumentServer<Server>` - which is also
+/// returned - is in use.
+pub struct AggregatorHandle {
+    join_handle: JoinHandle<()>,
+}
+
+impl Drop for AggregatorHandle {
+    fn drop(&mut self) {
+        self.join_handle.abort();
+    }
```
hawkw: I'm not convinced we should always do this. Instead, perhaps, the `ServerParts` should just give the user access to the `JoinHandle` so they can decide whether they want to abort the task? The behavior  of `serve_with` was specifically intended to ensure that the aggregator is shut down if the server task ends.

hds: @hawkw OK, that makes sense. I was hoping to avoid abstract away the fact that this is a join handle underneath. Do you think there's any value in that?

I see 2 options here:
* Keep the `AggregatorHandle` without the `impl Drop` and add an `abort()` method to explicitly shut down the aggregator.
* Expose the join handle as is.

Do you have a specific preference?

console-subscriber/src/lib.rs
hds: I've removed the drop behaviour as you suggested in a later comment, so I think this doesn't apply any more.

console-subscriber/src/lib.rs
hds: I've tried to add more context here.

console-subscriber/src/lib.rs
hds: @hawkw i went with the first option. Let me know what you think.

hawkw: > OK, that makes sense. I was hoping to avoid abstract away the fact that this is a join handle underneath

Hmm...now that I think about it a bit more, wouldn't the ideal lower-level interface just be to expose the future itself, instead of spawning it and handing the user a `JoinHandle`? Then, the user could spawn the aggregator task wherever they want to, which could be useful if the user is using a non-Tokio runtime, or has multiple Tokio runtime, and wants to control where the aggregator task is spawned. Also, we can avoid having to think about the `JoinHandle` entirely if it's up to the user to spawn the aggregator. 

What do you think?

hds: I think that makes a lot of sense.

Unfortunately I wasn't able to find a way to return and then later use a future (dyn, boxed, both, or otherwise), so I've made the Aggregator public with just the `run()` method which can be used to spawn it.

How does that look?

hawkw (approved):
Thanks for going along with my last-minute suggested changes, this looks great!

PR #452, opened 2023-07-18
test(subscriber): add initial integration tests

hds:
The `console-subscriber` crate has no integration tests. There are some
unit tests, but without very high coverage of features.

Recently, we've found or fixed a few errors which probably could have
been caught by a medium level of integration testing.

However, testing `console-subscriber` isn't straight forward. It is
effectively a tracing subscriber (or layer) on one end, and a gRPC
server on the other end.

This change adds enough of a testing framework to write some initial
integration tests. It is the first step towards closing #450.

Each test comprises 2 parts:
- One or more "expcted tasks"
- A future which will be driven to completion on a dedicated Tokio runtime.

Behind the scenes, a console subscriber layer is created and it's server
part is connected to a duplex stream. The client of the duplex stream
then records incoming updates and reconstructs "actual tasks". The layer
itself is set as the default subscriber for the duration of `block_on`
which is used to drive the provided future to completioin.

The expected tasks have a set of "matches", which is how we find the
actual task that we want to validate against. Currently, the only value
we match on is the task's name.

The expected tasks also have a set of expectations. These are other
fields on the actual task which are validated once a matching task is
found. Currently, the two fields which can have expectations set on them
are the `wakes` and `self_wakes` fields.

So, to construct an expected task, which will match a task with the name
`"my-task"` and then validate that the matched task gets woken once, the
code would be:

```rust
ExpectedTask::default()
    .match_name("my-task")
    .expect_wakes(1);
```

A future which passes this test could be:

```rust
async {
    task::Builder::new()
        .name("my-task")
        .spawn(async {
            tokio::time::sleep(std::time::Duration::ZERO).await
        })
}
```

The full test would then look like:

```rust
fn wakes_once() {
    let expected_task = ExpectedTask::default()
        .match_name("my-task")
        .expect_wakes(1);

    let future = async {
        task::Builder::new()
            .name("my-task")
            .spawn(async {
                tokio::time::sleep(std::time::Duration::ZERO).await
            })
    };

    assert_task(expected_task, future);
}
```

The PR depends on 2 others:
 - #447 which fixes an error in the logic that determines whether a task
   is retained in the aggregator or not.
 - #451 which exposes the server parts and is necessary to allow us to
   connect the instrument server and client via a duplex channel.

This change contains some initial tests for wakes and self wakes which
would have caught the error fixed in #430. Additionally there are tests
for the functionality of the testing framework itself.

### DIFF
--- a/console-subscriber/Cargo.toml
+++ b/console-subscriber/Cargo.toml
@@ -40,7 +40,7 @@ tonic = { version = "0.9", features = ["transport"] }
 tracing-core = "0.1.24"
 tracing = "0.1.26"
 tracing-subscriber = { version = "0.3.11", default-features = false, features = ["fmt", "registry"] }
-futures = { version = "0.3", default-features = false }
+futures-task = { version = "0.3", default-features = false }
 hdrhistogram = { version = "7.3.0", default-features = false, features = ["serialization"] }
 # The parking_lot dependency is renamed, because we want our `parking_lot`
 # feature to also enable `tracing-subscriber`'s parking_lot feature flag.
--- a/console-subscriber/Cargo.toml
+++ b/console-subscriber/Cargo.toml
@@ -55,6 +55,7 @@ crossbeam-channel = "0.5"
 
 [dev-dependencies]
 tokio = { version = "^1.21", features = ["full", "rt-multi-thread"] }
+tower = "0.4"
 futures = "0.3"
 
 [package.metadata.docs.rs]
--- a/console-subscriber/Cargo.toml
+++ b/console-subscriber/Cargo.toml
@@ -55,7 +55,7 @@ crossbeam-channel = "0.5"
 
 [dev-dependencies]
 tokio = { version = "^1.21", features = ["full", "rt-multi-thread"] }
-tower = "0.4"
+tower = { version = "0.4", default-features = false }
 futures = "0.3"
 
 [package.metadata.docs.rs]
--- a/console-subscriber/tests/framework.rs
+++ b/console-subscriber/tests/framework.rs
@@ -27,7 +27,7 @@ fn expect_present() {
 
 #[test]
 #[should_panic(expected = "Test failed: Task validation failed:
- - Task<name=main>: no expectations set, if you want to just expect that a matching task is present, use `expect_present()`
+ - Task { name=console-test::main }: no expectations set, if you want to just expect that a matching task is present, use `expect_present()`
 ")]
 fn fail_no_expectations() {
     let expected_task = ExpectedTask::default().match_default_name();
@@ -52,7 +52,7 @@ fn wakes() {
 
 #[test]
 #[should_panic(expected = "Test failed: Task validation failed:
- - Task<name=main>: expected `wakes` to be 5, but actual was 1
+ - Task { name=console-test::main }: expected `wakes` to be 5, but actual was 1
 ")]
 fn fail_wakes() {
     let expected_task = ExpectedTask::default().match_default_name().expect_wakes(5);
@@ -77,7 +77,7 @@ fn self_wakes() {
 
 #[test]
 #[should_panic(expected = "Test failed: Task validation failed:
- - Task<name=main>: expected `self_wakes` to be 1, but actual was 0
+ - Task { name=console-test::main }: expected `self_wakes` to be 1, but actual was 0
 ")]
 fn fail_self_wake() {
     let expected_task = ExpectedTask::default()
@@ -108,7 +108,7 @@ fn test_spawned_task() {
 
 #[test]
 #[should_panic(expected = "Test failed: Task validation failed:
- - Task<name=wrong-name>: no matching actual task was found
+ - Task { name=wrong-name }: no matching actual task was found
 ")]
 fn fail_wrong_task_name() {
     let expected_task = ExpectedTask::default().match_name("wrong-name".into());
@@ -151,7 +151,7 @@ fn multiple_tasks() {
 
 #[test]
 #[should_panic(expected = "Test failed: Task validation failed:
- - Task<name=task-2>: expected `wakes` to be 2, but actual was 1
+ - Task { name=task-2 }: expected `wakes` to be 2, but actual was 1
 ")]
 fn fail_1_of_2_expected_tasks() {
     let expected_tasks = vec![
--- a/console-subscriber/tests/framework.rs
+++ b/console-subscriber/tests/framework.rs
@@ -1,0 +1,184 @@
+//! Framework tests
+//!
+//! The tests in this module are here to verify the testing framework itself.
+//! As such, some of these tests may be repeated elsewhere (where we wish to
+//! actually test the functionality of `console-subscriber`) and others are
+//! negative tests that should panic.
+
+use std::time::Duration;
+
+use tokio::{task, time::sleep};
+
+mod support;
+use support::{assert_task, assert_tasks, ExpectedTask};
+
+#[test]
+fn expect_present() {
+    let expected_task = ExpectedTask::default()
+        .match_default_name()
+        .expect_present();
+
+    let future = async {
+        sleep(Duration::ZERO).await;
+    };
+
+    assert_task(expected_task, future);
+}
+
+#[test]
+#[should_panic(expected = "Test failed: Task validation failed:
+ - Task<name=main>: no expectations set, if you want to just expect that a matching task is present, use `expect_present()`
+")]
+fn fail_no_expectations() {
+    let expected_task = ExpectedTask::default().match_default_name();
+
+    let future = async {
+        sleep(Duration::ZERO).await;
+    };
+
+    assert_task(expected_task, future);
+}
+
+#[test]
+fn wakes() {
+    let expected_task = ExpectedTask::default().match_default_name().expect_wakes(1);
+
+    let future = async {
+        sleep(Duration::ZERO).await;
+    };
+
+    assert_task(expected_task, future);
+}
+
+#[test]
+#[should_panic(expected = "Test failed: Task validation failed:
+ - Task<name=main>: expected `wakes` to be 5, but actual was 1
+")]
+fn fail_wakes() {
+    let expected_task = ExpectedTask::default().match_default_name().expect_wakes(5);
+
+    let future = async {
+        sleep(Duration::ZERO).await;
+    };
+
+    assert_task(expected_task, future);
+}
+
+#[test]
+fn self_wakes() {
+    let expected_task = ExpectedTask::default()
+        .match_default_name()
+        .expect_self_wakes(1);
+
+    let future = async { task::yield_now().await };
+
+    assert_task(expected_task, future);
+}
+
+#[test]
+#[should_panic(expected = "Test failed: Task validation failed:
+ - Task<name=main>: expected `self_wakes` to be 1, but actual was 0
+")]
+fn fail_self_wake() {
+    let expected_task = ExpectedTask::default()
+        .match_default_name()
+        .expect_self_wakes(1);
+
+    let future = async {
+        sleep(Duration::ZERO).await;
+    };
+
+    assert_task(expected_task, future);
+}
+
+#[test]
+fn test_spawned_task() {
+    let expected_task = ExpectedTask::default()
+        .match_name("another-name".into())
+        .expect_present();
+
+    let future = async {
+        task::Builder::new()
+            .name("another-name")
+            .spawn(async { task::yield_now().await })
+    };
+
+    assert_task(expected_task, future);
+}
+
+#[test]
+#[should_panic(expected = "Test failed: Task validation failed:
+ - Task<name=wrong-name>: no matching actual task was found
+")]
+fn fail_wrong_task_name() {
+    let expected_task = ExpectedTask::default().match_name("wrong-name".into());
+
+    let future = async { task::yield_now().await };
+
+    assert_task(expected_task, future);
+}
+
+#[test]
+fn multiple_tasks() {
+    let expected_tasks = vec![
+        ExpectedTask::default()
+            .match_name("task-1".into())
+            .expect_wakes(1),
+        ExpectedTask::default()
+            .match_name("task-2".into())
+            .expect_wakes(1),
+    ];
+
+    let future = async {
+        let task1 = task::Builder::new()
+            .name("task-1")
+            .spawn(async { task::yield_now().await })
+            .unwrap();
+        let task2 = task::Builder::new()
+            .name("task-2")
+            .spawn(async { task::yield_now().await })
+            .unwrap();
+
+        tokio::try_join! {
+            task1,
+            task2,
+        }
+        .unwrap();
+    };
+
+    assert_tasks(expected_tasks, future);
+}
+
+#[test]
+#[should_panic(expected = "Test failed: Task validation failed:
+ - Task<name=task-2>: expected `wakes` to be 2, but actual was 1
+")]
+fn fail_1_of_2_expected_tasks() {
+    let expected_tasks = vec![
+        ExpectedTask::default()
+            .match_name("task-1".into())
+            .expect_wakes(1),
+        ExpectedTask::default()
+            .match_name("task-2".into())
+            .expect_wakes(2),
+    ];
+
+    let future = async {
+        let task1 = task::Builder::new()
+            .name("task-1")
+            .spawn(async { task::yield_now().await })
+            .unwrap();
+        let task2 = task::Builder::new()
+            .name("task-2")
+            .spawn(async { task::yield_now().await })
+            .unwrap();
+
+        tokio::try_join! {
+            task1,
+            task2,
+        }
+        .unwrap();
+    };
+
+    assert_tasks(expected_tasks, future);
+}
--- a/console-subscriber/tests/support/mod.rs
+++ b/console-subscriber/tests/support/mod.rs
@@ -1,0 +1,47 @@
+use futures::Future;
+
+mod state;
+mod subscriber;
+mod task;
+
+use subscriber::run_test;
+
+pub(crate) use subscriber::MAIN_TASK_NAME;
+pub(crate) use task::ExpectedTask;
+
+/// Assert that an `expected_task` is recorded by a console-subscriber
+/// when driving the provided `future` to completion.
+///
+/// This function is equivalent to calling [`assert_tasks`] with a vector
+/// containing a single task.
+///
+/// # Panics
+///
+/// This function will panic if the expectations on the expected task are not
+/// met or if a matching task is not recorded.
+#[track_caller]
+#[allow(dead_code)]
+pub(crate) fn assert_task<Fut>(expected_task: ExpectedTask, future: Fut)
+where
+    Fut: Future + Send + 'static,
+    Fut::Output: Send + 'static,
+{
+    run_test(vec![expected_task], future)
+}
+
+/// Assert that the `expected_tasks` are recorded by a console-subscriber
+/// when driving the provided `future` to completion.
+///
+/// # Panics
+///
+/// This function will panic if the expectations on any of the expected tasks
+/// are not met or if matching tasks are not recorded for all expected tasks.
+#[track_caller]
+#[allow(dead_code)]
+pub(crate) fn assert_tasks<Fut>(expected_tasks: Vec<ExpectedTask>, future: Fut)
+where
+    Fut: Future + Send + 'static,
+    Fut::Output: Send + 'static,
+{
+    run_test(expected_tasks, future)
+}
--- a/console-subscriber/tests/support/state.rs
+++ b/console-subscriber/tests/support/state.rs
@@ -53,6 +53,13 @@ impl TestState {
     ///
     /// This function will panic if the underlying channel gets closed.
     pub(super) async fn wait_for_step(&mut self, desired_step: TestStep) {
+        self.update_step();
+        tracing::info!(
+            target: "console_test::support::state",
+            "wait_for_step: {current} -> {desired_step}",
+            current = self.step,
+        );
+
         loop {
             if self.step >= desired_step {
                 break;
@@ -73,6 +80,11 @@ impl TestState {
     /// Check whether the desired step has been reached without blocking.
     pub(super) fn try_wait_for_step(&mut self, desired_step: TestStep) -> bool {
         self.update_step();
+        tracing::info!(
+            target: "console_test::support::state",
+            "try_wait_for_step: {current} -> {desired_step}",
+            current = self.step,
+        );
 
         self.step == desired_step
     }
@@ -90,6 +102,11 @@ impl TestState {
     #[track_caller]
     pub(super) fn advance_to_step(&mut self, next_step: TestStep) {
         self.update_step();
+        tracing::info!(
+            target: "console_test::support::state",
+            "advance_to_step: {current} -> {next_step}",
+            current = self.step,
+        );
 
         if self.step >= next_step {
             panic!(
--- a/console-subscriber/tests/support/state.rs
+++ b/console-subscriber/tests/support/state.rs
@@ -47,16 +47,13 @@ impl TestState {
         }
     }
 
-    /// Block asynchronously until the desired step has been reached.
+    /// Wait asynchronously until the desired step has been reached.
     ///
     /// # Panics
     ///
     /// This function will panic if the underlying channel gets closed.
     pub(super) async fn wait_for_step(&mut self, desired_step: TestStep) {
-        loop {
-            if self.step >= desired_step {
-                break;
-            }
+        while self.step < desired_step {
 
             match self.receiver.recv().await {
                 Ok(step) => self.step = step,
@@ -70,8 +67,8 @@ impl TestState {
         }
     }
 
-    /// Check whether the desired step has been reached without blocking.
-    pub(super) fn try_wait_for_step(&mut self, desired_step: TestStep) -> bool {
+    /// Returns `true` if the current step is `desired_step` or later.
+    pub(super) fn is_step(&mut self, desired_step: TestStep) -> bool {
         self.update_step();
 
         self.step == desired_step
@@ -86,25 +83,23 @@ impl TestState {
     /// # Panics
     ///
     /// This method will panic if the test state is not at the step prior to
-    /// `next_step` or if the underlying channel is closed.
+    /// `next_step`, or if the underlying channel is closed.
     #[track_caller]
     pub(super) fn advance_to_step(&mut self, next_step: TestStep) {
         self.update_step();
 
-        if self.step >= next_step {
-            panic!(
-                "cannot advance to previous or current step! current step: {current}, next step: {next_step}",
-                current = self.step);
-        }
+        assert!(
+            self.step < next_step,
+            "cannot advance to previous or current step! current step: {current}, next step: {next_step}",
+            current = self.step,
+        );
 
         match (&self.step, &next_step) {
             (TestStep::Start, TestStep::ServerStarted) |
             (TestStep::ServerStarted, TestStep::ClientConnected) |
             (TestStep::ClientConnected, TestStep::TestFinished) |
             (TestStep::TestFinished, TestStep::UpdatesRecorded) => {},
-            (_, _) => panic!(
-                "cannot advance more than one step! current step: {current}, next step: {next_step}",
-                current = self.step),
+            (current, _) => panic!("cannot advance more than one step! current step: {current}, next step: {next_step}"),
         }
 
         self.sender
--- a/console-subscriber/tests/support/state.rs
+++ b/console-subscriber/tests/support/state.rs
@@ -89,7 +89,8 @@ impl TestState {
 
         assert!(
             self.step < next_step,
-            "cannot advance to previous or current step! current step: {current}, next step: {next_step}",
+            "console-test error: cannot advance to previous or current step! \
+            current step: {current}, next step: {next_step}",
             current = self.step,
         );
 
@@ -98,12 +99,18 @@ impl TestState {
             (TestStep::ServerStarted, TestStep::ClientConnected) |
             (TestStep::ClientConnected, TestStep::TestFinished) |
             (TestStep::TestFinished, TestStep::UpdatesRecorded) => {},
-            (current, _) => panic!("cannot advance more than one step! current step: {current}, next step: {next_step}"),
+            (current, _) => panic!(
+                "console-test error: test cannot advance more than one step! \
+                current step: {current}, next step: {next_step}"
+            ),
         }
 
         self.sender
             .send(next_step)
-            .expect("failed to send the next test step, did the test abort?");
+            .expect(
+                "console-test error: failed to send the next test step, \
+                did the test abort?"
+            );
     }
 
     fn update_step(&mut self) {
--- a/console-subscriber/tests/support/state.rs
+++ b/console-subscriber/tests/support/state.rs
@@ -54,7 +54,6 @@ impl TestState {
     /// This function will panic if the underlying channel gets closed.
     pub(super) async fn wait_for_step(&mut self, desired_step: TestStep) {
         while self.step < desired_step {
-
             match self.receiver.recv().await {
                 Ok(step) => self.step = step,
                 Err(RecvError::Lagged(_)) => {
--- a/console-subscriber/tests/support/state.rs
+++ b/console-subscriber/tests/support/state.rs
@@ -39,8 +39,7 @@ pub(super) struct TestState {
 
 impl TestState {
     pub(super) fn new() -> Self {
-        // Capacity sufficient for all states (will never lose anything)
-        let (sender, receiver) = broadcast::channel(5);
+        let (sender, receiver) = broadcast::channel(1);
         Self {
             receiver,
             sender,
@@ -54,16 +53,6 @@ impl TestState {
     ///
     /// This function will panic if the underlying channel gets closed.
     pub(super) async fn wait_for_step(&mut self, desired_step: TestStep) {
-        {
-            let _guard = tracing::info_span!("wait_for_step").entered();
-            self.update_step();
-        }
-        tracing::info!(
-            target: "console_test::support::state",
-            "wait_for_step: {current} -> {desired_step}",
-            current = self.step,
-        );
-
         loop {
             if self.step >= desired_step {
                 break;
@@ -83,15 +72,7 @@ impl TestState {
 
     /// Check whether the desired step has been reached without blocking.
     pub(super) fn try_wait_for_step(&mut self, desired_step: TestStep) -> bool {
-        {
-            let _guard = tracing::info_span!("try_wait_for_step").entered();
-            self.update_step();
-        }
-        tracing::info!(
-            target: "console_test::support::state",
-            "try_wait_for_step: {current} -> {desired_step}",
-            current = self.step,
-        );
+        self.update_step();
 
         self.step == desired_step
     }
@@ -108,15 +89,7 @@ impl TestState {
     /// `next_step` or if the underlying channel is closed.
     #[track_caller]
     pub(super) fn advance_to_step(&mut self, next_step: TestStep) {
-        {
-            let _guard = tracing::info_span!("advance_to_step").entered();
-            self.update_step();
-        }
-        tracing::info!(
-            target: "console_test::support::state",
-            "advance_to_step: {current} -> {next_step}",
-            current = self.step,
-        );
+        self.update_step();
 
         if self.step >= next_step {
             panic!(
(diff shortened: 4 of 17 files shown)


console-subscriber/tests/support/state.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub(super) fn new() -> Self {
+        let (sender, receiver) = broadcast::channel(1);
+        Self {
+            receiver,
+            sender,
+            step: TestStep::Start,
+        }
+    }
+
+    /// Block asynchronously until the desired step has been reached.
```
hawkw: nit: i would probably not say "block asynchronously" --- if this is asynchronous, it doesn't block, by definition. maybe:
```suggestion
    /// Wait asynchronously until the desired step has been reached.
```

console-subscriber/tests/support/state.rs
```diff
@@ -0,0 +1,139 @@
...
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if the underlying channel gets closed.
+    pub(super) async fn wait_for_step(&mut self, desired_step: TestStep) {
+        loop {
+            if self.step >= desired_step {
+                break;
+            }
```
hawkw: nit: isn't this just a `while` loop?
```suggestion
        while self.step < desired_step {
```

console-subscriber/tests/support/state.rs
```diff
@@ -0,0 +1,139 @@
...
+                }
+                Err(RecvError::Closed) => {
+                    panic!("failed to receive current step, waiting for step: {desired_step}, did the test abort?");
+                }
+            }
+        }
+    }
+
+    /// Check whether the desired step has been reached without blocking.
+    pub(super) fn try_wait_for_step(&mut self, desired_step: TestStep) -> bool {
```
hawkw: nit: i get why this ended up being named "try_wait_for_step", because it's a version of `wait_for_step` that doesn't wait, but...couldn't we just call it
```suggestion
    /// Returns `true` if the current step is `desired_step` or later.
    pub(super) fn is_step(&mut self, desired_step: TestStep) -> bool {
```
(in particular, I would expect a function called `try_{foo}` to return a `Result`...)

console-subscriber/tests/support/state.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// Advance to the next step.
+    ///
+    /// The test must be at the step prior to the next step before starting.
+    /// Being in a different step is likely to indicate a logic error in the
+    /// test framework.
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if the test state is not at the step prior to
+    /// `next_step` or if the underlying channel is closed.
```
hawkw:
```suggestion
    /// `next_step`, or if the underlying channel is closed.
```

console-subscriber/tests/support/state.rs
```diff
@@ -0,0 +1,139 @@
...
+    #[track_caller]
+    pub(super) fn advance_to_step(&mut self, next_step: TestStep) {
+        self.update_step();
+
+        if self.step >= next_step {
+            panic!(
+                "cannot advance to previous or current step! current step: {current}, next step: {next_step}",
+                current = self.step);
+        }
```
hawkw: nit: `if <CONDITION> { panic!(...) }` could be simplified to `assert!(<CONDITION>, ...)`:
```suggestion
        assert!(
            self.step < next_step,
            "cannot advance to previous or current step! current step: {current}, next step: {next_step}",
            current = self.step,
        );
```

console-subscriber/tests/support/state.rs
```diff
@@ -0,0 +1,139 @@
...
+
+        match (&self.step, &next_step) {
+            (TestStep::Start, TestStep::ServerStarted) |
+            (TestStep::ServerStarted, TestStep::ClientConnected) |
+            (TestStep::ClientConnected, TestStep::TestFinished) |
+            (TestStep::TestFinished, TestStep::UpdatesRecorded) => {},
+            (_, _) => panic!(
+                "cannot advance more than one step! current step: {current}, next step: {next_step}",
+                current = self.step),
```
hawkw: nit: slightly weird formatting here, also we could maybe avoid the `current = ` if we just wrote
```suggestion
            (current, _) => panic!("cannot advance more than one step! current step: {current}, next step: {next_step}"),
```

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,318 @@
...
+
+            async move {
+                if let Some(client) = client {
+                    Ok(client)
+                } else {
+                    Err(std::io::Error::new(
+                        std::io::ErrorKind::Other,
+                        "Client already taken",
+                    ))
+                }
```
hawkw: nit: isn't this just `Option::ok_or_else`?
```suggestion
                client.ok_or_else(|| std::io::Error::new(std::io::ErrorKind::Other, "Client already taken"))
```

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,318 @@
...
+/// # Test State
+///
+/// 1. Waits for: `ServerStarted`
+/// 2. Advances to: `ClientConnected`
+async fn console_client(client_stream: DuplexStream, mut test_state: TestState) -> Vec<ActualTask> {
+    test_state.wait_for_step(TestStep::ServerStarted).await;
+
+    let mut client_stream = Some(client_stream);
+    let channel = Endpoint::try_from("http://[::]:6669")
+        .expect("Could not create endpoint")
```
hawkw: we never actually connect to this port in the test, and just use a `DuplexStream` rather than an actual network connection, right?

maybe worth documenting that:
```suggestion
    // Note: we won't actually try to connect to this port on localhost,
    // because we will call `connect_with_connector` with a service that
    // just returns the `DuplexStream`, instead of making an actual 
    // network connection.
    let endpoint = Endpoint::try_from("http://[::]:6669")
        .expect("Could not create endpoint");
    let channel = endpoint
```

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,318 @@
...
+) -> Vec<ActualTask> {
+    let mut client = InstrumentClient::new(client_channel);
+
+    let mut stream = loop {
+        let request = tonic::Request::new(InstrumentRequest {});
+        match client.watch_updates(request).await {
+            Ok(stream) => break stream.into_inner(),
+            Err(err) => panic!("Client cannot connect to watch updates: {err}"),
+        }
+    };
```
hawkw: i don't understand why this is a loop? it looks like we will only ever make one `watch_updates` request, since we either break if it's `Ok` or panic if it's `Err`. can't this just be:
```suggestion
    let mut stream = match client.watch_updates(tonic::Request::new(InstrumentRequest {})).await {
        Ok(stream) => stream.into_inner(),
        Err(err) => panic!("Client cannot connect to watch updates: {err}"),
    };
```

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,318 @@
...
+                            task.wakes = stats.wakes;
+                            task.self_wakes = stats.self_wakes;
+                        }
+                    }
+                }
+            }
+            Err(e) => {
+                panic!("update stream error: {}", e);
+            }
+        }
```
hawkw: style nit, take it or leave it: i'm not a huge fan of the amount of "rightward drift" in this code, would something like this work instead?
```suggestion
    while let Some(update) = stream.next().await {
        let update = match update {
            Ok(u) => u,
            Err(e) => panic!("update stream error: {e}"),
        };

        if let Some(task_update) = &update.task_update {
            for new_task in &task_update.new_tasks {
                let id = match new_task.id {
                    Some(id) => id,
                    None => continue,
                };
                let mut actual_task = ActualTask::new(id.id);
                let name = new_task.fields.iter().find_map(|field| {
                    match field.name? {
                        console_api::field::Name::StrName(name) if name == "task.name" => Some(field.value.clone()),
                        _ => None,
                    }
                });
                actual_task.name = match name {
                    Some(Value::DebugVal(value)) => Some(value.clone()),
                    Some(Value::StrVal(value)) => Some(value.clone()),
                    _ => None, // Anything that isn't string-like shouldn't be used as a name.
                };
                tasks.insert(actual_task.id, actual_task);
            }

            for (id, stats) in &task_update.stats_update {
                if let Some(task) = tasks.get_mut(id) {
                    task.wakes = stats.wakes;
                    task.self_wakes = stats.self_wakes;
                }
            }
        }
```

console-subscriber/tests/support/task.rs
```diff
@@ -0,0 +1,228 @@
...
+
+/// An expected task.
+///
+/// This struct contains the fields that an expected task will attempt to match
+/// actual tasks on, as well as the expectations that will be used to validate
+/// which the actual task is as expected.
+#[derive(Clone, Debug)]
+pub(crate) struct ExpectedTask {
+    match_name: Option<String>,
+
```
hawkw: nit: bonus whitespace here:
```suggestion
```

console-subscriber/tests/support/task.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+}
+
+impl ExpectedTask {
+    /// Returns whether or not an actual task matches this expected task.
+    ///
+    /// All matching rules will be run, if they all succeed, then `true` will
+    /// be returned, otherwise `false`.
+    pub(super) fn matches_actual_task(&self, actual_task: &ActualTask) -> bool {
+        if let Some(match_name) = &self.match_name {
```
hawkw, line 100: am I correct that this is saying that if an `ExpectedTask`'s `match_name` is not `Some`, the `ExpectedTask` will basically never do anything of value? why is `match_name` optional, then? couldn't we just always have there be a `match_name`, and have a constructor that takes the name?

console-subscriber/tests/support/task.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        if let Some(expected_self_wakes) = self.expect_self_wakes {
+            no_expectations = false;
+            if expected_self_wakes != actual_task.self_wakes {
+                return Err(TaskValidationFailure {
+                    expected: self.clone(),
+                    actual: Some(actual_task.clone()),
+                    failure: format!(
+                        "{self}: expected `self_wakes` to be {expected_self_wakes}, but actual was {actual_self_wakes}",
+                        actual_self_wakes = actual_task.self_wakes),
```
hawkw: weird formatting
```suggestion
                    failure: format!(
                        "{self}: expected `self_wakes` to be {expected_self_wakes}, but actual was {actual_self_wakes}",
                        actual_self_wakes = actual_task.self_wakes,
                    ),
```

console-subscriber/tests/support/task.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        if let Some(expected_wakes) = self.expect_wakes {
+            no_expectations = false;
+            if expected_wakes != actual_task.wakes {
+                return Err(TaskValidationFailure {
+                    expected: self.clone(),
+                    actual: Some(actual_task.clone()),
+                    failure: format!(
+                        "{self}: expected `wakes` to be {expected_wakes}, but actual was {actual_wakes}",
+                        actual_wakes = actual_task.wakes),
```
hawkw: weird formatting:
```suggestion
                    failure: format!(
                        "{self}: expected `wakes` to be {expected_wakes}, but actual was {actual_wakes}",
                        actual_wakes = actual_task.wakes,
                    ),
```

console-subscriber/tests/support/task.rs
```diff
@@ -0,0 +1,228 @@
...
+                });
+            }
+        }
+
+        if no_expectations {
+            return Err(TaskValidationFailure {
+                expected: self.clone(),
+                actual: Some(actual_task.clone()),
+                failure: format!(
+                    "{self}: no expectations set, if you want to just expect that a matching task is present, use `expect_present()`")
```
hawkw: weird formatting:
```suggestion
                failure: format!(
                    "{self}: no expectations set, if you want to just expect that a matching task is present, use `expect_present()`",
                ),
```

console-subscriber/tests/support/task.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+}
+
+impl fmt::Display for ExpectedTask {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let fields = match &self.match_name {
+            Some(name) => format!("name={name}"),
+            None => "(no fields to match on)".into(),
+        };
+        write!(f, "Task<{fields}>")
```
hawkw: nit, TIOLI: using angle brackets here feels a little weird to me, since i would expect them to represent a type parameter? maybe:
```suggestion
        write!(f, "Task< {{ {fields} }}")
```

console-subscriber/tests/support/task.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+}
+
+/// An expected task.
+///
+/// This struct contains the fields that an expected task will attempt to match
+/// actual tasks on, as well as the expectations that will be used to validate
+/// which the actual task is as expected.
+#[derive(Clone, Debug)]
+pub(crate) struct ExpectedTask {
```
hawkw, line 76: here's a thought: maybe constructing an `ExpectedTask` should capture the `core::panic::Location` where the `ExpectedTask` was constructed? we could have an `ExpectedTask` constructor with `#[track_caller]` on it, and include that location in the failure message. that way, if there are multiple `ExpectedTasks` in a test, it's clear which one was involved in a test failure.

this isn't a blocker for this PR, but it might be nice to add in the future...

console-subscriber/tests/support/task.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+}
+
+impl fmt::Debug for TaskValidationFailure {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &self.actual {
+            Some(actual) => write!(
+                f,
+                "Task Validation Failed!\n  Expected Task: {expected:?}\n  Actual Task:   {actual:?}\n  Failure:       {failure}",
+                expected = self.expected, failure = self.failure),
```
hawkw: nit: weird formatting
```suggestion
                expected = self.expected,
                failure = self.failure,
            ),
```

console-subscriber/tests/support/task.rs
```diff
@@ -0,0 +1,228 @@
...
+            None => write!(
+                f,
+                "Task Validation Failed!\n  Expected Task: {expected:?}\n  Failure:       {failure}",
+                expected = self.expected, failure = self.failure),
```
hawkw:
```suggestion
                "Task Validation Failed!\n  Expected Task: {expected:?}\n  Failure:       {failure}",
                expected = self.expected,
                failure = self.failure,
            ),
```

hawkw (commented):
overall, this looks great, thanks for working on this! i left a bunch of relatively minor comments, but overall, I'd be happy to merge this change!

hds:
> overall, this looks great, thanks for working on this! i left a bunch of relatively minor comments, but overall, I'd be happy to merge this change!

@hawkw Thanks for the review!

Unfortunately, I just realised today that the CI is telling me that I've got a race condition (and I really thought I'd removed all of them) which I'm not seeing locally and is also behaving differently on different platforms - some platforms fail, some platform actually hang forever it seems. So this PR is going to need a bit of work to find and fix that issue.

hawkw:
@hds is this branch ready for a review now?

hds:
@hawkw unfortunately not yet. I've still got a problem with the test run on CI when executing on Ubuntu. It hangs because the message that the test has been run never gets to the instrumentation client.

Work has been busy lately, so getting to the bottom of this problem has been slow going.

hawkw:
@hds okay, cool, thanks for letting me know!

console-subscriber/tests/support/state.rs
hds: It totally is...

console-subscriber/tests/support/state.rs
hds: Yep, agree with this.

console-subscriber/tests/support/subscriber.rs
hds: Yeah, that's a good point. Thanks!

console-subscriber/tests/support/subscriber.rs
hds: Good catch, I think the loop must have been left over from a previous fix attempt that didn't work.

console-subscriber/tests/support/task.rs
hds: My idea is that in the future we will probably (maybe) want to match tasks on things other than the name. For example, I want to match a task associated with a resource or async op. While it's not implemented yet, it seemed to make sense to follow the pattern that a task can match on different things from the outset.

console-subscriber/tests/support/task.rs
hds: That's a really nice idea. I'll look at that in a follow up.

console-subscriber/tests/support/subscriber.rs
hds: This is much nicer, thank you! I didn't see this comment until after further changes, so I couldn't apply it, but I copied it pretty much as is.

hds:
@hawkw This one is ready for re-review when you have a moment. Thanks!

hawkw:
> @hawkw This one is ready for re-review when you have a moment. Thanks!

awesome, thanks for all the time you've spent on this!

console-subscriber/tests/framework.rs
```diff
@@ -0,0 +1,184 @@
+//! Framework tests
+//!
+//! The tests in this module are here to verify the testing framework itself.
```
hawkw, line 3: including this is a great idea!

console-subscriber/Cargo.toml
```diff
@@ -55,6 +55,7 @@ crossbeam-channel = "0.5"
 
 [dev-dependencies]
 tokio = { version = "^1.21", features = ["full", "rt-multi-thread"] }
+tower = "0.4"
```
hawkw: do we need all the default features of `tower`, or can we disable them and only use the ones we need? this isn't a huge deal, since it's only a dev-only dependency, but it might make building the tests slightly faster. not important.

console-subscriber/tests/support/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+///
+/// This function is equivalent to calling [`assert_tasks`] with a vector
+/// containing a single task.
+///
+/// # Panics
+///
+/// This function will panic if the expectations on the expected task are not
+/// met or if a matching task is not recorded.
+#[track_caller]
+#[allow(dead_code)]
```
hawkw, line 23: why do we need the `allow(dead_code)` on these? is this because some test files only use `assert_task` and others only use `assert_tasks`?

console-subscriber/tests/support/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+/// are not met or if matching tasks are not recorded for all expected tasks.
+#[track_caller]
+#[allow(dead_code)]
+pub(crate) fn assert_tasks<Fut>(expected_tasks: Vec<ExpectedTask>, future: Fut)
+where
+    Fut: Future + Send + 'static,
+    Fut::Output: Send + 'static,
+{
+    run_test(expected_tasks, future)
+}
```
hawkw, line 47: tiny nit, take it or leave it: is there a reason this is a whole additional function, rather than just being a re-export of `run_test`? we could
```rust
pub use subscriber::run_test as assert_tasks;
```
if we want it to be named `assert_tasks` (but also, we could just name the original function that...)

console-subscriber/tests/support/state.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// This function will panic if the underlying channel gets closed.
+    pub(super) async fn wait_for_step(&mut self, desired_step: TestStep) {
+        while self.step < desired_step {
+            match self.receiver.recv().await {
+                Ok(step) => self.step = step,
+                Err(RecvError::Lagged(_)) => {
+                    // we don't mind being lagged, we'll just get the latest state
+                }
+                Err(RecvError::Closed) => {
+                    panic!("failed to receive current step, waiting for step: {desired_step}, did the test abort?");
```
hawkw: does this panic indicate an issue in the test framework, or a normal test failure? if this is a test framework issure, can we add something to the panic message that states "this is likely a test framework bug" or similar?

console-subscriber/tests/support/state.rs
```diff
@@ -0,0 +1,133 @@
...
+            (TestStep::Start, TestStep::ServerStarted) |
+            (TestStep::ServerStarted, TestStep::ClientConnected) |
+            (TestStep::ClientConnected, TestStep::TestFinished) |
+            (TestStep::TestFinished, TestStep::UpdatesRecorded) => {},
+            (current, _) => panic!("cannot advance more than one step! current step: {current}, next step: {next_step}"),
+        }
+
+        self.sender
+            .send(next_step)
+            .expect("failed to send the next test step, did the test abort?");
```
hawkw: if these panics indicate a bug in the test framework rather than in the test itself, can we add something like:
```suggestion
        assert!(
            self.step < next_step,
            "console-test error: cannot advance to previous or current step! \
            current step: {current}, next step: {next_step}",
            current = self.step,
        );

        match (&self.step, &next_step) {
            (TestStep::Start, TestStep::ServerStarted) |
            (TestStep::ServerStarted, TestStep::ClientConnected) |
            (TestStep::ClientConnected, TestStep::TestFinished) |
            (TestStep::TestFinished, TestStep::UpdatesRecorded) => {},
            (current, _) => panic!(
                "console-test error: test cannot advance more than one step! \
                current step: {current}, next step: {next_step}"
            ),
        }

        self.sender
            .send(next_step)
            .expect(
                "console-test error: failed to send the next test step, \
                did the test abort?"
            );
```

console-subscriber/tests/support/state.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    fn update_step(&mut self) {
+        loop {
+            match self.receiver.try_recv() {
+                Ok(step) => self.step = step,
+                Err(TryRecvError::Lagged(_)) => {
+                    // we don't mind being lagged, we'll just get the latest state
+                }
+                Err(TryRecvError::Closed) => {
+                    panic!("failed to update current step, did the test abort?")
```
hawkw: is this an internal test framework error, or a test failure?

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    let (client_stream, server_stream) = tokio::io::duplex(1024);
+    let (console_layer, server) = console_subscriber::ConsoleLayer::builder().build();
+    let registry = tracing_subscriber::registry().with(console_layer);
+
+    let mut test_state = TestState::new();
+    let mut test_state_test = test_state.clone();
+
+    let join_handle = thread::Builder::new()
+        .name("console::subscriber".into())
```
hawkw:
```suggestion
    
    let thread_name = {
        // Include the name of the test thread in the spawned subscriber thread,
        // to make it clearer which test it belongs to.
        let test = thread::current().name().unwrap_or("<unknown test">);
        format!("{test}-console::subscriber")
    }
    let join_handle = thread::Builder::new()
        .name(thread_name)
```

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,323 @@
...
+                    .spawn(console_client(client_stream, test_state.clone()))
+                    .expect("console-test error: could not spawn 'console-client' task")
+                    .await
+                    .expect("console-test error: failed to await 'console-client' task");
+
+                test_state.advance_to_step(TestStep::UpdatesRecorded);
+                actual_tasks
+            })
+        })
+        .expect("console subscriber could not spawn thread");
```
hawkw: this is an internal test error, right?

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,323 @@
...
+    let (client_stream, server_stream) = tokio::io::duplex(1024);
+    let (console_layer, server) = console_subscriber::ConsoleLayer::builder().build();
+    let registry = tracing_subscriber::registry().with(console_layer);
+
+    let mut test_state = TestState::new();
+    let mut test_state_test = test_state.clone();
+
+    let join_handle = thread::Builder::new()
+        .name("console::subscriber".into())
+        .spawn(move || {
```
hawkw, line 71: i was going to leave a comment asking why we needed to spawn this stuff in a background thread, but i realized that it's to ensure that any `tracing` spans emitted by the client and console server tasks aren't recorded by the test console subscriber. can we maybe include a comment explaining this? something like:
```suggestion
        // Run the test's console server and client tasks in a separate thread
        // from the main test, ensuring that any `tracing` emitted by the
        // console worker and the client are not collected by the subscriber
        // under test.
        .spawn(move || {
```

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,323 @@
...
+    } = server.into_parts();
+    let aggregate = task::Builder::new()
+        .name("console::aggregate")
+        .spawn(aggregator.run())
+        .expect("client-console error: couldn't spawn aggregator");
+    Server::builder()
+        .add_service(service)
+        .serve_with_incoming(futures::stream::iter(vec![Ok::<_, std::io::Error>(
+            server_stream,
+        )]))
```
hawkw, line 170: very tiny, unimportant nit: could be simplified to just
```suggestion
        .serve_with_incoming(futures::stream::once(Ok::<_, std::io::Error>(server_stream)))
```

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,323 @@
...
+        .await
+        .expect("client-console error: couldn't start instrument server.");
```
hawkw: tiny nit:
```suggestion
        .expect("console-test error: couldn't spawn aggregator");
    Server::builder()
        .add_service(service)
        .serve_with_incoming(futures::stream::iter(vec![Ok::<_, std::io::Error>(
            server_stream,
        )]))
        .await
        .expect("console-test error: couldn't start instrument server.");
```

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,323 @@
...
+            let client = client_stream.take();
+
+            async move {
+                client.ok_or_else(|| {
+                    std::io::Error::new(std::io::ErrorKind::Other, "Client already taken")
+                })
+            }
+        }))
+        .await
+        .expect("client-console error: couldn't create client");
```
hawkw: nit: maybe
```suggestion
        .expect("console-test client error: couldn't create client");
```

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,323 @@
...
+    // network connection.
+    let endpoint = Endpoint::try_from("http://[::]:6669").expect("Could not create endpoint");
+    let channel = endpoint
+        .connect_with_connector(service_fn(move |_: Uri| {
+            let client = client_stream.take();
+
+            async move {
+                client.ok_or_else(|| {
+                    std::io::Error::new(std::io::ErrorKind::Other, "Client already taken")
+                })
```
hawkw: nit, take it or leave it: probably better to panic if the client is already taken, since i believe the server will continue accepting connections if it sees a conn error (IIRC?)
```suggestion
                Ok(client.expect("console-test error: client already taken (this shouldn't happen)"))
```

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,323 @@
...
+    mut test_state: TestState,
+) -> Vec<ActualTask> {
+    let mut client = InstrumentClient::new(client_channel);
+
+    let mut stream = match client
+        .watch_updates(tonic::Request::new(InstrumentRequest {}))
+        .await
+    {
+        Ok(stream) => stream.into_inner(),
+        Err(err) => panic!("Client cannot connect to watch updates: {err}"),
```
hawkw: is this an internal error or a test failure?

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,323 @@
...
+    };
+
+    let mut tasks = HashMap::new();
+
+    let signal_task = ExpectedTask::default().match_name(END_SIGNAL_TASK_NAME.into());
+    let mut signal_task_read = false;
+    while let Some(update) = stream.next().await {
+        let update = update.expect("update stream error");
```
hawkw: nit: maybe
```suggestion
        let update = update.expect("console-test error: update stream error");
```

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,323 @@
...
+        .watch_updates(tonic::Request::new(InstrumentRequest {}))
+        .await
+    {
+        Ok(stream) => stream.into_inner(),
+        Err(err) => panic!("Client cannot connect to watch updates: {err}"),
+    };
+
+    let mut tasks = HashMap::new();
+
+    let signal_task = ExpectedTask::default().match_name(END_SIGNAL_TASK_NAME.into());
```
hawkw, line 253: this kind of in-band signalling feels a _bit_ sketchy --- would it be better to indicate that the test has ended using a `tokio::sync::oneshot` or something and `select!`ing over either `stream.next()` or the oneshot? that way, we don't have to rely on the console subscriber correctly recording the "end signal task" before we can end the test. 

i'm okay with merging this as-is, if a rewrite would be too time-consuming, since the current approach does seem to _work_. but, i would prefer to separate the end-of-test signal from the actual test tasks...

hawkw: or, if there's a reason we need this to be in-band (is it necessary to ensure that *all* the tasks have definitely been read, rather than just stopping as soon as we trigger a oneshot?) can we add a comment somewhere explaining why this is the best approach?

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,323 @@
...
+                            Some(field.value.clone())
+                        }
+                        _ => None,
+                    })
+                    .flatten();
+                actual_task.name = match name {
+                    Some(Value::DebugVal(value)) => Some(value.clone()),
+                    Some(Value::StrVal(value)) => Some(value.clone()),
+                    _ => None, // Anything that isn't string-like shouldn't be used as a name.
+                };
```
hawkw: i`m normally a big fan of cool iterator tricks, but isn't it simpler to just write something like
```suggestion
                for field in new_task.fields {
                    match field.name.as_ref() {
                        console_api::field::Name::StrName(name) if name == "task.name" => {
                            actual_task.name = match field.value.as_ref() {
                                Value::DebugVal(value) => actual_task.name = Some(value),
                                Value::StrVal(value) => actual_task.name = Some(value),
                                _ => continue;
                            }
                        },
                        _ => {}
                    }
                }
```

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,323 @@
...
+                    None => continue,
+                };
+                let name = new_task
+                    .fields
+                    .iter()
+                    .find_map(|field| match field.name.as_ref()? {
+                        console_api::field::Name::StrName(name) if name == "task.name" => {
+                            Some(field.value.clone())
+                        }
+                        _ => None,
```
hawkw: does this actually work? my assumption was that task names would generally be represented on the wire by metadata indices...i think we need to be handling the case where the name is a metadata index.

hawkw: update: oh, i looked at the way task fields are currently recorded, and it looks like we actually always send the string name...we should probably fix that so that task update messages are more efficient on the wire: https://github.com/tokio-rs/console/blob/d1764046e55a55ce79228c1ce7fdc7153484d601/console-subscriber/src/visitors.rs#L306-L346

but, the current code does work for now, and (i would hope) that changing the way tasks are serialized would break ~all the tests...

console-subscriber/tests/support/subscriber.rs
```diff
@@ -0,0 +1,323 @@
...
+    expected_tasks: Vec<ExpectedTask>,
+    actual_tasks: Vec<ActualTask>,
+) -> Result<(), TestFailure> {
+    let failures: Vec<_> = expected_tasks
+        .iter()
+        .map(|expected| validate_expected_task(expected, &actual_tasks))
+        .filter_map(|r| match r {
+            Ok(_) => None,
+            Err(validation_error) => Some(validation_error),
+        })
```
hawkw: very tiny nit: there's actually a [method on `Result`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.err) that just does this:
```suggestion
        .filter_map(Result::err)
```

console-subscriber/tests/support/task.rs
```diff
@@ -0,0 +1,234 @@
...
+        match &self.actual {
+            Some(actual) => write!(
+                f,
+                "Task Validation Failed!\n  Expected Task: {expected:?}\n  Actual Task:   {actual:?}\n  Failure:       {failure}",
+                expected = self.expected,
+                failure = self.failure,
+            ),
+            None => write!(
+                f,
+                "Task Validation Failed!\n  Expected Task: {expected:?}\n  Failure:       {failure}",
```
hawkw: maybe this message should clearly indicate that no actual task was recorded?

hawkw (approved):
overall, this looks good to me! i left a bunch of small suggestions, but none of them are major blockers.

console-subscriber/tests/framework.rs
hds: I learned the hard way that test frameworks should have tests as well. (-;

console-subscriber/tests/support/mod.rs
hds: It was more a case of putting the "internal public" functions together to make the documentation clearer. Otherwise we'd have "public" docs here and on the `run_test` function. Not sure what the best practice is in this case to be honest.

console-subscriber/tests/support/mod.rs
hds: Yes, exactly. Because each integration test is compiled individually (or at least, this is common when running a single test), tests that don't use one of these functions caused a warning, which was annoying.

console-subscriber/tests/support/subscriber.rs
hds: This is a great idea, thanks!

console-subscriber/tests/support/subscriber.rs
hds: Yeah, I really didn't want to do this. In fact, the `TestState` struct is supposed to do the job of the oneshot channel you describe. However, when the `TestFinished` state is reached, we don't have all the events from the test yet.

The reason for this is that the update loop in the aggregator tends to yield back to the runtime various times during the first full update. This means that once the test has finished, we have N update loops (even on the client side) until all of the events have been processed, and we don't know the value of N ahead of time (it seems to change from platform to platform too).

https://github.com/tokio-rs/console/blob/d1764046e55a55ce79228c1ce7fdc7153484d601/console-subscriber/src/aggregator/mod.rs#L223

I'll add a comment to explain this in the code.

console-subscriber/tests/support/subscriber.rs
hds: Yeah, I wasn't sure about this. This is something that should be improved, and that should be caught by the tests too, as you said.

console-subscriber/tests/support/subscriber.rs
hds: Nice! Hadn't used that before.

console-subscriber/tests/support/task.rs
hds: Created an issue #461 to remind me.

console-subscriber/Cargo.toml
hds: Good idea, turns out we don't need any features from tower at all!

console-subscriber/tests/support/state.rs
hds: Done.

console-subscriber/tests/support/state.rs
hds: Internal error, marked it as such.

console-subscriber/tests/support/subscriber.rs
hds: Yes, marked as such.

console-subscriber/tests/support/subscriber.rs
hds: Yes, marked as such.

console-subscriber/tests/support/task.rs
hds: Done.

console-subscriber/tests/support/subscriber.rs
hds: I created an issue for this improvement: #462.

hds:
> overall, this looks good to me! i left a bunch of small suggestions, but none of them are major blockers.

@hawkw Thank you so much for all these suggestions! I really appreciate the effort, the PR is much better for them.

PR #455, opened 2023-08-03
fix(console): fix task detail view Id, now using remote tokio::task::Id

inevity:
Now home page task list show id using remote tokio::task::Id, but task detail view show Id using local sequential id in the console client, so both id is not same.
The pr fix the console/client to show id using remote tokio::task::Id.

### DIFF
--- a/tokio-console/src/view/task.rs
+++ b/tokio-console/src/view/task.rs
@@ -142,7 +142,7 @@ impl TaskView {
         let mut overview = Vec::with_capacity(8);
         overview.push(Spans::from(vec![
             bold("ID: "),
-            Span::raw(format!("{} ", task.task_id())),
+            Span::raw(format!("{} ", task.id_str())),
             task.state().render(styles),
         ]));
 
--- a/tokio-console/src/view/task.rs
+++ b/tokio-console/src/view/task.rs
@@ -142,7 +142,7 @@ impl TaskView {
         let mut overview = Vec::with_capacity(8);
         overview.push(Spans::from(vec![
             bold("ID: "),
-            Span::raw(format!("{} ", task.id())),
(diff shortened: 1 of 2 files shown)


tokio-console/src/view/task.rs
```diff
@@ -142,7 +142,7 @@ impl TaskView {
         let mut overview = Vec::with_capacity(8);
         overview.push(Spans::from(vec![
             bold("ID: "),
-            Span::raw(format!("{} ", task.id())),
+            Span::raw(format!("{} ", task.task_id())),
```
hds: We already have a method which returns a `&str` of the tokio task ID (and so doesn't require `unwrap_or`).

```suggestion
            Span::raw(format!("{} ", task.id_str())),
```

hds (changes requested):
Thank you for finding and fixing this error! Your PR can be simplified a little bit using the same method as in the list view.

The tokio task ID is already stored on the `state::Task` struct as `id_str` (and already parsed to a string for quick rendering) and exposed by a method of the same name.

You can use that direclty instead of the new `task_id()` method.

If you would like to do so, it would make sense to rename the existing field and method to `task_id_str`. But this is optional.

tokio-console/src/view/task.rs
inevity: Chang to use id_str().
As for rename the existing field from id_str to task_id_str. Found AsyncOpStats have a member named task_id_str,and AsyncOp have method task_id_str. Resource has a member id_str, and member id_str. So look like the name convention is not consistency.  So should open another pr to rename all names, such as should not use id, use seq_id or another name to name the console seq id, so that we can see the difference for the names  of seqid, taskid, spanid, resid etc

hds (approved):
Great, thank you for your PR!

PR #457, opened 2023-08-17
chore: improve aggregator event queue efficiency

### LINKED ISSUE #184: Continuous memory leak with `console_subscriber`
When calling `console_subscriber::init()` there is a big memory leak that seems to grow faster the more futures are spawned.
I ran this application through valgrind and it said that all the blocks were still reachable (the pointers still exist) so I'm guessing it's some sort of Circular ref-counted data structure that isn't using Weak properly (or something equivalent).

When I commented out the `console_subscriber::init()` line the leak completely went away (now idling at 3mb when it would go to 252mb when running for a couple hours with the subscriber)

h33p:
- Implement queue drain pressure that is based on channel's capacity. Pressure usually leads to the system reaching equilibrium without piling the events up.
- Allow configuring channel's target fill capacity, before flushing.
- Lower event buffer capacity, because the current value let the buffer grow into gigabytes in size.

# Details

This reverts https://github.com/tokio-rs/console/pull/235, because I found that limit being unreasonably high - adding the following code would lead to tens of gigabytes of size events being piled up to the event queue:

```rust

    let spam = tokio::spawn(async {
        loop {
            let _ = tokio::spawn(async {
                futures::future::join_all((0..100).map(|_| {
                    tokio::spawn(async {
                        let now = tokio::time::Instant::now();
                        while now.elapsed() < tokio::time::Duration::from_secs(1) {
                        }
                    })
                }))
                .await;
            })
            .await;
        }
    });
```

The problem I found in testing is that the aggregator task would consistently fall behind the number of events being created and eventually the event queue would fill up, regardless of its size (increasing the size merely delays the problem). The reason why it fills up is rather subtle `recv().now_or_never()` eventually returns `Poll::Pending` due to task coop budget being used up. My immediate fix was to wrap the future with `tokio::task::unconstrained`, but I figured there's a reason why it's not doing `try_recv()` and that is to avoid livelocking the task.

What this PR does is expose the "fullness" level of the queue to the shared vars. This allows the aggregator task to have a secondary timer, alongside regular ticking, that adds pressure to clear out the event queue at the same rate as it's filling up.

Another important thing to note is that event queue filling up faster than it's being processed will lead to time dilation - the rate of events in the console would appear slower than it actually is. Adding pressure eventually leads to an equilibrium where events are processed at roughly equal rate.

One key change was adding a customizable variable `event_buffer_flush_fraction` - the higher the fraction the sooner this equilibrium will be reached. I set it to `10` for now (before the effective value was `2`), which, with the reduction of event buffer capacity, should lead to 50x+ faster reaction by the aggregator task, while keeping 9x extra space for random spikes. Note that this variable may be changed to something slightly easier to understand, like `event_buffer_flush_percentage`, but that's just a user-facing detail I have no strong opinion about.

I believe this should fix https://github.com/tokio-rs/console/issues/184, but I haven't done an actual long-term test in production, will see in a couple days. EDIT: it doesn't seem to fix it


hawkw:
Thank you for working on this! The current queue sizes and other parameters were chosen pretty arbitrarily in the interest of "just getting something working", so I had been hoping to improve the queue performance for some time. I'm excited to give this PR a review!

console-subscriber/src/builder.rs
```diff
@@ -23,6 +23,10 @@ pub struct Builder {
     /// the aggregator task.
     pub(super) event_buffer_capacity: usize,
 
+    /// Fraction of the event buffer needed before the aggregator task starts
+    /// to flush it more aggresively.
```
hawkw: can we describe more of how this fraction is represented here? is it relative to the `event_buffer_capacity`, or is it, say, out of 100%?

hawkw: also, i think there's a typo 
```suggestion
    /// Fraction of the event buffer needed before the aggregator task starts
    /// to flush it more aggressively.
```

console-subscriber/src/builder.rs
```diff
@@ -89,6 +94,21 @@ impl Builder {
         }
     }
 
+    /// Sets the target fraction for the fill level of channel of events sent
+    /// from subscriber layers to the aggregator task.
+    ///
+    /// When the channel is over `1 / event_buffer_flush_fraction` fill
+    /// capacity , the aggregator task will be woken up more frequently.
```
hawkw: typo:
```suggestion
    /// capacity, the aggregator task will be woken up more frequently.
```

console-subscriber/src/builder.rs
```diff
@@ -89,6 +94,21 @@ impl Builder {
...
+    ///
+    /// By default, this is
+    /// [`ConsoleLayer::DEFAULT_EVENT_BUFFER_FLUSH_FRACTION`].
+    pub fn event_buffer_flush_fraction(self, event_buffer_flush_fraction: usize) -> Self {
```
hawkw: since this is a fraction, would it make more sense to represent it as a floating point number from 0.0 - 1.0?

console-subscriber/src/lib.rs
```diff
@@ -163,6 +163,12 @@ struct Shared {
     /// flushed.
     flush: aggregator::Flush,
 
+    /// As fill percentage gets higher, aggregator task gets polled more often,
+    /// to avoid filling up to the flush point. While `flush` notifiew the
```
hawkw: a couple minor docs nits:
```suggestion
    /// As fill percentage gets higher, the aggregator task gets woken more often,
    /// to avoid filling up to the flush point. While `flush` notifies the
```

console-subscriber/src/aggregator/mod.rs
```diff
@@ -171,6 +171,9 @@ impl Aggregator {
     pub async fn run(mut self) {
         let mut publish = tokio::time::interval(self.publish_interval);
         loop {
+            let fill_percentage = self.shared.fill_percentage.load(Relaxed) as u32;
+            let sleep = tokio::time::sleep(self.publish_interval * (100 - fill_percentage) / 100);
```
hawkw: i think it would be nice to have a comment describing this logic and what the second sleep is for.

hawkw: does this actually flush _more_ frequently than the default publish interval? it seems like if the fill percentage is 100, this will be the same as the publish interval, otherwise it will be greater. is that the intention here, or is this supposed to fire _more_ frequently than the publish interval when the queue is more full?

h33p: If fill percentage is 100, then the interval will be `publish_interval * 0`, meaning there will not be any real sleeping happening.

h33p:
@hawkw I have pushed requested changes (can squash it later, but I think it's easier to see changes during review)

hds:
@h33p Just letting you know that we haven't forgotten this PR!

I've got a test case that I'd like to run your changes against to compare to the current behaviour. It pretty easily produces a situation where the channel between the subscriber and the aggregator grows forever (at least on my local computer using a single thread).

PR #458, opened 2023-08-18
DO NOT MERGE test(subscriber): additional testing of integration tests on CI

hds:
This is a second playground for fixing #452.


hds:
No longer needed.

PR #460, opened 2023-08-24
Remove futures dependency from console-subscriber

asonix:
I noticed that `console-subscriber` was my only dependency pulling in the `futures` top-level library so I decided to try my hand at removing that dependency. This still relies on `futures-task` for an optimized `NoopWaker`


hawkw (commented):
seems fine to me

hawkw (approved):

PR #463, opened 2023-09-11
docs(subscriber): fix unclosed code block

crawford:


crawford:
The effect of the change can be seen here: https://deploy-preview-463--tokio-console.netlify.app/console_subscriber/struct.builder#configuration (that staging job is spiffy!)

hds (approved):
Thank you very much for the fix!

PR #465, opened 2023-09-19
chore(subscriber): trace events drain loop stats

hds:
The number of each variant of event encountered while draining the mpsc
channel of events sent from the `ConsoleLayer` to the `Aggregator` is
useful debugging information.

The change adds a `DEBUG` level event which is recorded at the end of
each loop draining the channel. It contains the number of each variant
that has been received.

I've written this same code three separate times looking into different
issues, so I decided it was probably worth the small overhead of
aggregating the counts, even when the tracing event isn't output
anywhere.

### DIFF
--- a/console-subscriber/src/aggregator/mod.rs
+++ b/console-subscriber/src/aggregator/mod.rs
@@ -237,11 +237,11 @@ impl Aggregator {
                 };
             }
             tracing::debug!(
-                async_resource_op = counts.async_resource_op,
-                metadata = counts.metadata,
-                poll_op = counts.poll_op,
-                resource = counts.resource,
-                spawn = counts.spawn,
+                async_resource_ops = counts.async_resource_op,
+                metadatas = counts.metadata,
+                poll_ops = counts.poll_op,
+                resources = counts.resource,
+                spawns = counts.spawn,
                 total = counts.total(),
                 "event channel drain loop",
             );
--- a/console-subscriber/src/aggregator/mod.rs
+++ b/console-subscriber/src/aggregator/mod.rs
@@ -220,9 +220,11 @@ impl Aggregator {
             // to be woken when the flush interval has elapsed, or when the
             // channel is almost full.
             let mut drained = false;
+            let mut counts = EventCounts::new();
             while let Some(event) = recv_now_or_never(&mut self.events) {
                 match event {
                     Some(event) => {
+                        counts.update(&event);
                         self.update_state(event);
                         drained = true;
                     }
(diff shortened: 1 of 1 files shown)


console-subscriber/src/aggregator/mod.rs
```diff
@@ -220,9 +220,11 @@ impl Aggregator {
             // to be woken when the flush interval has elapsed, or when the
             // channel is almost full.
             let mut drained = false;
+            let mut counts = EventCounts::new();
             while let Some(event) = recv_now_or_never(&mut self.events) {
                 match event {
                     Some(event) => {
+                        counts.update(&event);
```
hawkw, line 227: take it or leave it: if we're concerned about the performance cost of aggregating these counts if we're not logging the `DEBUG` event, we could write something like this:
```suggestion
                        if tracing::event_enabled!(tracing::Level::DEBUG) {
                            counts.update(&event);
                        }
```

console-subscriber/src/aggregator/mod.rs
```diff
@@ -234,6 +236,15 @@ impl Aggregator {
                     }
                 };
             }
+            tracing::debug!(
+                async_resource_op = counts.async_resource_op,
+                metadata = counts.metadata,
+                poll_op = counts.poll_op,
+                resource = counts.resource,
+                spawn = counts.spawn,
+                total = counts.total(),
```
hawkw: tiny style nit: since these are counts, i would either pluralize the field names, like this:
```suggestion
                async_resource_ops = counts.async_resource_op,
                metadatas = counts.metadata,
                poll_ops = counts.poll_op,
                resources = counts.resource,
                spawns = counts.spawn,
                total = counts.total(),
```
or just leave the `counts.` prefix, like this:
```suggestion
                counts.async_resource_op,
                counts.metadata,
                counts.poll_op,
                counts.resource,
                counts.spawn,
                total = counts.total(),
```

up to you! :)

console-subscriber/src/aggregator/mod.rs
hawkw: but, i'm not sure if this is actually more efficient than just doing a single add per event. so i don't know if i would bother...

hawkw (approved):

console-subscriber/src/aggregator/mod.rs
hds: Given the updates are probably going to be very fast, I think I'll just sum them up anyway.

console-subscriber/src/aggregator/mod.rs
hds: Taken the first suggestion.

PR #466, opened 2023-09-21
chore: update release scripts

hds:
Improve the commit message added to release commits. It now includes a
complete change log for the release being made.

When the `console-api` version changes, it also needs to be changed in
the crates that depend on it. The release script now does this via cargo
update-version (from cargo-edit).

Committing the `Cargo.lock` file after publishing has been moved to
before adding the version tag to ensure that the correct commit is
tagged. The change log is also updated in the amended commit so that the
hash is correct.

The cliff configuration file has also been updated to fix some
incorrectly formatted commits.


bin/release.sh
```diff
@@ -71,6 +73,17 @@ update_version() {
...
+
+    if ! cargo --list | grep -q "set-version"; then
+        err "missing cargo-set-version executable (from cargo-edit)"
+        if confirm "       install it?"; then
+            cargo install cargo-edit
+        else
+            echo "okay, exiting"
+            exit 1
+        fi
+    fi
```
hawkw, line 84: what's the motivation for introducing an additional dep for doing this when we were previously able to do it with a single line of `sed`? i'm open to the change but would like to understand what this gets us that the sed script doesn't.

bin/release.sh
```diff
@@ -79,9 +92,51 @@ update_version() {
...
+        status "Amending" "release commit"
+        amend="$1"
+    else
+        status "Creating" "release commit"
+        amend=""
+    fi
+
+    # Prepare a commit message including the changelog from just this release.
+    tmp_changelog_path="tmp-changelog"
+    tmp_commit_msg_path="tmp-commit-msg"
```
hawkw: hmm, this is a bit unfortunate, since it creates the temporary files in the current working dir, and if the script exits early due to a command failing, the temporary files will be left behind.

we should maybe consider using `mktmp` to create a temporary dir files in the system's temporary directory (e.g. `/tmp`), instead. that way, even if we're not able to properly clean up our tempfiles, the system will eventually delete them, and we won't have issues with a pre-existing temp file with the same name.

see [here](https://unix.stackexchange.com/a/84980) for details on how to use `mktemp` in a cross platform (linux & BSDs like MacOS) way.

bin/release.sh
```diff
@@ -111,7 +174,7 @@ publish() {
         echo "# " "${git_push_tags[@]}"
     else
         "${git_tag[@]}"
-        "${git_push_tags[@]}"
+        #"${git_push_tags[@]}"
```
hawkw: why is this commented out? should we delete it? or, do we need to put it back so that the tag is actually pushed?

cliff.toml
```diff
@@ -68,6 +68,17 @@ filter_unconventional = true
...
 # regex for preprocessing the commit messages
 commit_preprocessors = [
+  # Fixes for specific commit messages that we wish we'd committed differently
+  { pattern = """^bump tonic\\/prost dependencies and MSRV to 1.64 \\(#464\\)\\s+
+(.+\\s+.+)""", replace = """feat(api): update `tonic` to v0.10 and increase MSRV to 1.64 (#464)
+
+BREAKING CHANGE:
+This is a breaking change for users of `console-api` and `console-subscriber`,
+as it changes the public `tonic` dependency to a semver-incompatible version.
+This breaks compatibility with `tonic` 0.9. and `prost` 0.11.""" },
```
hawkw: sigh...this is a good call, but it's unfortunate that we have to do this. maybe at some point we should go back and edit the git history so that that commit isn't screwy...this looks good for now though.

bin/release.sh
```diff
@@ -7,7 +7,9 @@ USAGE:
 FLAGS:
     -h, --help      Show this help text and exit.
     -v, --verbose   Enable verbose output.
-    -d, --dry-run   Do not change any files or commit a git tag."
+    -d, --dry-run   Do not change any files or commit a git tag.
+    --publish-dry-run
+                    Perform a dry run on the publish step only."
```
hawkw: very tiny style nit: i would prefer to format this so that it's on the same line, and just go back and indent the other flags slightly more:
```suggestion
    --publish-dry-run    Perform a dry run on the publish step only."
```

hawkw (commented):
I had some minor nits/questions about the bash script changes. Thanks so much for working on this!

cliff.toml
hawkw: okay, i went and rewrote history, so the commit now looks more like how we'd like it to: https://github.com/tokio-rs/console/commit/882318a006d060c763f97afa7e03a45ef9736fe6

we can remove this now
```suggestion
```

bin/release.sh
hds: I used `cargo ugrade` from `cargo-edit` to set the version of `console-api` on `console-subscriber` and `tokio-console`. I could probably have done it with a `sed` script too, but it's a bit harder (and more difficult to maintain).

Since I was installing `cargo-edit` for that anyway, it made sense to use it here too (again, I feel it's easier to maintain).

Using `sed` also has the downside of not always being compatible of different unixes (*cough* BSD/macOS *cough*).

bin/release.sh
hds: Done.

cliff.toml
hds: Removed now that you updated the commit.

bin/release.sh
hds: Updated: I moved the push to a separate step after publish, with it's own confirmation.

bin/release.sh
hds: Done.

bin/release.sh
hawkw: cool, makes sense to me! 👍

bin/release.sh
```diff
@@ -79,9 +91,49 @@ update_version() {
...
+    if ! [[ -z "${1+1}" ]] && [[ "$1" == "--amend" ]]; then
+        status "Amending" "release commit"
+        amend="$1"
+    else
+        status "Creating" "release commit"
+        amend=""
+    fi
+
+    # Prepare a commit message including the changelog from just this release.
+    tmp_dir=$(mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
```
hawkw: let's use a name for the tempdir that actually describes what it's for?

```suggestion
    tmp_dir=$(mktemp -d 2>/dev/null || mktemp -d -t 'tokio-console-release')
```

hawkw (approved):

bin/release.sh
hds: Oops. I might have copied this straight from SO. 😅

cliff.toml
hds: I thought I'd removed this...

hds: This isn't in the latest version, so this is done. I think I did something wrong when pushing it.

PR #468, opened 2023-10-01
fix(console): ignore key release events

fenhl:
On supported platforms, a short key press generates two crossterm events: One for the initial press and another when the key is released again. The code was not checking the event kind, so it would treat a key release as a second key press. Filter out key releases entirely since they're not explicitly used for anything.

### DIFF
--- a/tokio-console/src/main.rs
+++ b/tokio-console/src/main.rs
@@ -3,7 +3,7 @@ use console_api::tasks::TaskDetails;
 use state::State;
 
 use futures::{
-    future::ready,
+    future,
     stream::{StreamExt, TryStreamExt},
 };
 use ratatui::{
@@ -14,7 +14,10 @@ use ratatui::{
 };
 use tokio::sync::{mpsc, watch};
 
-use crate::view::{bold, UpdateKind};
+use crate::{
+    input::{Event, KeyEvent, KeyEventKind},
+    view::{bold, UpdateKind},
+};
 
 mod config;
 mod conn;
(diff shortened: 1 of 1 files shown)


tokio-console/src/main.rs
```diff
@@ -68,7 +71,15 @@ async fn main() -> color_eyre::Result<()> {
...
         ])
         .with_retain_for(retain_for);
-    let mut input = input::EventStream::new();
+    let mut input = Box::pin(input::EventStream::new().try_filter(|event| {
+        ready(!matches!(
+            event,
+            input::Event::Key(input::KeyEvent {
+                kind: input::KeyEventKind::Release,
+                ..
+            })
```
hawkw: style nit, not important: it's up to you, but i might import `Event`, `KeyEvent`, and `KeyEventKind` so we can drop the `input::` prefix here and make the match pattern here a bit less wordy.

tokio-console/src/main.rs
```diff
@@ -68,7 +71,15 @@ async fn main() -> color_eyre::Result<()> {
             warnings::Linter::new(warnings::NeverYielded::default()),
         ])
         .with_retain_for(retain_for);
-    let mut input = input::EventStream::new();
+    let mut input = Box::pin(input::EventStream::new().try_filter(|event| {
+        ready(!matches!(
```
hawkw: style nit, take it or leave it: my preference would have been to call this as
```suggestion
        future::ready(!matches!(
```
since there are a few different things named `ready` (eg the `ready!` macro) and it makes it a little clearer that we're constructing a future. not a big deal though!

hawkw (approved):
thanks for the fix, this looks good! i commented on some unimportant style nits, but feel free to ignore them. :)

PR #469, opened 2023-10-01
chore(console): update Clap to v4.1

0xPoe:
Based on https://github.com/tokio-rs/console/pull/384

close https://github.com/tokio-rs/console/pull/384

---

Commit message as updated by @hds:

Tokio console was previously using Clap v3, which is in maintenance and                             
doesn't always receive necessary updates. For example, the fix for a new                            
lint in Clippy was submitted by us (and updated in #414).                                           
                                                                                                    
This change updates to Clap 4.1, as the 4.x versions are generally                                  
better supported. This is the latest version of Clap which is compatible                            
with our current MSRV of 1.64. From the 4.2 versions there are some                                 
depdendencies which require Rust 1.70.                                                              
                                                                                                    
Additionally, we switched from using a `tracing_subscriber::EnvFilter`                              
to a `tracing_subscriber::filter::Targets` for filtering the console's                              
internal log messages. This is because `clap` requires that all arg                                 
values be `Clone` in v4, and we weren't using `EnvFilter` span filtering                            
anyway, so we felt like this was the simplest option.                                               
                                                                                                    
This change shouldn't result in any change in behavior besides                                      
differences in CLI help text formatting.

This change is based on the work originally done in #384.

### DIFF
--- a/tokio-console/Cargo.toml
+++ b/tokio-console/Cargo.toml
@@ -30,8 +30,8 @@ keywords = [
 atty = "0.2"
 console-api = { version = "0.6.0", path = "../console-api", features = ["transport"] }
 # Use 1.64.0 compatible version of `clap@4.1.14`.
-clap = { version = "=4.1.14", features = ["wrap_help", "cargo", "derive", "env"] }
-clap_complete = "=4.1.6"
+clap = { version = "~4.1.14", features = ["wrap_help", "cargo", "derive", "env"] }
+clap_complete = "~4.1.6"
 tokio = { version = "1", features = ["full", "rt-multi-thread"] }
 tonic = { version = "0.10", features = ["transport"] }
 futures = "0.3"
--- a/tokio-console/Cargo.toml
+++ b/tokio-console/Cargo.toml
@@ -29,7 +29,8 @@ keywords = [
 [dependencies]
 atty = "0.2"
 console-api = { version = "0.6.0", path = "../console-api", features = ["transport"] }
-clap = { version = "4.1.14", features = [
+# Use 1.64.0 compatible version of `clap@4.1.14`.
+clap = { version = "=4.1.14", features = [
     # these are all enabled by default
     "std", "color", "help", "usage", "error-context", "suggestions",
     # auto-wrap help text to fit the terminal width
@@ -41,7 +42,7 @@ clap = { version = "4.1.14", features = [
     # parse values from env vars
     "env"
 ] }
-clap_complete = "4.1.6"
+clap_complete = "=4.1.6"
 tokio = { version = "1", features = ["full", "rt-multi-thread"] }
 tonic = { version = "0.10", features = ["transport"] }
 futures = "0.3"
--- a/tokio-console/Cargo.toml
+++ b/tokio-console/Cargo.toml
@@ -29,7 +29,7 @@ keywords = [
 [dependencies]
 atty = "0.2"
 console-api = { version = "0.6.0", path = "../console-api", features = ["transport"] }
-clap = { version = "4", features = [
+clap = { version = "4.1.14", features = [
     # these are all enabled by default
     "std", "color", "help", "usage", "error-context", "suggestions",
     # auto-wrap help text to fit the terminal width
@@ -41,7 +41,7 @@ clap = { version = "4", features = [
     # parse values from env vars
     "env"
 ] }
-clap_complete = "4"
+clap_complete = "4.1.6"
 tokio = { version = "1", features = ["full", "rt-multi-thread"] }
 tonic = { version = "0.10", features = ["transport"] }
 futures = "0.3"
--- a/tokio-console/Cargo.toml
+++ b/tokio-console/Cargo.toml
@@ -30,6 +30,7 @@ keywords = [
 atty = "0.2"
 console-api = { version = "0.6.0", path = "../console-api", features = ["transport"] }
 clap = { version = "3", features = ["cargo", "derive", "env"] }
+clap_complete = "3"
 tokio = { version = "1", features = ["full", "rt-multi-thread"] }
 tonic = { version = "0.10", features = ["transport"] }
 futures = "0.3"
@@ -49,4 +50,3 @@ humantime = "2.1.0"
 serde = { version = "1", features = ["derive"] }
 toml = "0.5"
 dirs = "4"
-clap_complete = "3"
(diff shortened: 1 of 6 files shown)


tokio-console/src/config.rs
```diff
@@ -153,7 +155,7 @@ impl Serialize for RetainFor {
 #[clap(group = ArgGroup::new("colors").conflicts_with("no-colors"))]
 pub struct ViewOptions {
     /// Disable ANSI colors entirely.
-    #[clap(name = "no-colors", long = "no-colors", takes_value = false)]
+    #[clap(name = "no-colors", long = "no-colors",action = ArgAction::SetTrue)]
```
hds: nit: It's only sort of related to your change, but in the `help` output, `--no-colors` no longer appears next to the other color options.

It would be great if you could move this field down below `ascii_only` so that they're together again.

hds (changes requested):
Just a small nit about the ordering of the options in the help text after the change.

tokio-console/src/config.rs
0xPoe: Updated!

tokio-console/Cargo.toml
```diff
@@ -29,14 +29,27 @@ keywords = [
...
+    "std", "color", "help", "usage", "error-context", "suggestions",
+    # auto-wrap help text to fit the terminal width
+    "wrap_help",
+    # use Cargo env vars when determining help text
+    "cargo",
+    # enable derive attributes
+    "derive",
+    # parse values from env vars
+    "env"
+] }
```
hds: A couple of things here.

1. Let's keep a single line per dependency the same as the other dependencies and remove the default features as we don't need to specify them
2. From the tags in the clap repo, clap 4.3.24 also has an MSRV of 1.64, so we can use that one. Also, we don't need to fix an exact version, but instead use `~4.3` which means `>= 4.3.0` and `< 4.4.0`.

Mostly this is about trusting the clap maintainers to uphold semver. Default features shouldn't be removed except in a major version change and it's the norm to not update the MSRV in a patch update.

```suggestion
clap = { version = "~4.3", features = ["wrap_help", "cargo", "derive", "env"] }
```

tokio-console/Cargo.toml
```diff
@@ -29,14 +29,27 @@ keywords = [
...
+clap_complete = "=4.1.6"
```
hds: This should also be 4.3 to match `clap` I believe.

```suggestion
clap_complete = "~4.3"
```

tokio-console/Cargo.toml
```diff
@@ -29,14 +29,27 @@ keywords = [
...
 tokio = { version = "1", features = ["full", "rt-multi-thread"] }
 tonic = { version = "0.10", features = ["transport"] }
 futures = "0.3"
 ratatui = { version = "0.20.1", default-features = false, features = ["crossterm"] }
 tower = "0.4.12"
 tracing = "0.1"
-tracing-subscriber = { version = "0.3.0", features = ["env-filter"] }
+tracing-subscriber = { version = "0.3.0" }
```
hds: nit: we can collapse this down to the short version now that we're not pulling in non-default features:

```suggestion
tracing-subscriber = "0.3"
```

xtask/Cargo.toml
```diff
@@ -10,5 +10,5 @@ publish = false
 tonic-build = { version = "0.10", default-features = false, features = [
     "prost", "transport"
 ] }
-clap = { version = "3", features = ["derive"] }
+clap = { version = "4", features = ["derive"] }
```
hds: Since this crate also declares MSRV 1.64, let's use the same version specifier as for `tokio-console`.

```suggestion
clap = { version = "~4.3", features = ["derive"] }
```

hds (changes requested):
Sorry, I missed a couple of things in the `Cargo.toml` files on my first read through.

tokio-console/Cargo.toml
0xPoe: Because Cargo doesn't consider the MSRV when resolving the deps. So if we try to use `~4.3`. Then we need to update some other deps to some specific version.

For example:
```console
console on  rustin-patch-clap [!] via 🦀 v1.64.0 
❯ cargo b
    Updating crates.io index
error: package `anstyle-parse v0.2.2` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.64.0

console on  rustin-patch-clap [!] via 🦀 v1.64.0 
❯ cargo update -p anstyle-parse --precise 0.2.1       
    Updating crates.io index
    Updating anstyle-parse v0.2.2 -> v0.2.1

console on  rustin-patch-clap [!] via 🦀 v1.64.0 took 6s 
❯ cargo b                                      
error: package `anstyle v1.0.4` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.64.0

console on  rustin-patch-clap [!] via 🦀 v1.64.0 
❯ cargo update -p anstyle --precise 1.0.1 
    Updating crates.io index
    Updating anstyle v1.0.4 -> v1.0.1

console on  rustin-patch-clap [!] via 🦀 v1.64.0 
❯ cargo b                                
error: package `clap_lex v0.5.1` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.64.0

console on  rustin-patch-clap [!] via 🦀 v1.64.0 

```

0xPoe: Removed all comments.

tokio-console/Cargo.toml
0xPoe: Fixed.

tokio-console/Cargo.toml
hds: That's anoying, even `4.1.14` doesn't work?

0xPoe: `4.1.14` works well. Because it doesn't have those deps.

hds: OK, then let's use `"^4.1.14"` for the version then.

0xPoe: `^4.1.14` same as `4.1.14`: 4.1.14  :=  >=4.1.14, <5.0.0.

So it wouldn't help us to use a compliable version.

hds: Sorry! I meant `"~4.1.14"`.

0xPoe: Sounds good. Updated!

xtask/Cargo.toml
0xPoe: Fixed.

hds (approved):
Thanks for your work!

0xPoe:

Issue #473, opened 2023-10-11
Flaky `console-subscriber` integration tests


hds:
### What crate(s) in this repo are involved in the problem?

console-subscriber

### What is the issue?

The `console-subscriber` integration tests show flaky behavior locally.

The tests were originally introduced in #452.

### How can the bug be reproduced?

Run all the `framework` tests locally. Sometimes the `self_wakes` test fails.

To reproduce this, I ran the tests in a loop until they failed. On my machine it typically takes under 5 minutes to produce a failure.

```sh
echo "Start!" && time while [ $? -eq 0 ]; do cargo test -p console-subscriber --test framework; done
```

### Versions

* `main` branch.
* `console-subscriber` v0.2.0

### Possible solution

It appears that the root cause for this issue is the one described in tokio-rs/tracing#2743.

### Would you like to work on fixing this bug?

yes

0xPoe:
Here is another one: `double_sleep_wakes `
https://github.com/tokio-rs/console/actions/runs/6989256116/job/19017617547?pr=494

hds:
These two tests are so flaky that I'm going to delete them until we fix the problem. From `console-subscriber/tests/wake.rs`:

```rust
#[test]
fn sleep_wakes() {
    let expected_task = ExpectedTask::default()
        .match_default_name()
        .expect_wakes(1)
        .expect_self_wakes(0);

    let future = async {
        sleep(Duration::ZERO).await;
    };

    assert_task(expected_task, future);
}

#[test]
fn double_sleep_wakes() {
    let expected_task = ExpectedTask::default()
        .match_default_name()
        .expect_wakes(2)
        .expect_self_wakes(0);

    let future = async {
        sleep(Duration::ZERO).await;
        sleep(Duration::ZERO).await;
    };

    assert_task(expected_task, future);
}
```

PR #474, opened 2023-10-11
fix(subscriber): mitigate race in `Callsites::contains`

hds:
The `ConsoleLayer` uses the `Callsites` struct to store and check for
callsites for specific kinds of traces, for example spawn spans or waker
events.

`Callsites` stores a fixed size array of pointers to the `Metadata` for
each callsite and a length to indicate how many callsites it has
registered. The length and each individual pointer are stored in
atomics.

Since it is possible for these values to change individually, if a
callsite lookup fails, we check if the length of the array has changed
while we were checking the pointers, if it has, the lookup is started
again.

However, there is still a possible race condition. If the length
changes, but the lookup occurs before the callsite pointer is actually
written, then we may miss a callsite that is in the process of being
registered. In this case, the pointer which is loaded from the
`Callsites` array will be null.

This change adds a check for this case (null ptr), and reperforms the
lookup if it occurs.

This race condition was found while chasing down the source of #473. It
doesn't solve the flakiness, but it can reduce the likelihood of it
occuring, thus it is a mitigation only.

In reality, neither of these race condition checks should be needed, as
we would expect that `tracing` guarantees that `ConsoleLayer` completes
`register_callsite()` before `on_event()` or `new_span()` are called.

### DIFF
--- a/console-subscriber/src/callsites.rs
+++ b/console-subscriber/src/callsites.rs
@@ -46,8 +46,13 @@ impl<const MAX_CALLSITES: usize> Callsites<MAX_CALLSITES> {
         let mut len = self.len.load(Ordering::Acquire);
         loop {
             for cs in &self.ptrs[start..len] {
-                if ptr::eq(cs.load(Ordering::Acquire), callsite) {
+                let recorded = cs.load(Ordering::Acquire);
+                if ptr::eq(recorded, callsite) {
                     return true;
+                } else if ptr::eq(recorded, ptr::null_mut()) {
+                    // We have read a recorded callsite before it has been
+                    // written. We need to check again.
+                    continue;
                 }
             }
 


console-subscriber/src/callsites.rs
```diff
@@ -46,8 +46,13 @@ impl<const MAX_CALLSITES: usize> Callsites<MAX_CALLSITES> {
         let mut len = self.len.load(Ordering::Acquire);
         loop {
             for cs in &self.ptrs[start..len] {
-                if ptr::eq(cs.load(Ordering::Acquire), callsite) {
+                let recorded = cs.load(Ordering::Acquire);
+                if ptr::eq(recorded, callsite) {
                     return true;
+                } else if ptr::eq(recorded, ptr::null_mut()) {
```
hawkw, line 52: style nit, take it or leave it: this could be 
```suggestion
                } else if recorded.is_null() {
```

console-subscriber/src/callsites.rs
```diff
@@ -46,8 +46,13 @@ impl<const MAX_CALLSITES: usize> Callsites<MAX_CALLSITES> {
...
+                    // We have read a recorded callsite before it has been
+                    // written. We need to check again.
+                    continue;
```
hawkw, line 55: hmm, this restarts the whole loop over again. should we, instead, have an inner loop for loading the specific array index until it's no longer null?

hawkw:
it would be really nice to have a test that reproduces this raciness. maybe we should add [`loom` tests](https://github.com/tokio-rs/loom) for `Callsites` to help catch this kind of issue...

console-subscriber/src/callsites.rs
hds: @hawkw After some thinking (and then forgetting about this), I beleive that this change doesn't make sense.

In fact, I'm not sure that the retry mechanism makes sense in general. If `tracing` gives the guarantee that `Subscriber::register_callsite` will be called (**and finish!**) before `Subscriber::event` get's called, then the retry shouldn't be necessary at all.

If it doesn't give that guarantee, then the retry is increasing our chances of finding the callsite we are interested in during a race condition, but I don't think that it actually solves the problem either.

For now I'm going to close this PR. After looking at tokio-rs/tracing#2743 a bit more I'll revisit.

hds:
Closing this without merging as I don't think it actually makes sense. See https://github.com/tokio-rs/console/pull/474#discussion_r1395588616 for more details.

PR #475, opened 2023-10-11
test(subscriber): prefer `sleep` over `yield_now` in tests

hds:
A flakiness problem has been discovered with the `console-subscriber`
integration tests introduced in #452. Issue #473 is tracking the issue.

It has been observed that we only "miss" the wake operation event when
it comes from `yield_now()`, but not when it comes from a task that
yielded due to `sleep`, even when the duration is zero. it is likely
that this is due to nature of the underlying race condition.

This change removes all the calls to `yield_now()` from the `framework`
tests, except those where we wish to actually test self wakes.

### DIFF
--- a/console-subscriber/tests/framework.rs
+++ b/console-subscriber/tests/framework.rs
@@ -100,7 +100,7 @@ fn test_spawned_task() {
     let future = async {
         task::Builder::new()
             .name("another-name")
-            .spawn(async { task::yield_now().await })
+            .spawn(async { sleep(Duration::ZERO).await })
     };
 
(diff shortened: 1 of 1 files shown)


hawkw (approved):
overall, this looks good to me! 

i think it might be good if there were comments explaining the use of zero-duration sleeps rather than `yield_now`s, so that nobody comes along later and thinks "these seem like they could be replaced with `yield_now`", without the original context for why we use `sleep` here.

if you don't mind adding comments, let's get this merged!

hds:
Comments added!

PR #476, opened 2023-10-12
test(subscriber): add tests for excessive poll counts

hds:
A defect in the way polls are counted was fixed in #440. The defect
caused the polls of a "child" task to also be counted on the "parent"
task. Where in this case, the "parent" is a task that spawns the "child"
task.

This change adds a regression test for that fix.

To do so, functionality to record and validate poll counts is added to
the `console-subscriber` integration test framework. The new
functionality is also used to add some basic tests for poll counts.

### DIFF
--- a/console-subscriber/tests/support/task.rs
+++ b/console-subscriber/tests/support/task.rs
@@ -259,7 +259,7 @@ impl ExpectedTask {
         self
     }
 
-    /// Expects taht a task has a specific value for `polls`.
+    /// Expects that a task has a specific value for `polls`.
     ///
     /// To validate, the actual task must have a count of polls (on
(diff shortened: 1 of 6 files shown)


hawkw (approved):

console-subscriber/tests/support/task.rs
```diff
@@ -229,6 +258,16 @@ impl ExpectedTask {
         self.expect_self_wakes = Some(self_wakes);
         self
     }
+
+    /// Expects taht a task has a specific value for `polls`.
```
hawkw:
```suggestion
    /// Expects that a task has a specific value for `polls`.
```

hawkw (approved):

PR #470, opened 2023-10-08
docs: update outdated main README.md

0xPoe:
follow up https://github.com/tokio-rs/console/commit/caba1a36ae2b4c4c489f5fde3499d21f3ae8fbdb

PR #477, opened 2023-10-12
test(console): use try_cmd to generate the help text

0xPoe:
I tried to use `try_cmd` to avoid [this kind of problem](https://github.com/tokio-rs/console/pull/470). 

1. It is easier to use `TRYCMD=overwrite` to update all outputs.
2. We also test the output from the README file.

### DIFF
--- a/tokio-console/src/lib.rs
+++ b/tokio-console/src/lib.rs
@@ -12,7 +12,7 @@
 /// `tokio-console`:
 ///
 /// ```text
-#[doc = include_str!("../args.example")]
+#[doc = include_str!("../cli-ui.stdout")]
 /// ```
 ///
 /// This text can also be displayed by running `tokio-console help`.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,54 +27,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anstream"
-version = "0.6.4"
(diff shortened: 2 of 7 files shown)


tokio-console/src/lib.rs
```diff
@@ -12,7 +12,7 @@
 /// `tokio-console`:
 ///
 /// ```text
-#[doc = include_str!("../args.example")]
```
hds, line 15: Could we also delete `args.example` please.

hds (changes requested):
Looks good and much simpler to deal with. Thank you!

Although could you please remove `args.example` as I think it's not longer used.

tokio-console/src/lib.rs
0xPoe: I think I already did:
image

0xPoe: Thanks for your review!  💚 💙 💜 💛 ❤️

hds: You're right! Sorry, I missed that it was renamed.

This is probably fine to go then. I'm going to give it a once over from my computer and then it can probably be merged.

Thanks for your contributions!

hawkw (approved):
neat!

hds (approved):
Looks good. Thanks!

PR #478, opened 2023-10-16
feat(console): replace target column with kind column in tasks view

hds:
In the `tokio-console` tasks view, there are a fixed set of columns and
any remaining fields are included in a "Fields" column at the end. One
of the fields which is always present on a task, but doesn't receive a
dedicated column is the kind field, which currently takes one of the
following values:
* `task` (a "normal" async task)
* `blocking`
* `block_on`
* `local`

Meanwhile, there is a dedicated column for the task span's target, which
currently takes one of the following values:
* `tokio::task`
* `tokio::task::blocking`

The target for tasks with kind `block_on` and `local` is also
`tokio::task`.

This change replaces the target column with a kind column as it provides
more information in fewer characters. The target value is moved
(somewhat artificially) to the fields which appear in the final column.

The `target` is also left on the `state::Task` struct as we expect to
want to filter by it in the future.

Additionally, the `console-subscriber` examples have been updated so
that there are options to visualize `blocking`, `block_on`, and `local`
tasks. The `app` example has been updated to include an optional task
which calls `tokio::spawn_blocking`. A new example `local` has been
added which creates a `LocalSet` and spawns local tasks onto it.

### DIFF
--- a/console-subscriber/examples/local.rs
+++ b/console-subscriber/examples/local.rs
@@ -1,0 +1,28 @@
+use std::time::Duration;
+use tokio::{runtime::Runtime, task};
+
+fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    console_subscriber::init();
+
+    let rt = Runtime::new().unwrap();
+    let local = task::LocalSet::new();
+    local.block_on(&rt, async {
+        loop {
+            let mut join_handles = Vec::new();
+            for _ in 0..10 {
+                let jh = task::spawn_local(async {
+                    tokio::time::sleep(Duration::from_millis(100)).await;
+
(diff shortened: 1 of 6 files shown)


console-subscriber/examples/local.rs
```diff
@@ -0,0 +1,28 @@
+use std::time::Duration;
```
hawkw, line 8: can we include a quick doc comment or something in this example explaining what it demonstrates? not a big deal.

hawkw (approved):

console-subscriber/examples/local.rs
hds: Good point. Done.

PR #483, opened 2023-10-26
chore(console): fix format issue in select!

0xPoe:
I think we need a space here.


hds (approved):
Yep, that's correct. Thanks!

PR #484, opened 2023-10-30
refactor(console): bail out the error earlier

0xPoe:
We don't need to create the endpoint first.


hawkw (approved):
thanks, looks good to me!

PR #486, opened 2023-11-02
fix(console): only accept `file`, `http://`, `https://` URI

### LINKED ISSUE #401: When connecting to url without `http://` weird things are happening

tokio-console

### What is the issue?

First time using `tokio-console`, I did `tokio-console 127.0.0.1:1234` and it kind of worked, not really. The display is blank, but when changing its size it displays something update reconnecting and `t = tasks, r = resources`, but doesn't seem to update anything.

Eventually I figured out that I'm supposed to do `tokio-console http://127.0.0.1:1234` and that gives me the UI.

`tokio-console 127.0.0.1:1234` (no `http://`)

### Logs, error output, etc

_No response_

### Versions

```
> tokio-console -V
tokio-console 0.1.7
```

### Possible solution

_No response_

### Additional context

_No response_

no

0xPoe:
close https://github.com/tokio-rs/console/issues/401

I am not sure if we support `https://`.

### DIFF
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -363,11 +363,22 @@ impl Config {
         self.retain_for.unwrap_or_default().0
     }
 
-    pub(crate) fn target_addr(&self) -> Uri {
-        self.target_addr
+    pub(crate) fn target_addr(&self) -> color_eyre::Result<Uri> {
+        let target_addr = self
+            .target_addr
             .as_ref()
             .unwrap_or(&default_target_addr())
-            .clone()
+            .clone();
+        match target_addr.scheme_str() {
+            Some("file" | "http" | "https") => {}
+            _ => {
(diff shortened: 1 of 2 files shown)


0xPoe:
Tested locally:
```console
console on  main [!⇡] via 💎 v3.0.0 via 🦀 v1.64.0 
❯ cargo run -- aadada 
    Finished dev [unoptimized + debuginfo] target(s) in 0.32s
     Running `target/debug/tokio-console aadada`
Error: 
   0: invalid scheme for target address aadada, must be one of 'file', 'http', or 'https'

Location:
   tokio-console/src/config.rs:375

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

hds (approved):
Looks good. Thanks!

PR #489, opened 2023-11-11
Run the upload-bins github actions step on release

gyfis:
Skip creating the release, since `taiki-e/create-gh-release-action` is not compatible with the Changelog format

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,34 +1,14 @@
 name: Upload Binaries
 
 on:
-  push:
-    tags:
-      - console-subscriber-v[0-9]+.*
-      - console-api-v[0-9]+.*
-      - tokio-console-v[0-9]+.*
+  release:
+    types: [created]
 
 jobs:
-  create-release:
-    name: Create GitHub release
-    # only publish from the origin repository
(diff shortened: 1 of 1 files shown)


hds:
@gyfis Thanks for your PR! I need to have a bit of a look at it, as ideally we'd replace this step with creating a release in a different way and I'd prefer to have this still in the code until then (it's a sort of marker of intent, even if it doesn't work).

gyfis:
Hi @hds, thanks! Looking at the release page it seems that releases are being done fine https://github.com/tokio-rs/console/releases so I'd hope this change would hook right into the new approach (introduced here https://github.com/tokio-rs/console/pull/466).

hds:
@gyfis The new approach only pushes a commit and a tag, it doesn't create the GitHub release. If you check closely, you'll see that the latest releases in GitHub aren't the latest ones on crates.io (we're missing the latest one release for all 3 crates).

gyfis:
I see now, thanks! Gotcha, it'd be ideal to unify 😅 that's out of scope for me I'm afraid, feel free to close this PR (or I can do that)

gyfis:
Closing since there's no need for this change in the PR

Issue #488, opened 2023-11-09
Sorting does not work on all columns in the resources view


AEnterprise:
### What crate(s) in this repo are involved in the problem?

tokio-console

### What is the issue?

When using the resources view the sorting does not work on the following columns:
- parent
- target
- type
- vis
- location
- attributs

### How can the bug be reproduced?

Just using the tokio console to display metrics from a application with some different types of locks/channels

### Versions

tokio-console 0.1.10

### Would you like to work on fixing this bug?

maybe

hawkw:
Sounds like a bug, thanks for the report!

I'm assuming the other columns in the resources view do work for sorting?

AEnterprise:
Yes the other columns work fine

PR #491, opened 2023-11-22
fix(console): fix column sorting in resources tab (#488)

mbiggio:
In the Resources tab, columns names were being associates to the wrong sorting keys, and some column names were just missing from the `SortBy` implementation.

All columns have been associated to the corresponding sorting keys, and for the `Location` column, a structured type has been stored in the resource object, in order to sort lexicographically on the path and with conventional numeric ordering on row and column. Also, sorting for attributes as been implemented by sorting lexicographically on the first attribute of each resource row, even if that is probably sub-optimal


hds (changes requested):
Hi! 👋 Thank your PR!

I really like the change to fix the location sorting, that's a really great idea. However, I'd like to propose that we break that out into a separate PR. The reasons are:

1. It will make this PR smaller and more focussed (we will still have location sorting, it will just be sub-optimal)
2. I would propose that we additionally store the location as a string (actually `InternedStr`) on the task and resource location. This will avoid the need for extra allocations every time we redraw and I expect that it will be quite efficient as normally there are only a limited number of locations that tasks and resources get created from.

How does that sound?

mbiggio:
> Hi! 👋 Thank your PR!
> 
> I really like the change to fix the location sorting, that's a really great idea. However, I'd like to propose that we break that out into a separate PR. The reasons are:
> 
> 1. It will make this PR smaller and more focussed (we will still have location sorting, it will just be sub-optimal)
> 2. I would propose that we additionally store the location as a string (actually `InternedStr`) on the task and resource location. This will avoid the need for extra allocations every time we redraw and I expect that it will be quite efficient as normally there are only a limited number of locations that tasks and resources get created from.
> 
> How does that sound?

Hi @hds, yeah that makes a lot of sense, thanks for the comment!
Will change this PR and open a new one for the location

tokio-console/src/state/resources.rs
```diff
@@ -7,6 +7,7 @@ use crate::state::{
 use crate::view;
 use console_api as proto;
 use ratatui::{style::Color, text::Span};
+use std::borrow::Cow;
```
hds: Could we move this one into the group with the other `std` imports?

hds (changes requested):
One small nit regarding the imports. Thank you!

tokio-console/src/state/resources.rs
mbiggio: oh yeah right. Thank you for reviewing!

hds (approved):
Thank you! Approving now, and I'll merge later when I'm on my computer again (merging from the mobile app has some problems).

tokio-console/src/state/resources.rs
```diff
@@ -71,27 +76,50 @@ struct ResourceStats {
...
+                .sort_unstable_by_key(|resource| resource.upgrade().map(|r| r.borrow().visibility)),
+            Self::Location => resources.sort_unstable_by_key(|resource| {
+                resource.upgrade().map(|r| r.borrow().location.clone())
+            }),
+            Self::Attributes => resources.sort_unstable_by_key(|resource| {
+                resource.upgrade().map(|r| {
+                    // FIXME - we are taking only the first attribute as sorting key here,
+                    // and we are sorting each attribute as a String.
+                    // Instead, attributes should probably be parsed and sorted according to their actual values,
+                    // not just as strings.
```
hawkw: we should maybe open a new issue for adding this?

tokio-console/src/state/resources.rs
```diff
@@ -71,27 +76,50 @@ struct ResourceStats {
...
+                    r.borrow()
+                        .formatted_attributes()
+                        .first()
+                        .into_iter()
+                        .flat_map(|v| v.iter())
+                        .map(|s| <Cow<str> as std::borrow::Borrow<str>>::borrow(&s.content))
+                        .collect::<String>()
```
hawkw: i'm not sure why the `collect::<String>`  call is necessary here. can't we just call `.next()` on this iterator, returning the first element, since there should only be a single element in the iterator? that way, we don't have to allocate a big pile of `String`s every time we sort the list...
```suggestion
                        .next()
```

mbiggio: The thing is that `formatted_attributes()` returns a slice of vectors: `&[Vec<Span>]` , each one of the `Vec<Span>` representing an attribute, that is a key-value pair, such as `duration=7001ms`. In this case the vector is basically composed of `["duration", "=", "7001", "ms"]`.

Calling `next` instead of `collect` would basically just return the key of the attribute, `duration` in the example, while `collect` is returning a freshly allocated string `"duration=7001ms"`. With `next` we would be basically sorting on the key of the first attribute, without any new allocation, which could be fine for the moment. In any case we should probably open an issue to do a more meaningful sorting of attributes, and the same goes for location btw.

Do you think that would be fine?

tokio-console/src/state/resources.rs
mbiggio: Maybe the issue should track the improvement of attributes sorting in general? See answer to your other comment

tokio-console/src/state/resources.rs
hds: Let's just sort by the first attribute name for now. We will need to consider the best way to sort that column, since different resources have different attributes anyway.

Could you please open an issue for the improvement and then you can reference it in the FIXME (it's always nice to have an issue for each FIXME).

mbiggio: Thanks, sorting is now done only on the attribute name. Opened issue #496 and referenced it in the fixme

PR #493, opened 2023-11-25
feat: add flags and configurations for warnings

0xPoe:
I addressed the TODO item.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,27 @@ Options:
           
           [env: RUST_LOG=]
 
+      --linters <LINTERS>
+          Enable or disable specific linters.
+          
+          This is a comma-separated list of linters to enable or
+          disable.
+          
+          Each linter is specified by its name, which is one of:
+          
+          * `self-wake-percent` -- Warns when a task wakes itself more
+          than a certain percentage of its total wakeups.
+          
+          * `lost-waker` -- Warns when a task is dropped without being
+          woken.
+          
+          * `never-yielded` -- Warns when a task has never yielded.
+          
+          [default: self-wake-percent, lost-waker, never-yielded]
+          
+          [possible values: self-wake-percent, lost-waker,
+          never-yielded]
+
       --log-dir <LOG_DIRECTORY>
           Path to a directory to write the console's internal logs to.
           
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ Options:
           
           [env: RUST_LOG=]
 
-      --linters <LINTERS>
+      --linters <LINTERS>...
           Enable or disable specific linters.
           
           This is a comma-separated list of linters to enable or
--- a/README.md
+++ b/README.md
@@ -226,8 +226,7 @@ Options:
           
           * `never-yielded` -- Warns when a task has never yielded.
           
-          [default: self-wake-percent, lost-waker, never-yielded]
-          
+          [default: self-wake-percent lost-waker never-yielded]
           [possible values: self-wake-percent, lost-waker,
           never-yielded]
 
--- a/README.md
+++ b/README.md
@@ -210,13 +210,12 @@ Options:
           
           [env: RUST_LOG=]
 
-      --linters <LINTERS>...
-          Enable or disable specific linters.
+  -W, --warn <WARNS>...
+          Enable lint warnings.
           
-          This is a comma-separated list of linters to enable or
-          disable.
+          This is a comma-separated list of warnings to enable.
           
-          Each linter is specified by its name, which is one of:
+          Each warning is specified by its name, which is one of:
           
           * `self-wakes` -- Warns when a task wakes itself more than a
           certain percentage of its total wakeups.
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ Options:
           
           [env: RUST_LOG=]
 
-  -W, --warn <WARNS>...
+  -W, --warn <WARNINGS>...
           Enable lint warnings.
           
           This is a comma-separated list of warnings to enable.
--- a/README.md
+++ b/README.md
@@ -218,17 +218,16 @@ Options:
           
           Each linter is specified by its name, which is one of:
           
-          * `self-wake-percent` -- Warns when a task wakes itself more
-          than a certain percentage of its total wakeups.
+          * `self-wakes` -- Warns when a task wakes itself more than a
+          certain percentage of its total wakeups.
           
           * `lost-waker` -- Warns when a task is dropped without being
           woken.
           
           * `never-yielded` -- Warns when a task has never yielded.
           
-          [default: self-wake-percent lost-waker never-yielded]
-          [possible values: self-wake-percent, lost-waker,
-          never-yielded]
+          [default: self-wakes lost-waker never-yielded]
+          [possible values: self-wakes, lost-waker, never-yielded]
 
       --log-dir <LOG_DIRECTORY>
           Path to a directory to write the console's internal logs to.
--- a/README.md
+++ b/README.md
@@ -218,7 +218,8 @@ Options:
           Each warning is specified by its name, which is one of:
           
           * `self-wakes` -- Warns when a task wakes itself more than a
-          certain percentage of its total wakeups.
+          certain percentage of its total wakeups. Default percentage is
+          50%.
           
           * `lost-waker` -- Warns when a task is dropped without being
           woken.
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -62,7 +62,7 @@ pub struct Config {
     ///
     /// * `never-yielded` -- Warns when a task has never yielded.
     ///
-    #[clap(long = "linters")]
+    #[clap(long = "linters", value_delimiter = ',', num_args = 1..)]
     #[clap(default_values_t = vec![
         KnownWarnings::SelfWakePercent,
         KnownWarnings::LostWaker,
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -1,4 +1,6 @@
+use crate::state::tasks::Task;
 use crate::view::Palette;
+use crate::warnings;
 use clap::builder::{PossibleValuesParser, TypedValueParser};
 use clap::{ArgAction, ArgGroup, CommandFactory, Parser as Clap, Subcommand, ValueHint};
 use clap_complete::Shell;
@@ -48,6 +50,20 @@ pub struct Config {
     #[clap(long = "log", env = "RUST_LOG")]
     log_filter: Option<LogFilter>,
 
+    /// Enable or disable specific linters.
+    ///
+    /// This is a comma-separated list of linters to enable or disable.
+    ///
+    /// Each linter is specified by its name, which is one of:
+    /// * `self-wake-percent` -- Warns when a task wakes itself more than a
+    ///  certain percentage of its total wakeups.
+    /// * `lost-waker` -- Warns when a task is dropped without being woken.
+    /// * `never-yielded` -- Warns when a task has never yielded.
+    ///
+    /// [default: self-wake-percent, lost-waker, never-yielded]
+    #[clap(long = "linters")]
+    pub(crate) linters: Vec<KnownWarnings>,
+
     /// Path to a directory to write the console's internal logs to.
     ///
     /// [default: /tmp/tokio-console/logs]
@@ -96,6 +112,26 @@ pub struct Config {
     /// attempting to connect to a remote server.
     #[clap(subcommand)]
     pub subcmd: Option<OptionalCmd>,
+}
+
+/// Known warnings that can be enabled or disabled.
+#[derive(clap::ValueEnum, Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub(crate) enum KnownWarnings {
+    SelfWakePercent,
+    LostWaker,
+    NeverYielded,
+}
+
+impl From<&KnownWarnings> for warnings::Linter<Task> {
+    fn from(warning: &KnownWarnings) -> Self {
+        match warning {
+            KnownWarnings::SelfWakePercent => {
+                warnings::Linter::new(warnings::SelfWakePercent::default())
+            }
+            KnownWarnings::LostWaker => warnings::Linter::new(warnings::LostWaker),
+            KnownWarnings::NeverYielded => warnings::Linter::new(warnings::NeverYielded::default()),
+        }
+    }
 }
 
 #[derive(Debug, Subcommand, PartialEq, Eq)]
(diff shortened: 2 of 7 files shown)


tokio-console/src/config.rs
```diff
@@ -48,6 +50,22 @@ pub struct Config {
...
+    ///
+    /// Each linter is specified by its name, which is one of:
+    ///
+    /// * `self-wake-percent` -- Warns when a task wakes itself more than a certain percentage of its total wakeups.
+    ///
+    /// * `lost-waker` -- Warns when a task is dropped without being woken.
+    ///
+    /// * `never-yielded` -- Warns when a task has never yielded.
+    ///
+    /// [default: self-wake-percent, lost-waker, never-yielded]
```
hds: I tested this locally and this doesn't seem to be working. By default no lints are enabled.

It also seems that the only way to specify all lints is to provide the `--linters` flag multiple times, like this:

```sh
--linters lost-waker --linters self-wake-percent --linters never-yielded
```

whereas, from the name, I expected to be able to do this:

```sh
--linters lost-waker,self-wake-percent,never-yielded
```

I think that we should either have the second way, or the first way should be called `--linter` (or `--lint`) instead of `--linters` in plural, since you can only specify one at a time.

tokio-console/src/config.rs
```diff
@@ -98,6 +116,26 @@ pub struct Config {
     pub subcmd: Option<OptionalCmd>,
 }
 
+/// Known warnings that can be enabled or disabled.
+#[derive(clap::ValueEnum, Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub(crate) enum KnownWarnings {
+    SelfWakePercent,
```
hds: I feel that the "percent" bit here isn't necessary (I can't think of another way we could lint self wakes). Let's rename this `SelfWakes`.

```suggestion
    SelfWakes,
```

tokio-console/src/config.rs
```diff
@@ -48,6 +50,22 @@ pub struct Config {
...
+    ///
+    /// * `self-wake-percent` -- Warns when a task wakes itself more than a certain percentage of its total wakeups.
+    ///
+    /// * `lost-waker` -- Warns when a task is dropped without being woken.
+    ///
+    /// * `never-yielded` -- Warns when a task has never yielded.
+    ///
+    /// [default: self-wake-percent, lost-waker, never-yielded]
+    #[clap(long = "linters")]
+    pub(crate) linters: Vec<KnownWarnings>,
```
hds: One other part of this issue that would be really great is the ability to tune the lint parameters. For example, the self wakes warning and the never yielded warning both have thresholds which could be configured.

Before we decide on how the command line arguments should be, it would be great to see if there's a way to model this iteratively.

For example, could we have something like: `--lint never-yielded=2s` to set the threshold to 2 seconds instead of the default 1 second for that lint?

hds (changes requested):
Thanks for working on this, it's a great feature to have!

For me the default lints aren't working, and I think we need to consider how the command line arguments should be.

tokio-console/src/config.rs
hawkw: this is a bit nitpicky, but i'm not convinced i like the name `linters` for this argument. elsewhere, the documentation refers to these as "warnings", rather than as "lints", and i think it might be better to use `--warn`/`-W` as the name of the command line argument, instead, since that's consistent with the naming that `rustc` uses.

what do you think, @hds?

tokio-console/src/config.rs
```diff
@@ -48,6 +50,22 @@ pub struct Config {
...
+    ///
+    /// This is a comma-separated list of linters to enable or disable.
+    ///
+    /// Each linter is specified by its name, which is one of:
+    ///
+    /// * `self-wake-percent` -- Warns when a task wakes itself more than a certain percentage of its total wakeups.
+    ///
+    /// * `lost-waker` -- Warns when a task is dropped without being woken.
+    ///
+    /// * `never-yielded` -- Warns when a task has never yielded.
```
hawkw: it would be really nice if the list of warnings that can be enabled could be generated programmatically, somehow. i'm not sure if there's an easy way to do this, but we could potentially look into a macro or something.

i'm fine with that being something we add later, though; it doesn't have to be part of this PR.

tokio-console/src/config.rs
hawkw: I definitely agree that it would be nice to include a way to configure parameters for warnings, potentially in the CLI or at least in the config file. However, I also kind of think that just adding a way to configure what warnings are enabled is already better than the status quo, where there's no configuration surface for warnings, so I would personally be okay with moving forwards with this PR and then adding configuration for individual lints separately...

hds: OK. Agreed, let's go ahead with this PR and consider how to configure warning parameters later on.

tokio-console/src/config.rs
0xPoe: Oops. I only added the default values to the config file instead of the CLI flags.

0xPoe: Fixed in https://github.com/tokio-rs/console/pull/493/commits/c3eae4ff966a485cc62b1f48d692c6fbfc18f600

image

tokio-console/src/config.rs
0xPoe: Fixed in https://github.com/tokio-rs/console/pull/493/commits/d161904d91356f90a06f05e6381bb4a45cf08d92

tokio-console/src/config.rs
0xPoe: Good idea. I will try to support it in my next PR.

tokio-console/src/config.rs
0xPoe: Sounds good to me. We can follow the rustc style.

0xPoe: Fixed in https://github.com/tokio-rs/console/pull/493/commits/87559eac5e7f0804ff7181cb27cb5cf9b11c37a0

README.md
```diff
@@ -210,6 +210,24 @@ Options:
           
           [env: RUST_LOG=]
 
+  -W, --warn <WARNS>...
```
0xPoe: I'm not sure whether we should go with `--warn` or `--warns` as the command-line option.

0xPoe: Because we also have a config file, in the config file we call it `warns`.

tokio-console/src/config.rs
```diff
@@ -98,6 +120,35 @@ pub struct Config {
...
+    fn from(warning: &KnownWarnings) -> Self {
+        match warning {
+            KnownWarnings::SelfWakes => warnings::Linter::new(warnings::SelfWakePercent::default()),
+            KnownWarnings::LostWaker => warnings::Linter::new(warnings::LostWaker),
+            KnownWarnings::NeverYielded => warnings::Linter::new(warnings::NeverYielded::default()),
+        }
+    }
+}
+
+impl fmt::Display for KnownWarnings {
```
0xPoe, line 139: Maybe we can avoid implementing it manually.

0xPoe (commented):
🔢 Self-check

README.md
hawkw: we should call it `--warn`, for consistency with `rustc`, and also because the correct pluralization would be "warnings" rather than "warns".

tokio-console/src/config.rs
```diff
@@ -240,6 +291,7 @@ impl FromStr for LogFilter {
 struct ConfigFile {
     default_target_addr: Option<String>,
     log: Option<String>,
+    warns: Vec<KnownWarnings>,
```
hawkw: let's rename this field to
```suggestion
    warnings: Vec<KnownWarnings>,
```
as "warns" is not the correct pluralization.

tokio-console/src/config.rs
```diff
@@ -48,6 +50,26 @@ pub struct Config {
     #[clap(long = "log", env = "RUST_LOG")]
     log_filter: Option<LogFilter>,
 
+    /// Enable lint warnings.
+    ///
+    /// This is a comma-separated list of warnings to enable.
+    ///
+    /// Each warning is specified by its name, which is one of:
+    ///
+    /// * `self-wakes` -- Warns when a task wakes itself more than a certain percentage of its total wakeups.
```
hawkw, line 59: since the percentage is not currently configurable, we should probably include the default percentage in the documentation for now.

tokio-console/src/config.rs
hawkw: i think the manual implementation is fine, at least for now...

tokio-console/src/config.rs
```diff
@@ -48,6 +50,26 @@ pub struct Config {
...
+    ///
+    /// * `never-yielded` -- Warns when a task has never yielded.
+    ///
+    #[clap(long = "warn", short = 'W', value_delimiter = ',', num_args = 1..)]
+    #[clap(default_values_t = vec![
+        KnownWarnings::SelfWakes,
+        KnownWarnings::LostWaker,
+        KnownWarnings::NeverYielded
+    ])]
+    pub(crate) warns: Vec<KnownWarnings>,
```
hawkw: let's rename the field to 
```suggestion
    pub(crate) warnings: Vec<KnownWarnings>,
```

as that's the correct plural. also, in the help text, the argument should be displayed as `--warn <WARNINGS>` rather than `--warn <WARNS>`, which i think is a bit nicer.

tokio-console/src/config.rs
```diff
@@ -48,6 +50,26 @@ pub struct Config {
...
+    ///
+    /// Each warning is specified by its name, which is one of:
+    ///
+    /// * `self-wakes` -- Warns when a task wakes itself more than a certain percentage of its total wakeups.
+    ///
+    /// * `lost-waker` -- Warns when a task is dropped without being woken.
+    ///
+    /// * `never-yielded` -- Warns when a task has never yielded.
+    ///
+    #[clap(long = "warn", short = 'W', value_delimiter = ',', num_args = 1..)]
```
hawkw, line 66: if i understand correctly, the use of `num_args = 1..` mean that there's currently no way to disable _all_ warnings from the CLI currently? since passing `--warn` with no warning names is invalid, and by default we enable all warnins.

hawkw (changes requested):
overall, this looks pretty close! i had some fairly minor suggestions.

tokio-console/src/config.rs
0xPoe: Updated. https://github.com/tokio-rs/console/pull/493/commits/23cdd53fe1d1f5b7fbc37ca86aa9f15992c9c41b

tokio-console/src/config.rs
0xPoe: Added https://github.com/tokio-rs/console/pull/493/commits/4c76370ad5e8a5f6df048ad3785e4bb972e813a9

tokio-console/src/config.rs
0xPoe: Updated.

tokio-console/src/config.rs
0xPoe: As we discussed on Discord, we will add the `--Allow` flag to disable warnings. @hds proposed adding the `ALL` value to `--Allow` to disable all warnings.

0xPoe (commented):
🔢 Self-check

0xPoe:
@hds @hawkw Friendly ping~ 
Could you please take a look? Thanks! 💚 💙 💜 💛 ❤️

0xPoe:
I've rebased the PR. @hds Could you please take a look when you have time? Then we can move forward to add the following `--Allow` flag.

hds (approved):
This looks good now from my side. @hawkw, would you mind having a look as well?

0xPoe:
The failed test caused by this issue: https://github.com/assert-rs/trycmd/issues/121

I've bumped the trycmd version to fix it [in this commit.](https://github.com/tokio-rs/console/pull/493/commits/a359250fd1e45999742d96ebfbca28af299a8295)

0xPoe:
It passed now.

tokio-console/src/config.rs
```diff
@@ -442,6 +502,11 @@ impl Default for Config {
             log_filter: Some(LogFilter(
                 filter::Targets::new().with_default(filter::LevelFilter::OFF),
             )),
+            warnings: vec![
+                KnownWarnings::SelfWakes,
+                KnownWarnings::LostWaker,
+                KnownWarnings::NeverYielded,
+            ],
```
hawkw: this list of default warnings is repeated here and in the CLI arguments. it would be nice to define it in one place, perhaps like this:
```rust
impl KnownWarnings {
    fn default_enabled_warnings() -> Vec<Self> {
           vec![
                KnownWarnings::SelfWakes,
                KnownWarnings::LostWaker,
                KnownWarnings::NeverYielded,
            ]
    }
}
```

then, we could change this to:
```suggestion
            warnings: KnownWarnings::default_enabled_warnings(),
```
and change the `clap(default_values_t = ...)` attribute in the CLI args to:
```rust
    #[clap(default_values_t = KnownWarnings::default_enabled_warnings())]
```

This way, if we change the list of warnings that are enabled by default, such as by removing a default warning or adding a new one, we can update that list in only one place, and both the default config file and the CLI argument defaults will reflect that change.

tokio-console/src/config.rs
```diff
@@ -428,6 +481,13 @@ impl Config {
             log_directory: other.log_directory.or(self.log_directory),
             target_addr: other.target_addr.or(self.target_addr),
             log_filter: other.log_filter.or(self.log_filter),
+            warnings: {
+                let mut warns: Vec<KnownWarnings> = other.warnings;
+                warns.extend(self.warnings);
+                warns.sort_unstable();
+                warns.dedup();
+                warns
```
hawkw, line 495: Potentially, we might want to store the list of known warnings as a `BTreeSet` rather than as a `Vec`? That way, the properties that the list is sorted and contains no duplicates will be ensured at all times, rather than just when we ensure they are true.

On the other hand, I'm not sure whether that changes how it's deserialized from TOML...so, just a thought.

hawkw (approved):
Overall, this looks good to me! I had a couple fairly minor suggestions you may want to address, but nothing that would block merging IMO.

tokio-console/src/config.rs
0xPoe: Good suggestion! Addressed.

tokio-console/src/config.rs
0xPoe: > On the other hand, I'm not sure whether that changes how it's deserialized from TOML...so, just a thought.

`default_values_t` seems to only support the vec as the default value.

hawkw (approved):
looks great, thank you!

hawkw:
I restarted the failed CI job, since the failure looks flaky (possibly a racy test --- cc @hds): https://github.com/tokio-rs/console/actions/runs/7585395360/job/20661161429?pr=493#step:7:64

PR #494, opened 2023-11-25
test: make sure do not ignore the UI tests

0xPoe:
Found it from https://github.com/tokio-rs/console/actions/runs/6988785469/job/19016699616?pr=493.

If the binary does not exist, it will ignore the test cases. 

After this change, it will bail out an error.

I moved it to integration tests dir, we don't need to build the binary manually.


0xPoe (commented):
🔢 Self-check

0xPoe (commented):
🔢 Self-check

README.md
```diff
@@ -168,7 +168,7 @@ its behavior. The `help` command will print a list of supported command-line
 flags and arguments:
 
 ```console
-$ tokio-console --help
+$ tokio-console[EXE] --help
```
hds: I don't understand what the `[EXE]` is for. Could you please explain?

tokio-console/tests/ui/mod.rs
```diff
@@ -0,0 +1,23 @@
+use std::env;
```
hds: Generally I think it's easier to read if we use a single file module, rather than a directory.

Could you please change this file to `tokio-console/tests/ui.rs`.

hds (changes requested):

README.md
0xPoe: So on the Windows platform, we use the tokio-console.exe instead of the tokio-console. The `[EXE]` will handle this case if we run it on the Windows platform.

tokio-console/tests/ui/mod.rs
0xPoe: Sounds good. I will move it out.

README.md
hds: Got it. Thanks!

0xPoe:
image

hds (approved):
Looks good. Thank you!

0xPoe:

Issue #497, opened 2023-12-01
subscriber: add grpc-web support to console-subscriber


0xPoe:
### What problem are you trying to solve?

Currently, we only support standard gRPC. However, if we aim to develop a web console in the browser, it will require the use of [grpc-web](https://github.com/grpc/grpc-web) to connect to the console-subscriber. Since grpc-web only supports HTTP 1.1, we cannot directly connect to the console-subscriber. To facilitate this, we must use [envoy](https://www.envoyproxy.io/) as a proxy for forwarding. While this approach is effective, it introduces a new component to the system.

### How should the problem be solved?

Because tonic has [`tonic-web`](https://github.com/hyperium/tonic/tree/master/tonic-web) layer, we can add a new feature to help users to enable it. Then we can connect the console-subscriber with grpc-web. I have implemented it. It worked well.

### Any alternatives you've considered?

Use envoy as a proxy to workaround.

### How would users interact with this feature?

Users only need to enable the `grpc-web` or `web` feature.

### Would you like to work on this feature?

yes

hawkw:
This seems like a good idea to me! A web console would be neat to have eventually!

yerke:
@hi-rustin since https://github.com/tokio-rs/console/pull/498 got merged, should this issue be closed?

0xPoe:
Yes, I think we can close it now.

0xPoe:
Thank you

PR #498, opened 2023-12-01
feat(subscriber): support grpc-web and add `grpc-web` feature

0xPoe:
ref https://github.com/tokio-rs/console/issues/497.

See more at https://discord.com/channels/500028886025895936/838895414455435335/1180164585236471898

## Description
This pull request adds support for `grpc-web` to `console-subscriber`. Once you enable this feature by calling the `enable_grpc_web` function, you can connect the console-subscriber gRPC server using a browser client.

## Explanation of Changes
1. Added a new feature called `grpc-web` which requires the `tonic-web` crate as a dependency.
2. A new API named `serve_with_grpc_web` has been introduced. It appears to be similar to the `serve_with` API. However, if we were to use the same API with `serve_with`, it would result in a bound issue. We attempted to combine `serve_with_grpc_web` and `serve_with`, but it would create a very complex trait bound for the function. Therefore, we decided to introduce a new API to address this problem.
4. Added a new example named `grpc_web` to show how to use the `into_parts` API to customize the CORS layer.


0xPoe:
Tested in my console-web project(still private, I will open source it after I finish the basic features and views):
js code:
```vue
<script>
import { createPromiseClient } from "@connectrpc/connect";
import { createGrpcWebTransport } from "@connectrpc/connect-web";
import { Instrument } from "./gen/instrument_connect";
import { InstrumentRequest } from "./gen/instrument_pb";

const transport = createGrpcWebTransport({
  baseUrl: "http://127.0.0.1:8888",
});

const client = createPromiseClient(Instrument, transport);
const stream = client.watchUpdates(new InstrumentRequest());
(async () => {
  for await (const value of stream) {
    console.log(value);
  }
})();
</script>

```

console:

image

console-subscriber/Cargo.toml
```diff
@@ -28,6 +28,7 @@ keywords = [
 default = ["env-filter"]
 parking_lot = ["parking_lot_crate", "tracing-subscriber/parking_lot"]
 env-filter = ["tracing-subscriber/env-filter"]
+grpc-web = ["tonic-web"]
```
0xPoe, line 31: Do you have any opinions about the feature name?

console-subscriber/examples/grpc_web.rs
```diff
@@ -0,0 +1,165 @@
...
+    burn        Includes a (misbehaving) task that spins CPU with self-wakes
+    coma        Includes a (misbehaving) task that forgets to register a waker
+    noyield     Includes a (misbehaving) task that spawns tasks that never yield
+"#;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    ConsoleLayer::builder()
+        .with_default_env()
+        .enable_grpc_web(true)
```
0xPoe: I copied the example app and enabled the gRPC web support.

console-subscriber/src/lib.rs
```diff
@@ -922,7 +922,39 @@ impl Server {
...
     /// [`serve_with`]: Server::serve_with
     pub async fn serve(self) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
-        self.serve_with(tonic::transport::Server::default()).await
+        self.serve_with(tonic::transport::Server::builder()).await
+    }
+
+    #[cfg(feature = "grpc-web")]
+    /// Starts the gRPC service with the default gRPC settings and gRPC-Web
+    /// support.
+    pub async fn serve_with_grpc_web(
```
0xPoe: I am not sure if there is a better way to do this. I copied the `start_with` function here.

0xPoe (commented):
🔢 Self-check

console-subscriber/examples/grpc_web.rs
```diff
@@ -0,0 +1,165 @@
...
+//! Run with:
+//! ```sh
+//! cargo run --example grpc_web --features grpc-web
+//! ```
+use std::time::Duration;
+
+use console_subscriber::ConsoleLayer;
+
+static HELP: &str = r#"
+Example console-instrumented app
```
hds: Can we update this short description to mentioned grpc_web as well.

console-subscriber/src/lib.rs
hds: Looking at this function, it seems that it's exactly the same as `serve_with()` except for the `.accept_http1(true)` line. Is that correct?

If that's the case, I'm not even sure that we need a separate method for this. Instead, could the user of the API do the following:

```rust
let (console_layer, server) = console_subscriber::ConsoleLayer::new();
// Do something with the console_layer
server.serve_with(tonic::transport::Server::default().accept_http1(true)).await;
```
In that case, I think we can add documentation to `serve_with` to that effect.

I still think that perhaps being able to use `spawn` to launch a fully configured grpc-web server is valuable (otherwise, gprc-web forces the user to use the super custom way to define everything, which has some drawbacks).

What do you think?

console-subscriber/Cargo.toml
hds: I think that `grpc-web` sounds fine.

console-subscriber/src/builder.rs
```diff
@@ -501,8 +520,21 @@ impl Builder {
...
+                        server
+                            .serve()
+                            .await
+                            .expect("console subscriber server failed")
+                    }
+
+                    #[cfg(not(feature = "grpc-web"))]
                     server
                         .serve()
                         .await
```
hds: Having the `else` implementation effectively twice isn't ideal. I wonder if there's a better way to do this.

If we only need to modify the transport (based on comments in `serve_with_grpc`), then perhaps something like this would work:


```suggestion
                    let transport = tonic::transport::Server::default();
                    
                    #[cfg(feature = "grpc-web")]
                    let transport = transport.accept_http1(self.enable_grpc_web);
                    
                    server
                        .serve_with(transport)
                        .await
```

Note, I haven't actually tried to compile this code.

hds (changes requested):
This is looking great. I've tried to answer your questions and give some feedback.

I'm not sure how feasible it would be, but is there any way we could include a single HTML+JS file which is able to read from the gRPC-web endpoint as part of this example? Or do we need to pull in too many JS dependencies?

console-subscriber/src/lib.rs
0xPoe: > Looking at this function, it seems that it's exactly the same as `serve_with()` except for the `.accept_http1(true)` line. Is that correct?

It's not only about HTTP 1.1, also we need to use `tonic_web::enable(instrument_server)` to wrap the service.

0xPoe:
> I'm not sure how feasible it would be, but is there any way we could include a single HTML+JS file which is able to read from the gRPC-web endpoint as part of this example? Or do we need to pull in too many JS dependencies?

I believe using pure HTML and JavaScript might be somewhat challenging due to the requirement of generating protocol files, necessitating a node.js project. However, I'm open to adding a basic node.js example. My plan is to use '[connect](https://connectrpc.com/docs/web/getting-started/)' as it greatly simplifies the process of generating protocol files. What are your thoughts on this approach?

console-subscriber/examples/grpc_web.rs
0xPoe: I will update it later.

hds:
That sounds like a good approach, using connect. I'm not very familiar with JS tech these days, I just think that if we're adding support for gRPC-web then it would be nice to have a small example which shows it in use.

Thank you!

console-subscriber/src/lib.rs
hds: It looks like maybe it's possible to do this by adding a layer: `.layer(GrpcWebLayer::new())`?

See the second example here: https://docs.rs/tonic-web/latest/tonic_web/

0xPoe: After we added the layer, it would not be a server and we can not pass it to serve_with.

hds: It looks like it's still a server, but not an identity server (`Server<L = [Identity]>`). I wonder if we can get away with modifying `serve_with` to accept a generic server without making it a breaking API change.

0xPoe: Although we can pass it to `serve_with()`, I don't think it's a good API for users. This is because `GrpcWebLayer::new()` doesn't set the CORS configuration, which is usually required for web applications.

0xPoe: @hds Could you please provide some help here? I am not sure what kind of generic I can use here. Because we want to accept both `Server<L = [Identity]>` and `Server<
            tower::layer::util::Stack<
                tonic_web::GrpcWebLayer,
                tower::layer::util::Stack<
                    tower_http::cors::CorsLayer,
                    tower::layer::util::Identity,
                >,
            >,
        >` here. Do you have any thoughts or suggestion?

0xPoe:
I've updated my PR again. 
1. A new API called serve_with_grpc_web will allow users to pass a server build and cors layer and it serve it.
2. For subscriber builder, we only server it with grpc support when grpc-web feature is enabled and users pass a cors layer.
3. Nothing changes to server_with, users can still use it to pass a builder with some basic gPRC settings.

The reason I made this change is that it seems the point of serve_with is not for supporting passing any server builder with layers, it is just for doing some basic gRPC settings. If I am wrong, please correct me. Thanks!

console-subscriber/src/builder.rs
```diff
@@ -268,6 +274,14 @@ impl Builder {
         Self { self_trace, ..self }
     }
 
+    #[cfg(feature = "grpc-web")]
+    pub fn with_cors(self, cors: tower_http::cors::CorsLayer) -> Self {
+        Self {
+            cors_layer: Some(cors),
+            ..self
+        }
+    }
```
hds: I know that CORS is a common use case, but I think that we want to avoid exposing this directly in the public API.

One reason is that it adds types from (yet another) pre-1.0 crate to our public API, forcing us to release a breaking change every time that `tower-http` changes minor version. We already have this with `tonic` and it means that we have to bump our minor version every time that tonic does - we should avoid this if possible.

As we discussed on discord, I think that `into_parts()` is the correct solution to this, even if it does make things a bit longer.

On the other hand, we should definitely include a full set up including CORS in the example.

console-subscriber/src/builder.rs
```diff
@@ -56,6 +56,10 @@ pub struct Builder {
     /// Any scheduled times exceeding this duration will be clamped to this
     /// value. Higher values will result in more memory usage.
     pub(super) scheduled_duration_max: Duration,
+
+    #[cfg(feature = "grpc-web")]
+    /// Cors layer for grpc-web.
+    cors_layer: Option<tower_http::cors::CorsLayer>,
```
hds: As described below, I think this shouldn't be part of `console-subscriber`.

console-subscriber/src/builder.rs
```diff
@@ -71,6 +75,8 @@ impl Default for Builder {
             recording_path: None,
             filter_env_var: "RUST_LOG".to_string(),
             self_trace: false,
+            #[cfg(feature = "grpc-web")]
+            cors_layer: None,
```
hds: As described below, I think this shouldn't be part of `console-subscriber`.

console-subscriber/examples/grpc_web.rs
```diff
@@ -0,0 +1,191 @@
...
+                    .name("noyield")
+                    .spawn(no_yield(20))
+                    .unwrap();
+            }
+            "blocking" => {
+                tokio::task::Builder::new()
+                    .name("spawns_blocking")
+                    .spawn(spawn_blocking(5))
+                    .unwrap();
+            }
```
hds: What do you think about making this example simpler? This will avoid duplicating so much code from the `app` example.

Alternatively, I would instead extend the `app` example to support gRPC-web based on command line flags and make it the "catch-all-example" that includes everything.

console-subscriber/src/lib.rs
```diff
@@ -922,7 +923,42 @@ impl Server {
     /// ```
     /// [`serve_with`]: Server::serve_with
     pub async fn serve(self) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
-        self.serve_with(tonic::transport::Server::default()).await
+        self.serve_with(tonic::transport::Server::builder()).await
```
hds: I don't understand the reason for changing this line. Can we keep it the same?

If not, we should update the documentation as well.

console-subscriber/src/lib.rs
```diff
@@ -922,7 +923,42 @@ impl Server {
...
+    }
+
+    #[cfg(feature = "grpc-web")]
+    /// Starts the gRPC service with the default gRPC settings and gRPC-Web
+    /// support.
+    pub async fn serve_with_grpc_web(
```
hds: Let's move this below `serve_with`.

hds (changes requested):
As discussed, I think this is the right approach, however I believe we should keep the cors layer (and `tower-http`) out of our public API.

console-subscriber/src/lib.rs
0xPoe: Updated.

console-subscriber/src/lib.rs
0xPoe: I accidentally changed it. I've reverted it. Thanks!

console-subscriber/examples/grpc_web.rs
0xPoe: I simplified the example only to keep task1 and task2. Also, include a full example for grpc-web with cors. Do you think we need to add another grpc-web with the default config?

console-subscriber/src/builder.rs
0xPoe: Removed.

console-subscriber/src/builder.rs
0xPoe: Removed.

console-subscriber/src/builder.rs
0xPoe: Removed. Make sense!

0xPoe (commented):
I planned to add a real web client example after we merge this API change first. I think it would help reduce the noise from adding a node app to it.

console-subscriber/examples/grpc_web.rs
```diff
@@ -0,0 +1,122 @@
...
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    let (console_layer, server) = ConsoleLayer::builder().with_default_env().build();
+    thread::Builder::new()
+        .name("subscriber".into())
+        .spawn(move || {
+            // Do not trace anything in this thread.
+            let _subscriber_guard;
+            _subscriber_guard =
+                tracing::subscriber::set_default(tracing_core::subscriber::NoSubscriber::default());
```
hds: nit: this can be put in a single statement in this case.

```suggestion
            let _subscriber_guard =
                tracing::subscriber::set_default(tracing_core::subscriber::NoSubscriber::default());
```

console-subscriber/src/builder.rs
```diff
@@ -268,6 +274,14 @@ impl Builder {
         Self { self_trace, ..self }
     }
 
+    #[cfg(feature = "grpc-web")]
+    pub fn enable_grpc_web(self, enable_grpc_web: bool) -> Self {
```
hds, line 292: Please add some documentation here. A brief description of what this option enables and also an explanation that  `serve_with_grpc_web` can be used to provide more advanced configuration options (including a link to the docs for that method).

console-subscriber/src/lib.rs
```diff
@@ -962,6 +962,38 @@ impl Server {
         res?.map_err(Into::into)
     }
 
+    #[cfg(feature = "grpc-web")]
+    /// Starts the gRPC service with the default gRPC settings and gRPC-Web
+    /// support.
+    pub async fn serve_with_grpc_web(
```
hds, line 1063: I think we should add an example here (see `into_parts` below) and a description of how this can be used to set CORS.

In fact, I think it would be good to include 2 examples, 1 basic one and 1 that is more complex (similar to or the same as the `grpc_web.rs` example) which shoulds how to set CORS middleware.

console-subscriber/examples/grpc_web.rs
hds: I don't think we need another full example, but we should have an example in the docs for `serve_with_grpc_web` as I mentioned there.

hds (changes requested):
Everything is looking great! I think we just need a bit more documentation (including some examples / doctests) and this PR will be good to merge.

console-subscriber/examples/grpc_web.rs
0xPoe: Updated! Good catch.

console-subscriber/src/builder.rs
0xPoe: Added.

console-subscriber/src/lib.rs
0xPoe: Added. Thanks!

0xPoe: image

0xPoe (commented):
🔢 Self-check

0xPoe (commented):
I have a concern about the default port for the `grpc-web` server. 6669 is a [restricted port](https://superuser.com/questions/188006/how-to-fix-err-unsafe-port-error-on-chrome-when-browsing-to-unsafe-ports) on chrome. So unless users change it, they won't be able to access it in chrome. So maybe we should change the default port when users enable grpc-web support. But I also worry that this would cause more chaos about the default value. Do you have any suggestions?

console-subscriber/src/builder.rs
```diff
@@ -56,6 +56,10 @@ pub struct Builder {
     /// Any scheduled times exceeding this duration will be clamped to this
     /// value. Higher values will result in more memory usage.
     pub(super) scheduled_duration_max: Duration,
+
+    #[cfg(feature = "grpc-web")]
+    /// Whether to enable the grpc-web support.
```
hds: Typically the `#[cfg]` goes below the doc comment:
```suggestion
    /// Whether to enable the grpc-web support.
    #[cfg(feature = "grpc-web")]
```

console-subscriber/src/builder.rs
```diff
@@ -268,6 +274,24 @@ impl Builder {
         Self { self_trace, ..self }
     }
 
+    #[cfg(feature = "grpc-web")]
```
hds: The `#[cfg]` line should be below the doc comments. It all works the same, but it is a bit easier to reason about.

hds:
> I have a concern about the default port for the grpc-web server. 6669 is a [restricted port](https://superuser.com/questions/188006/how-to-fix-err-unsafe-port-error-on-chrome-when-browsing-to-unsafe-ports) on chrome. So unless users change it, they won't be able to access it in chrome. So maybe we should change the default port when users enable grpc-web support. But I also worry that this would cause more chaos about the default value. Do you have any suggestions?

I wouldn't dynamically change the default port based on using grpc-web. However, it's probably worth setting the port to some other value in the grpc-web example, perhaps something typical like 8080. And also add a comment stating why a different port is used. That way, at least for people copying from the example, there will be extra information.

Perhaps also adding a line to the documentation for `enable_grpc_web` stating that some browsers may block the default port 6669 would be a good idea.

console-subscriber/src/lib.rs
```diff
@@ -962,6 +962,133 @@ impl Server {
...
+    /// # async fn docs() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+    /// # let (_, server) = console_subscriber::ConsoleLayer::new();
+    /// server.serve_with_grpc_web(tonic::transport::Server::default()).await
+    /// # }
+    /// ```
+    ///
+    /// To serve the instrument server with gRPC-Web support and a custom CORS configuration, use the
+    /// following code:
+    ///
+    /// ```rust
```
hds, line 983: This example is great!

console-subscriber/src/builder.rs
```diff
@@ -268,6 +274,24 @@ impl Builder {
...
+    ///
+    /// By default, this is `false`. If enabled, the console subscriber will
+    /// serve the gRPC-Web protocol in addition to the standard gRPC protocol.
+    /// This is useful for serving the console subscriber to web clients.
+    ///
+    /// [serve_with_grpc_web] is used to provide more advanced configuration
+    /// for the gRPC-Web server.
+    ///
+    /// [serve_with_grpc_web]: crate::Server::serve_with_grpc_web
+    pub fn enable_grpc_web(self, enable_grpc_web: bool) -> Self {
```
hds, line 292: As mentioned in the stand alone comment, perhaps it's worth adding a line here suggesting that setting a different port will overcome the behaviour of some browsers.

console-subscriber/src/builder.rs
```diff
@@ -268,6 +274,24 @@ impl Builder {
...
+    /// Sets whether to enable the grpc-web support.
+    ///
+    /// By default, this is `false`. If enabled, the console subscriber will
+    /// serve the gRPC-Web protocol in addition to the standard gRPC protocol.
+    /// This is useful for serving the console subscriber to web clients.
+    ///
+    /// [serve_with_grpc_web] is used to provide more advanced configuration
+    /// for the gRPC-Web server.
+    ///
+    /// [serve_with_grpc_web]: crate::Server::serve_with_grpc_web
```
hds: Use monospace for function names:
```suggestion
    /// [`serve_with_grpc_web`] is used to provide more advanced configuration
    /// for the gRPC-Web server.
    ///
    /// [`serve_with_grpc_web`]: crate::Server::serve_with_grpc_web
```

hds (commented):
Looking good, some suggestions based on your comment regarding the default port and also a couple of style suggestions that you could apply directly.

console-subscriber/src/builder.rs
0xPoe: Updated.

console-subscriber/src/builder.rs
0xPoe: Added.

0xPoe:
> However, it's probably worth setting the port to some other value in the grpc-web example, perhaps something typical like 8080.

I've already set it to 9999 before. It worked well on my different browsers. (chrome/arc/firefox/safari)



> Perhaps also adding a line to the documentation for `enable_grpc_web` stating that some browsers may block the default port 6669 would be a good idea.

Added. Thanks for your suggestion!

console-subscriber/examples/grpc_web.rs
```diff
@@ -0,0 +1,122 @@
...
+                tokio::spawn(aggregator.run());
+                let router = tonic::transport::Server::builder()
+                    // Accept gRPC-Web requests and enable CORS.
+                    .accept_http1(true)
+                    .layer(cors)
+                    .layer(GrpcWebLayer::new())
+                    .add_service(instrument_server);
+                let serve = router.serve(std::net::SocketAddr::new(
+                    std::net::IpAddr::V4(std::net::Ipv4Addr::new(127, 0, 0, 1)),
+                    // 6669 is a restricted port on Chrome, so we cannot use it. We use a different port instead.
```
0xPoe, line 73: cc: @hds

console-subscriber/src/lib.rs
```diff
@@ -962,6 +962,133 @@ impl Server {
...
+    ///
+    /// // Serve the instrument server with gRPC-Web support and the CORS configuration.
+    /// let router = tonic::transport::Server::builder()
+    ///     .accept_http1(true)
+    ///     .layer(cors)
+    ///     .layer(GrpcWebLayer::new())
+    ///     .add_service(instrument_server);
+    /// let serve = router.serve(std::net::SocketAddr::new(
+    ///     std::net::IpAddr::V4(std::net::Ipv4Addr::new(127, 0, 0, 1)),
+    /// // 6669 is a restricted port on Chrome, so we cannot use it. We use a different port instead.
```
0xPoe: cc: @hds

0xPoe (commented):
🔢 Self-check

console-subscriber/src/lib.rs
```diff
@@ -962,6 +962,133 @@ impl Server {
...
+    ///     9999,
```
hds: Small indentation fix.

```suggestion
    ///     std::net::IpAddr::V4(std::net::Ipv4Addr::new(127, 0, 0, 1)),
    ///     // 6669 is a restricted port on Chrome, so we cannot use it. We use a different port instead.
    ///     9999,
```

0xPoe:
@hds Before we move forward with this PR, is there anything I need to do?

hds (approved):
Looks good. Thanks for all the time you spent on this PR, it's great!

hds:
> @hds Before we move forward with this PR, is there anything I need to do?

@hi-rustin Sorry! I thought I merged this already. Done now! Thank you!

0xPoe:

PR #499, opened 2023-12-07
chore: add comments to h2 dependency

0xPoe:
Added comments to the h2 dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,7 +93,7 @@ dependencies = [
  "bitflags 1.3.2",
  "bytes",
  "futures-util",
- "http",
+ "http 0.2.9",
  "http-body",
  "hyper",
  "itoa",
@@ -119,7 +119,7 @@ dependencies = [
  "async-trait",
  "bytes",
  "futures-util",
- "http",
+ "http 0.2.9",
  "http-body",
  "mime",
  "rustversion",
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -93,7 +93,7 @@ dependencies = [
  "bitflags 1.3.2",
  "bytes",
  "futures-util",
- "http",
+ "http 0.2.9",
```
hds: We'll now have 2 versions of the `http` crate in our dependencies. I'm not against, this, but I'd like to double check that you're aware of this before merging.

hds (changes requested):
Thanks for your PR. Would you mind commenting on the fact that we'll now have 2 versions of the `http` crate in our dependency tree (because `tonic` hasn't updated to the latest `axum` yet, and `axum` 0.6 is still depending on `http` 0.2).

I'm not against merging this, I just want to make sure that we're aware of what we're doing.

seanmonstar:
I think the `h2` dependency is used in the Console to access stuff inside Tonic. In that case, they should be kept in sync.

hds:
> I think the `h2` dependency is used in the Console to access stuff inside Tonic. In that case, they should be kept in sync.

Yep, @seanmonstar is right, upgrading the version of `h2` would break this check:

https://github.com/tokio-rs/console/blob/ef61081caa6c20f9495563855969e8d820559dfe/tokio-console/src/conn.rs#L43-L47

Good catch!

@hi-rustin Let's close this PR.

seanmonstar:
As a side note, I'd recommend trying to figure out a way to do that without downcasting the error. Either in Console, or in Tonic. It is not part of [hyper's stability promise](https://docs.rs/hyper/latest/hyper/struct.Error.html#source).

0xPoe:
> upgrading the version of `h2` would break this check:

How about I add some comments about it? I think it would be useful to be aware of it.

seanmonstar:
A comment in the Cargo.toml file could be helpful, right next to the dependency declaration.

hds:
Yes, that's a good idea. It would be nice if we could add some sort of a unit test for the behaviour in `conn.rs` as well.

hds (approved):
Great, thanks for clarifying this!

0xPoe:

PR #501, opened 2023-12-14
fix(subscriber): Don't save poll_ops if no-one is receiving them

grahamking:
Do not record poll_ops if there are no current connected clients (watchers). Without this `Aggregator::poll_ops` would grow forever.

Follow up to https://github.com/tokio-rs/console/pull/311 and fix for these two:
- https://github.com/tokio-rs/console/issues/184
- https://github.com/tokio-rs/console/pull/500

### DIFF
--- a/console-subscriber/src/aggregator/mod.rs
+++ b/console-subscriber/src/aggregator/mod.rs
@@ -270,7 +270,7 @@ impl Aggregator {
             .drop_closed(&mut self.resource_stats, now, self.retention, has_watchers);
         self.async_ops
             .drop_closed(&mut self.async_op_stats, now, self.retention, has_watchers);
-        if !has_watchers && !self.poll_ops.is_empty() {
+        if !has_watchers {
             self.poll_ops.clear();
         }
     }
@@ -472,6 +472,7 @@ impl Aggregator {
                 task_id,
                 is_ready,
             } => {
+                // CLI doesn't show historical poll ops, so don't save them if no-one is watching
                 if self.watchers.is_empty() {
                     return;
                 }
--- a/console-subscriber/src/aggregator/mod.rs
+++ b/console-subscriber/src/aggregator/mod.rs
@@ -1,12 +1,3 @@
-use super::{Command, Event, Shared, Watch};
-use crate::{
-    stats::{self, Unsent},
-    ToProto, WatchRequest,
-};
-use console_api as proto;
-use proto::resources::resource;
-use tokio::sync::{mpsc, Notify};
-
 use std::{
     sync::{
         atomic::{AtomicBool, Ordering::*},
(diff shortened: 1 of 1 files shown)


console-subscriber/src/aggregator/mod.rs
```diff
@@ -1,21 +1,22 @@
...
+use proto::resources::resource;
+use tokio::sync::{mpsc, Notify};
 use tracing_core::{span::Id, Metadata};
 
+use super::{Command, Event, Shared, Watch};
+use crate::{
+    stats::{self, Unsent},
+    ToProto, WatchRequest,
+};
+
```
hawkw, line 19: this is a minor nit, but were these imports reordered by rustfmt or some other editor feature? and if so, can you ensure that you're using the default rustfmt settings? i'm open to merging style changes such as reordering imports, but would prefer to make them in separate PRs from those that make functional changes, and to ensure that if someone else edits this file with the default rustfmt settings,  that the change won't be undone. this helps to minimize churn in Git diffs.

thanks!

console-subscriber/src/aggregator/mod.rs
```diff
@@ -472,6 +472,9 @@ impl Aggregator {
                 task_id,
                 is_ready,
             } => {
+                if self.watchers.is_empty() {
```
hawkw, line 476: take it or leave it: maybe worth including a comment here explaining why we drop the poll op if no watches are connected, since this behavior is different from the behavior for other data such as tasks?

console-subscriber/src/aggregator/mod.rs
```diff
@@ -269,6 +270,9 @@ impl Aggregator {
             .drop_closed(&mut self.resource_stats, now, self.retention, has_watchers);
         self.async_ops
             .drop_closed(&mut self.async_op_stats, now, self.retention, has_watchers);
+        if !has_watchers && !self.poll_ops.is_empty() {
+            self.poll_ops.clear();
```
hawkw: nit, take it or leave it: I believe `clear` is a no-op if the collection is empty, so it should be more or less equivalent to just write
```suggestion
        if !has_watchers {
            self.poll_ops.clear();
```

hawkw (approved):
overall, this change looks good to me, thanks for the fix!

my one concern before merging this is just to make sure that the change to import ordering won't be undone if someone else saves this file while using default `rustfmt` settings. if you don't mind replying to my comment about that, I'll be happy to merge this PR once that's addressed!

thanks again!

console-subscriber/src/aggregator/mod.rs
grahamking: AFAICT yes this is default rustfmt.

I was confused too by this. I have rustfmt 1.7.0-nightly (a577704 2023-11-16). It definitely isn't loading a `.rustfmt.toml` (or `rustfmt.toml`). I `strace`d it and it tries a lot of different paths, all of them are "No such file or directory". I ran rustfmt from stable too. Same result. 

Originally I was going to leave that part our of the PR, but I figured if it really is default rustfmt, it should be included.

Let me know if you want me to remove it.

console-subscriber/src/aggregator/mod.rs
grahamking: Will do.

console-subscriber/src/aggregator/mod.rs
grahamking: Nice! I originally wrote it like that but then was worried about the cost of clear. Good to know, will add thanks.

ethercflow:
@hawkw Excuse me, when is this PR expected to be merged?

PR #503, opened 2023-12-15
feat(subscriber): Reduce retention period to fit in max message size

grahamking:
If the initial update message would be too big for tokio-console's grpc decoder, reduce the retention period and try again.

Currently the default retention period is 1 hour. That can easily grow to more than the max grpc decode message size (4 MiB), at which point tokio-console won't connect.

There's really no minimum safe duration for retention. It depends on how busy the app is and on how much trace data runtime and tokio log. Here we repeatedly divide the retention period in half until it fits in the message.

### DIFF
--- a/console-subscriber/src/aggregator/mod.rs
+++ b/console-subscriber/src/aggregator/mod.rs
@@ -294,8 +294,8 @@ impl Aggregator {
                     metadata: (*self.all_metadata).clone(),
                 }),
             };
-            let el = update.encoded_len();
-            if el < MAX_MESSAGE_SIZE {
+            let message_size = update.encoded_len();
+            if message_size < MAX_MESSAGE_SIZE {
                 // normal case
                 break Some(update);
             }
@@ -305,7 +305,7 @@ impl Aggregator {
             self.cleanup_closed();
             tracing::debug!(
                 retention = ?self.retention,
-                message_size = el,
+                message_size,
                 max_message_size = MAX_MESSAGE_SIZE,
                 "Message too big, reduced retention",
             );
--- a/console-subscriber/src/aggregator/mod.rs
+++ b/console-subscriber/src/aggregator/mod.rs
@@ -7,6 +7,7 @@ use std::{
 };
 
 use console_api as proto;
+use prost::Message;
 use proto::resources::resource;
 use tokio::sync::{mpsc, Notify};
 use tracing_core::{span::Id, Metadata};
@@ -21,6 +22,9 @@ mod id_data;
 mod shrink;
 use self::id_data::{IdData, Include};
 use self::shrink::{ShrinkMap, ShrinkVec};
+
+/// Should match tonic's (private) codec::DEFAULT_MAX_RECV_MESSAGE_SIZE
+const MAX_MESSAGE_SIZE: usize = 4 * 1024 * 1024;
 
 /// Aggregates instrumentation traces and prepares state for the instrument
 /// server.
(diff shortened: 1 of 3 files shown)


console-subscriber/src/aggregator/mod.rs
```diff
@@ -274,24 +282,48 @@ impl Aggregator {
...
+            // User will only get updates
+            self.watchers.push(subscription);
+            return;
+        }
 
         // Send the initial state --- if this fails, the subscription is already dead
-        if subscription.update(update) {
+        if subscription.update(&update.unwrap()) {
             self.watchers.push(subscription)
         }
```
hawkw: nit: we could remove the call to `update.unwrap()` if we combined these two `if`s into a single `match`, like this:
```rust
match update {
     // Send the initial state
    Some(update) => if !subscription.update(update) {
         // If sending the initial update fails, the subscription is already dead,
         // so don't add it to `watchers`.
         return;
    },
    // User will only get updates.
    None => tracing::error!(
        min_retention = ?MIN_RETENTION,
        "Message too big. Start with smaller retention.",
    ),
}

self.watchers.push(subscription);

File: console-subscriber/src/aggregator/mod.rs [hunk 2]
```diff
@@ -274,24 +282,48 @@ impl Aggregator {
...
+                new_metadata: Some(proto::RegisterMetadata {
+                    metadata: (*self.all_metadata).clone(),
+                }),
+            };
+            let el = update.encoded_len();
+            if el < MAX_MESSAGE_SIZE {
+                // normal case
+                break Some(update);
+            }
+            // If the grpc message is bigger than tokio-console will accept throw away the oldest
```
hawkw: nit: missing comma here:
```suggestion
            // If the grpc message is bigger than tokio-console will accept, throw away the oldest
```

console-subscriber/src/aggregator/mod.rs
```diff
@@ -274,24 +282,48 @@ impl Aggregator {
...
+                retention = ?self.retention,
+                message_size = el,
+                max_message_size = MAX_MESSAGE_SIZE,
+                "Message too big, reduced retention",
+            );
+
+            if self.retention <= MIN_RETENTION {
+                self.retention = MIN_RETENTION;
+                break None;
+            }
```
hawkw: I don't think hard-coding the minimum retention to 1 second makes sense. Updates are only published to watchers every time the `Aggregator`'s `publish_interval` elapses, which is a user-configurable value that defaults to 2 seconds. If we publish updates every two seconds, but only retain data for one second, we'll be throwing away half of the data about events that occurred since the last update, which seems like unacceptable data loss to me.

Instead, I think the minimum allowable retention duration probably ought to be `self.publish_interval`. 
```suggestion
            if self.retention <= self.publish_interval {
                self.retention = self.publish_interval;
                break None;
            }
```

This way, we will always retain data for at least long enough to include it in an update --- we would be discarding historical data that future clients might be interested in, but we would at least always be ensuring that actively connected clients are still receiving all real-time data as it's produced. 

Incidentally, this would mean that users could also tune the publish interval to allow running with lower retention values.

hawkw (changes requested):

console-subscriber/src/aggregator/mod.rs
grahamking: Nice, thanks.

grahamking:
@hawkw I think this is ready to merge. Could you take another look?

hawkw:
@grahamking  i'll give it a look today! thanks!

console-subscriber/src/aggregator/mod.rs
```diff
@@ -278,26 +282,57 @@ impl Aggregator {
...
+                // normal case
+                break Some(update);
+            }
+            // If the grpc message is bigger than tokio-console will accept, throw away the oldest
+            // inactive data and try again
+            self.retention /= 2;
+            self.cleanup_closed();
+            tracing::debug!(
+                retention = ?self.retention,
+                message_size = el,
```
hawkw: very tiny style nit, not a big deal: why not just use the name `message_size` for the variable, rather than `el`? i have a minor preference for `tracing` event fields to match the name of the variable, when possible...

hawkw (approved):
looks good to me! i had one last style nit, if you want to address it, but i'm also happy to merge this as-is!

console-subscriber/src/aggregator/mod.rs
grahamking: That is indeed nicer, thanks. Updated.

grahamking:
Can we re-trigger that test (without me pushing a blank commit)? I can't see how variable rename would break it.

PR #510, opened 2024-01-20
Update README.md

SimonsMine:
Added note about windows

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,6 +330,14 @@ Options:
 
 ```
 
+#### running the console on windows
+
+For best results on windows use a utf-8 enabled Terminal such as the new Windows [Terminal]([http://example.com](https://learn.microsoft.com/en-us/windows/terminal/install).
+In order to get graph output and other visual features to work, make sure to explicitly call tokio-console with the utf-8 language flag set.
+```
+tokio-console --lang en_US.UTF-8
+```
+
 ## for development
 
 the `console-subscriber/examples` directory contains **some potentially useful
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ Options:
 
 #### running the console on windows
 
-For best results on windows use a utf-8 enabled Terminal such as the new Windows [Terminal]([http://example.com](https://learn.microsoft.com/en-us/windows/terminal/install).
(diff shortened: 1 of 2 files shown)


0xPoe (changes requested):
image


It seemed you lost a `)` here.

README.md
```diff
@@ -312,6 +312,14 @@ Options:
 
 ```
 
+#### running the console on windows
+
+For best results on windows use a utf-8 enabled Terminal such as the new Windows [Terminal]([http://example.com](https://learn.microsoft.com/en-us/windows/terminal/install).
```
SimonsMine:
```suggestion
For best results on windows use a utf-8 enabled Terminal such as the new Windows [Terminal](https://learn.microsoft.com/en-us/windows/terminal/install).
```

SimonsMine:
fixed : )

README.md
```diff
@@ -312,6 +312,14 @@ Options:
 
 ```
 
+#### running the console on windows
+
+For best results on windows use a utf-8 enabled Terminal such as the new Windows [Terminal](https://learn.microsoft.com/en-us/windows/terminal/install).
+In order to get graph output and other visual features to work, make sure to explicitly call tokio-console with the utf-8 language flag set.
```
hawkw: I might rephrase this a bit:
```suggestion
The console uses the UTF-8 character set to display graphs and other visual
features in the terminal. In order to display this rich terminal UI on Windows,
it's necessary to use a UTF-8-enabled terminal emulator, such as the new
[Windows Terminal](https://learn.microsoft.com/en-us/windows/terminal/install).

If you're using a terminal that supports UTF-8, make sure to explicitly call
tokio-console with the UTF-8 language flag set: 
```

hawkw (changes requested):
Thanks for helping to improve the documentation!

I suggested some edits to the note on UTF-8 support, let me know what you think?

Also, would you mind adding a similar note to the `tokio-console` crate's README, in [the "running the console" section](https://github.com/tokio-rs/console/blob/main/tokio-console/README.md#using-the-console)? That documentation is rendered on https://crates.io/crates/tokio-console and https://docs.rs/tokio-console, so it would be nice to include the Windows support information there as well.

SimonsMine:
Applied suggestions : )

hawkw (approved):

0xPoe (approved):
Looks good to me. Thanks! 👍

PR #513, opened 2024-01-29
feat: add `--allow` flag

0xPoe:
ref https://github.com/tokio-rs/console/pull/493#discussion_r1410610481

## Description
This pull request adds a new flag to `tokio-console` that allows warning lints to be selectively disabled. Additionally, you can use the value `all` to disable all warnings.
For example, you can use it like this: following manner: `cargo run -- --warn "self-wakes,lost-waker,never-yielded" --allow "self-wakes,never-yielded"`

## Explanation of Changes
Added an `AllowedWarnings` enum to the command line interface to represent the allowed warnings. `All` for disable all warnings. `Explicit` explicitly indicates which warning can be ignored.

### DIFF
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -79,7 +79,7 @@ pub struct Config {
     ///
     /// * `never-yielded` -- Warns when a task has never yielded.
     #[clap(long = "allow", short = 'A')]
-    pub(crate) allow_warnings: AllowedWarnings,
+    pub(crate) allow_warnings: Option<AllowedWarnings>,
 
     /// Path to a directory to write the console's internal logs to.
     ///
@@ -203,10 +203,6 @@ impl From<&str> for AllowedWarnings {
 }
 
 impl AllowedWarnings {
-    fn default() -> Self {
-        AllowedWarnings::Some(BTreeSet::default())
-    }
-
     fn merge(&self, allowed: &Self) -> Self {
         match (self, allowed) {
             (AllowedWarnings::All, _) => AllowedWarnings::All,
@@ -363,7 +359,7 @@ struct ConfigFile {
     default_target_addr: Option<String>,
     log: Option<String>,
     warnings: Vec<KnownWarnings>,
-    allow_warnings: AllowedWarnings,
+    allow_warnings: Option<AllowedWarnings>,
     log_directory: Option<PathBuf>,
     retention: Option<RetainFor>,
     charset: Option<CharsetConfig>,
@@ -559,7 +555,12 @@ impl Config {
                 warns.dedup();
                 warns
             },
-            allow_warnings: { self.allow_warnings.merge(&other.allow_warnings) },
+            allow_warnings: {
+                match (self.allow_warnings, other.allow_warnings) {
+                    (Some(a), Some(b)) => Some(a.merge(&b)),
+                    (a, b) => a.or(b),
+                }
+            },
             retain_for: other.retain_for.or(self.retain_for),
             view_options: self.view_options.merge_with(other.view_options),
             subcmd: other.subcmd.or(self.subcmd),
@@ -575,7 +576,7 @@ impl Default for Config {
                 filter::Targets::new().with_default(filter::LevelFilter::OFF),
             )),
             warnings: KnownWarnings::default_enabled_warnings(),
-            allow_warnings: AllowedWarnings::default(),
+            allow_warnings: None,
             log_directory: Some(default_log_directory()),
             retain_for: Some(RetainFor::default()),
             view_options: ViewOptions::default(),
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -73,14 +73,14 @@ pub struct Config {
     ///
     /// Each warning is specified by its name, which is one of:
     ///
-    /// * `all` -- Allow all warnings.
-    ///
     /// * `self-wakes` -- Warns when a task wakes itself more than a certain percentage of its total wakeups.
     ///                  Default percentage is 50%.
     ///
     /// * `lost-waker` -- Warns when a task is dropped without being woken.
     ///
     /// * `never-yielded` -- Warns when a task has never yielded.
+    ///
+    /// If this is set to `all`, all warnings are allowed.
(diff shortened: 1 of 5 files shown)


0xPoe (commented):
Still working on adding the `--allow all` support.

0xPoe:
Tested locally:
1. panic on main.rs with `panic!("{:?}", warnings);`
2. Try:
```console
$ cargo run -- --warn "self-wakes" --allow all
     Running `target/debug/tokio-console --warn self-wakes --allow all`
The application panicked (crashed).
Message:  []
Location: tokio-console/src/main.rs:79
```
3. Another one:
```console
$ cargo run -- --warn "self-wakes" --allow ohno
error: invalid value 'ohno' for '--allow <ALLOW_WARNINGS>...': failed to parse warning: unknown warning: ohno

For more information, try '--help'.
```
4. Comma list:
```console
$ cargo run -- --warn "self-wakes,lost-waker,never-yielded" --allow "self-wakes,never-yielded"
     Running `target/debug/tokio-console --warn self-wakes,lost-waker,never-yielded --allow self-wakes,never-yielded`
The application panicked (crashed).
Message:  [LostWaker]
Location: tokio-console/src/main.rs:79
```

tokio-console/src/config.rs
```diff
@@ -62,11 +63,29 @@ pub struct Config {
...
+    /// * `self-wakes` -- Warns when a task wakes itself more than a certain percentage of its total wakeups.
+    ///                  Default percentage is 50%.
+    ///
+    /// * `lost-waker` -- Warns when a task is dropped without being woken.
+    ///
+    /// * `never-yielded` -- Warns when a task has never yielded.
+    ///
+    /// If this is set to `all`, all warnings are allowed.
+    ///
+    /// [possible values: all, self-wakes, lost-waker, never-yielded]
```
0xPoe, line 85: I hope to set this automatically. But because we are using `FromStr` to parse value, we can not get it.

0xPoe (commented):
🔢 Self-check

tokio-console/src/config.rs
```diff
@@ -156,6 +188,44 @@ impl KnownWarnings {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub(crate) enum AllowedWarnings {
```
hds, line 194: nit: it would be great to add a short docstring to this enum and its options so that we know what they mean.

tokio-console/src/main.rs
```diff
@@ -64,9 +65,20 @@ async fn main() -> color_eyre::Result<()> {
...
+    let warnings = match args.allow_warnings {
+        Some(allow_warnings) => match allow_warnings {
+            AllowedWarnings::All => vec![],
+            AllowedWarnings::Some(allow_warnings) => args
+                .warnings
+                .iter()
+                .filter(|lint| !allow_warnings.contains(lint))
+                .collect::<Vec<_>>(),
+        },
+        None => args.warnings.iter().collect::<Vec<_>>(),
```
hds: This can be flattened.

```suggestion
        Some(AllowedWarnings::All) => vec![],
        Some(AllowedWarnings::Some(allow_warnings)) => args
            .warnings
            .iter()
            .filter(|lint| !allow_warnings.contains(lint))
            .collect::<Vec<_>>(),
        None => args.warnings.iter().collect::<Vec<_>>(),
```

tokio-console/src/config.rs
```diff
@@ -156,6 +188,44 @@ impl KnownWarnings {
...
+    All,
+    Some(BTreeSet<KnownWarnings>),
```
hds: I think `Some` here might be confusing, because it's similar to `Option::Some`. Could we call this option `Explicit` or `Set` or something like that instead?

hds (changes requested):
This looks great, just a couple of small things.

tokio-console/src/main.rs
0xPoe: Thank you! Nice suggestion.

tokio-console/src/config.rs
0xPoe: Updated. Nice point!

tokio-console/src/config.rs
0xPoe: Added.

0xPoe (commented):
🔢 Self-check

tokio-console/src/config.rs
hawkw: Can we switch to using `clap`'s `ValueEnum` derive, so that we can have it automatically generate this?

0xPoe: ValueEnum only supports unit variants. `error: `#[derive(ValueEnum)]` only supports unit variants. Non-unit variants must be skipped`

hawkw: Ah, that makes sense. Never mind. It might be nice to add a comment on the enum stating that we can't derive `ValueEnum` for it?

hds (approved):
Looks good. Perhaps just add the comment on `AllowedWarnings` regarding `ValueEnum` that Eliza suggested and it's good to go!

tokio-console/src/config.rs
0xPoe: Good suggestion.Thanks! Added.

0xPoe:
> Perhaps just add the comment on `AllowedWarnings` regarding `ValueEnum` that Eliza suggested and it's good to go!

Added. 

hds:
> Added.
> Thanks for your review! 💚 💙 💜 💛 ❤️

Great, thank you! Would you mind adding a bit of a longer description to this MR (we'll then use that as the commit message).

Ideally 2 paragraphs: the first one explaining what the PR is trying to achieve (why are we making the change). The second paragraph should explain at a high level what is being changed.

This then becomes a really valuable resource for the future to understand why certain changes have been made. 
Thanks!

hawkw (approved):
This looks good to me! Please make sure to update the commit message the way @hds described, and then we'd love to merge this PR!

0xPoe:
> Great, thank you! Would you mind adding a bit of a longer description to this MR (we'll then use that as the commit message).
> 
> Ideally 2 paragraphs: the first one explaining what the PR is trying to achieve (why are we making the change). The second paragraph should explain at a high level what is being changed.

Added! Thanks for your remainder. I will add it to all my PRs.

PR #515, opened 2024-02-03
chore(deps): bump ratatui to 0.26.2 and crossterm to 0.27.0

joshka:
Bumps MSRV to 1.74.0

This necessitated 3 changes to the codebase:
- `Spans` was renamed to the more ergonomic `Line` type.
- `Frame` no longer requires a backend type parameter.
- `Table::new` requires a widths parameter, so we use
  `Table::default().rows(rows)` instead of `Table::new(rows)`.

Crossterm on Windows triggers events for key press as well as release
and repeat, which causes duplicate key presses. This change filters out
those events.


hawkw:
It looks like this change requires a MSRV bump, as the new `ratatui` seems not to support Rust 1.64.0: https://github.com/tokio-rs/console/actions/runs/7763271246/job/21175134601#step:8:1

joshka:
> It looks like this change requires a MSRV bump, as the new `ratatui` seems not to support Rust 1.64.0: https://github.com/tokio-rs/console/actions/runs/7763271246/job/21175134601#step:8:1

Thanks - bumped this to 1.70

hawkw:
@joshka mind rebasing this branch onto the latest `master` and resolving the lockfile conflicts? Once that's done, this should be good to merge!

hawkw (approved):

joshka:
```
1:31 PM]joshka: I notice that some tests are failing now - digging to check what's happening
[1:31 PM]Hayden (he/him): if it's the self wakes test then it's known flaky. 
[1:31 PM]Hayden (he/him): (I need to remove it)
[1:37 PM]joshka: Ah got it - yep it was that. I did't a quick cargo update on main and saw the tests fail there too (it didn't fail before then)
```

joshka:
Squashed and rebased

Cargo.lock
```diff
@@ -198,62 +256,59 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "clap"
-version = "4.1.14"
+version = "4.5.4"
```
0xPoe: https://github.com/clap-rs/clap/blob/5e4facf76fb38e04edba6cb1ab67d4f2b1ec8550/Cargo.toml#L17

I don't think this really compiles on v1.70.0.

```console
console on  fork/dev:dev [$] via 🦀 v1.70.0 
❯ cargo r                           
error: package `clap v4.5.4` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.70.0
Either upgrade to rustc 1.74 or newer, or use
cargo update -p clap@4.5.4 --precise ver
where `ver` is the latest version of `clap` supporting rustc 1.70.0

```

0xPoe (commented):
Regarding the failed test, it is failing consistently. Maybe it is not an unstable test. More like we updated some dependencies causing a change in behaviour?
Thanks for your contribution! 🤟 🖤

Cargo.lock
joshka: Yeah, you're right. Unless there's some great reason to avoid bumping the MSRV for this, it's at 1.74 now.

joshka: Bumped in https://github.com/tokio-rs/console/pull/515/commits/25e1efcb6445828eaa950e5362ae0f83a4836c13

0xPoe:
There are some proto changes:
```console
/// Each `ResourceUpdate` contains any resource data that has changed since the last
diff --git a/console-api/src/generated/rs.tokio.console.tasks.rs b/console-api/src/generated/rs.tokio.console.tasks.rs
index 41b8396..e8ee5fa 100644
--- a/console-api/src/generated/rs.tokio.console.tasks.rs
+++ b/console-api/src/generated/rs.tokio.console.tasks.rs
@@ -1,3 +1,4 @@
+// This file is @generated by prost-build.
 /// A task state update.
 ///
 /// Each `TaskUpdate` contains any task data that has changed since the last
diff --git a/console-api/src/generated/rs.tokio.console.trace.rs b/console-api/src/generated/rs.tokio.console.trace.rs
index [220](https://github.com/tokio-rs/console/actions/runs/8603263859/job/23574759021?pr=515#step:6:221)b55a..be3516e 100644
--- a/console-api/src/generated/rs.tokio.console.trace.rs
+++ b/console-api/src/generated/rs.tokio.console.trace.rs
@@ -1,3 +1,4 @@
+// This file is @generated by prost-build.
 /// Start watching trace events with the provided filter.
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
test bootstrap ... FAILED
```


Because there are some other issues in this PR, I would recommend:
1. This PR only focuses on bumping the rataui version. 
2. If the rataui requires a higher MSRV, we bump the MSRV but do not change other dependencies' versions.


Perhaps we could address the issue of the failed test that was caused by changes in other dependencies.

Also, this would help us make sure the scope of this PR is not extended. We can only focus on bumping the rataui version.

Thank you for your patch again!

tokio-console/Cargo.toml
hawkw, line 1: Please undo the unrelated formatting changes in this file.

tokio-console/README.md
```diff
@@ -57,12 +58,12 @@ to record `tokio-console` telemetry. The easiest way  to do this is [using the
 emit [`tracing`] data in a format that the console can record. For programs that
 use the [Tokio] runtime, this means that:
 
-- Tokio's [unstable features][unstable] must be enabled. See [the `console-subscriber`
+* Tokio's [unstable features][unstable] must be enabled. See [the `console-subscriber`
   documentation][unstable] for details.
-- A [compatible Tokio version][versions] must be used. Tokio v1.0 or greater is required
+* A [compatible Tokio version][versions] must be used. Tokio v1.0 or greater is required
   to use the console, and some features are only available in later versions.
```
hawkw: Can we remove these unrelated formatting changes?

tokio-console/README.md
```diff
@@ -57,12 +58,12 @@ to record `tokio-console` telemetry. The easiest way  to do this is [using the
...
   See [the `console-subscriber` documentation][versions] for details.
- 
+
```
hawkw:
```suggestion
```

tokio-console/README.md
```diff
@@ -91,13 +92,15 @@ tokio-console http://192.168.0.42:9090
 ```
 
 A DNS name can also be provided as the target address:
+
```
hawkw:
```suggestion
```

tokio-console/README.md
```diff
@@ -91,13 +92,15 @@ tokio-console http://192.168.0.42:9090
...
 ```shell
 tokio-console http://my.instrumented.application.local:6669
 ```
 
 See [here][cli-ref] for a complete list of all command-line arguments.
 
 Tokio Console has a numnber of different views:
+
```
hawkw:
```suggestion
```

xtask/Cargo.toml
```diff
@@ -3,12 +3,13 @@ name = "xtask"
...
+rust-version = "1.74.0"
 publish = false
 
 [dependencies]
-tonic-build = { version = "0.10", default-features = false, features = [
-    "prost", "transport"
+tonic-build = { version = "0.10.2", default-features = false, features = [
+    "prost",
+    "transport",
 ] }
```
hawkw: Can we undo the unrelated formatting changes here please?

xtask/Cargo.toml
```diff
@@ -3,12 +3,13 @@ name = "xtask"
...
-clap = { version = "~4.1.14", features = ["derive"] }
-color-eyre = "0.6"
+clap = { version = "4.4.18", features = ["derive"] }
+color-eyre = "0.6.2"
```
hawkw: Were these dependency updates necessary as part of the `ratatui` update?

console-subscriber/README.md
hawkw, line 1: Can we please undo the unrelated formatting changes in this file?

console-subscriber/Cargo.toml
hawkw, line 1: Please undo the unrelated formatting changes in this file.

hawkw (dismissed):
I've noticed a large number of unrelated formatting changes. I would really prefer not to merge those as part of this PR, because then, when viewing the Git history, those lines will have been changed as part of the Ratatui update, when it was not actually part of that change. I'm happy to accept formatting changes if they are considered better or more idiomatic style, but please make those changes in a separate branch.

If you're manually reformatting TOML as part of this PR, can you please put it back the way it was? Or, if your editor is doing this automatically, can you please disable that setting for this repo? Thank you!

tokio-console/Cargo.toml
joshka: I realised that the VSCode extension "Better TOML" was doing this automatically. I thought it was just a standard TOML formatting and not one that's done by a more opinionated tool. Will revert on that basis.

joshka: Done

joshka:
I've fixed the PR to just focus on the Ratatui changes (and bumped it to 0.26.2 which was released after this PR was created). I've also bumped crossterm to 0.27.0 to avoid incompatible versions, and fixed a small bug which hits most crossterm based apps on Windows (duplicate key presses).

I've bumped all the projects to MSRV 1.74 as this is the version that we build with in Ratatui. I'm not sure if it was right to just bump the tokio-console project or all the files, but this seems to be the least likely to cause CI pain (e.g. building the libs with 1.64 and the tui app with 1.74).

hds (approved):
Thank you so much for following through with this change @joshka!

This looks good to me now.

@hawkw Your change request will be blocking this change from being merged now. Would you mind having another look.

tokio-console/src/main.rs
```diff
@@ -95,6 +95,11 @@ async fn main() -> color_eyre::Result<()> {
                 let input = input
                     .ok_or_else(|| eyre!("keyboard input stream ended early"))
                     .with_section(|| "this is probably a bug".header("Note:"))??;
+
+                if input::should_ignore_key_event(&input) {
```
0xPoe, line 87: https://github.com/tokio-rs/console/blob/bb463c1b80f0cdaa639c547f397f1102da259c9c/tokio-console/src/main.rs#L81

We already have a filter here to ignore the release event. Maybe we should consider merging them together.

0xPoe (commented):
The rest looks good to me.

Thank you very much for working on this!

tokio-console/src/main.rs
joshka: I missed seeing that, fixing it by just using the should_ignore_key_event instead of the filter. Either way would be fine I guess.

0xPoe: nit: @hds @hawkw Do you have any thoughts about which one is better? It seems both work well. The only difference is that the try_filter will filter it asynchronously.

0xPoe (approved):
Looks good to me. Thanks! 👍

0xPoe:
Lint CI fixed in https://github.com/tokio-rs/console/pull/548

0xPoe:
@joshka There are some unused imports after your change. Could you please help to remove them? Thanks!

joshka:
> @joshka There are some unused imports after your change. Could you please help to remove them? Thanks!

Done

hds:
@hawkw Could you please rereview and approve (if you're happy with it) this PR? Your request for changes is blocking at the moment. Thanks!

Issue #516, opened 2024-02-07
`blocking` and `block_on` tasks have false positive lints


hds:
### What crate(s) in this repo are involved in the problem?

tokio-console

### What is the issue?

When using the `#[tokio::main]` macro, some non-async tasks are missed because the `console-subscriber` is initialized after the runtime has already started. Specifically

* `block_on` task
* `blocking` tasks for the workers in a multi-thread runtime

This hides a problem which is only visible if the `console-susbcriber` is started before the runtime. Both these types of tasks trigger false-positive warnings from lints.

* `block_on` triggers the [lost waker lint](https://github.com/tokio-rs/console/blob/b0e3196bdd25f7b191e76c52e532930212684c66/tokio-console/src/warnings.rs#L171)
* `blocking` tasks trigger the [never yielded lint](https://github.com/tokio-rs/console/blob/b0e3196bdd25f7b191e76c52e532930212684c66/tokio-console/src/warnings.rs#L219)  

### How can the bug be reproduced?

The following code reproduces both false positives:

```rust
fn main() {
    console_subscriber::init();

    let rt = tokio::runtime::Builder::new_multi_thread()
        .enable_all()
        .build()
        .unwrap();

    rt.block_on(async {
        _ = tokio::spawn(async {
            tokio::time::sleep(std::time::Duration::from_secs(6000)).await;
        })
        .await;
    });
}
```

### Logs, error output, etc

The view in Tokio Console is:

<img width="1218" alt="Screenshot 2024-02-07 at 12 33 37" src="https://github.com/tokio-rs/console/assets/89589/6e86ddbb-36f2-4629-b39f-91c81aff62cf">

### Versions

* `tokio-console` 0.1.10 (also on [`main`](https://github.com/tokio-rs/console/tree/b0e3196bdd25f7b191e76c52e532930212684c66) at the time of reporting)

### Possible solution

* Modify the lints to first check the [`kind`](https://github.com/tokio-rs/console/blob/b0e3196bdd25f7b191e76c52e532930212684c66/tokio-console/src/state/tasks.rs#L106) (added in #478) of the task being checked.

Only if the `kind` is `task` or `local` should the actual check be performed, if the `kind` is `block_on` or `blocking` then [`Warning::Ok`](https://github.com/tokio-rs/console/blob/b0e3196bdd25f7b191e76c52e532930212684c66/tokio-console/src/warnings.rs#L56) should be returned.

[...]

PR #517, opened 2024-02-07
fix(console): lints trigger with non-async tasks

javihernant:
change existing lints to only trigger with tasks that are async; those are all except those with kind 'blocking' and 'block_on'

this commit fixes issue #516

### DIFF
--- a/tokio-console/src/state/tasks.rs
+++ b/tokio-console/src/state/tasks.rs
@@ -354,6 +354,10 @@ impl Task {
         self.stats.last_wake > self.stats.last_poll_started
     }
 
+    pub(crate) fn is_blocking(&self) -> bool {
+        matches!(self.kind.as_ref(), "block_on" | "blocking")
+    }
+
     pub(crate) fn is_completed(&self) -> bool {
         self.stats.total.is_some()
     }
--- a/tokio-console/src/warnings.rs
+++ b/tokio-console/src/warnings.rs
@@ -148,6 +148,10 @@ impl Warn<Task> for SelfWakePercent {
     }
 
     fn check(&self, task: &Task) -> Warning {
+        // Don't fire warning for tasks that are not async
+        if task.is_blocking() {
+            return Warning::Ok;
+        }
         let self_wakes = task.self_wake_percent();
         if self_wakes > self.min_percent {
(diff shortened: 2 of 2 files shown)


tokio-console/src/warnings.rs
```diff
@@ -148,6 +148,10 @@ impl Warn<Task> for SelfWakePercent {
     }
 
     fn check(&self, task: &Task) -> Warning {
+        // Don't fire warning for tasks that are not async
+        if task.kind() == "block_on" || task.kind() == "blocking" {
+            return Warning::Ok
+        }
```
hawkw: nit, take it or leave it: this seems like it could potentially be factored out into a method on `Task`, so that we could instead just write:
```suggestion
        // Don't fire warning for tasks that are not async
        if task.is_blocking() {
            return Warning::Ok
        }
```

hawkw (commented):
Looks good to me overall, although I think it might be nice to factor out the repeated logic into a method. What do you think?

javihernant:
yea, good suggestion. I can do that!

javihernant:
I made the change. Please, don't hesitate to ask for any other change (nit or anything). I really want to learn, so I really appreciate any kind of suggestion! :)

tokio-console/src/state/tasks.rs
```diff
@@ -354,6 +354,10 @@ impl Task {
         self.stats.last_wake > self.stats.last_poll_started
     }
 
+    pub(crate) fn is_blocking(&self) -> bool {
+        self.kind.as_ref() == "block_on" || self.kind.as_ref() == "blocking"
```
hawkw: Take it or leave it: this could also be written using pattern matching, using the [`matches!` macro](https://doc.rust-lang.org/stable/std/macro.matches.html). So, you could also write:
```suggestion
        matches!(self.kind.as_ref() , "block_on" | "blocking")
```
which is maybe a bit more concise.

I'm fine with either version of this code, though!

hawkw (approved):
This looks good to me! I had one additional suggestion, but it's not a hard blocker --- I'm happy to merge this either way. Thank you!

hawkw:
looks lovely, thank you!

PR #520, opened 2024-02-13
chore: add FlakeHub publishing for Nix flake

hawkw:


.github/workflows/flakehub-publish-tagged.yml
```diff
@@ -0,0 +1,27 @@
+name: "Publish tags to FlakeHub"
+on:
+  push:
+    tags:
+      - "tokio-console-v[0-9]+.[0-9]+.[0-9]+*"
```
hds, line 5: Do we only publish flakes for tokio console (I know very little about nix).

hds (approved):
Looks good based on what little I know about nix!

.github/workflows/flakehub-publish-tagged.yml
hawkw: Yes, we only want to publish the binary; there's no real value in publishing a Nix flake for the Rust crates. The flake is mostly a way for Nix users to install the `tokio-console` binary.

PR #522, opened 2024-02-13
chore: switch to `cargo dist` for release publishing

hawkw:


hds (approved):
Using cargo dist is mostly about preparing binary distributions of tokio-console, right?

In that case, this looks good. I've been meaning to try release-plz for the actual release process as I think it might make that easier.

hawkw:
@hds: 
> Using cargo dist is mostly about preparing binary distributions of tokio-console, right?

Yeah, that's right; I haven't touched our existing script for updating changelogs and generating release tags. `cargo-dist` will just build lots of binaries, and also manage creating GitHub Releases (which is nice, because our current thing for that is broken...).

> In that case, this looks good. I've been meaning to try release-plz for the actual release process as I think it might make that easier.

I believe [`cargo-dist` integrates with `cargo-release`](https://opensource.axo.dev/cargo-dist/book/workspaces/cargo-release-guide.html), which I think can do some of the other release prep stuff we're currently using the bash script for. But I haven't looked into that yet. I'm not familiar with `release-plz` and I don't really know how it compares with `cargo-release`, but I'm happy with any release automation we can add that works nicely with the way we're currently publishing stuff...the `cargo-release`/`cargo-dist` integration seems potentially appealing, but I'd need to look closer at it?

Issue #512, opened 2024-01-23
Console doesn't detect self wakes from `tokio::task::yield_now()` after Tokio 1.23


hds:
(edit(2024-01-24): Updated title and description to clarfiy that it is only self wakes coming from `tokio::task::yield_now()` that are affected)

### What is the issue?

When observing an instrumented program built with Tokio 1.23 or later, Tokio Console isn't able to detect self wakes coming from [`tokio::task::yield_now()`].

This is due to tokio-rs/tokio#5223, where the self wake was changed to be deferred so that self waking tasks wouldn't starve the resource drivers.

[`tokio::task::yield_now()`]: https://docs.rs/tokio/1.35.1/tokio/task/fn.yield_now.html

### How can the bug be reproduced?

The following code should show a task with a single self wake:

```rust
    _ = tokio::spawn(async {
        tokio::task::yield_now().await;
    })
    .await;
```

If build with Tokio 1.22.0 then we see the wake counted as a self wake in Tokio Console. We can confirm this using [`ari-subscriber`] where we get the output (truncated):

```
2024-01-23T16:24:05.167259Z TRACE runtime.spawn[1]{kind=task, task.name=, task.id=18, loc.file="src/main.rs", loc.line=15, loc.col=9} new
2024-01-23T16:24:05.167786Z TRACE runtime.spawn[1]{kind=task, task.name=, task.id=18, loc.file="src/main.rs", loc.line=15, loc.col=9} enter
2024-01-23T16:24:05.167911Z TRACE runtime.spawn[1]{kind=task, task.name=, task.id=18, loc.file="src/main.rs", loc.line=15, loc.col=9} tokio::task::waker: op="waker.wake_by_ref", task.id=1
2024-01-23T16:24:05.168007Z TRACE runtime.spawn[1]{kind=task, task.name=, task.id=18, loc.file="src/main.rs", loc.line=15, loc.col=9} exit
```

Specifically there is a wake operation (`op="waker.wake_by_ref"`) inside the `runtime.spawn` span.

When updating to Tokio 1.23.0, Tokio Console no longer counts the single wake as a self wake. Again, we can confirm using [`ari-subscriber`] (truncated output):

```
2024-01-23T16:33:26.834206Z TRACE runtime.spawn[1]{kind=task, task.name=, task.id=18, loc.file="src/main.rs", loc.line=15, loc.col=9} new
[...]

PR #525, opened 2024-02-16
test(subscriber): test with custom `self_wake()` function

hds:
Part of the testing performed in the `console-subscriber` integration
tests is detecting self wakes. This relied upon the `yield_now()` from
Tokio.

However, the behavior of this function was changed in
tokio-rs/tokio#5223 and since Tokio 1.23 the wake doesn't occur in the
task that `yield_now()` is called from. This breaks the test when using
a newer version of Tokio.

This change replaces the use of `yield_now()` with a custom
`self_wake()` function that returns a future which does perform a self
wake (wakes the task from within itself before returning
`Poll::Pending`).

The same custom `self_wake()` is also included in the `app` example so
that it shows self wakes correctly.

Tokio has been updated to 1.28.2 in the lock file (the last with
compatible MSRV) so that this fix is tested.

Ref #512

### DIFF
--- a/console-subscriber/tests/support/mod.rs
+++ b/console-subscriber/tests/support/mod.rs
@@ -1,14 +1,14 @@
-use futures::Future;
+use std::{future::Future, task::Poll};
+
+use tokio::task::JoinHandle;
 
 mod state;
 mod subscriber;
 mod task;
 
 use subscriber::run_test;
-
 pub(crate) use subscriber::MAIN_TASK_NAME;
 pub(crate) use task::ExpectedTask;
-use tokio::task::JoinHandle;
 
 /// Assert that an `expected_task` is recorded by a console-subscriber
 /// when driving the provided `future` to completion.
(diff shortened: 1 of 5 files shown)


console-subscriber/tests/support/mod.rs
```diff
@@ -1,14 +1,14 @@
-use futures::Future;
+use std::{future::Future, task::Poll};
+
+use tokio::task::JoinHandle;
```
0xPoe, line 3: Maybe this change is unnecessary. But I guess it is OK.

console-subscriber/tests/support/mod.rs
```diff
@@ -62,3 +62,45 @@ where
...
+/// was improved in [tokio-rs/tokio#5223] to avoid starving the resource
+/// drivers.
+///
+/// Awaiting the future returned from this function will result in a
+/// self-wake being recorded.
+///
+/// [`yield_now()`]: fn@tokio::task::yield_now
+/// [tokio-rs/tokio#5223]: https://github.com/tokio-rs/tokio/pull/5223
+#[allow(dead_code)]
+pub(crate) fn self_wake() -> impl Future<Output = ()> {
```
0xPoe, line 82: Maybe we should move it to a separate crate. Then we don't need to duplicate it in the examples. But I think it is OK to have this kind of small duplication.

0xPoe (approved):
Looks good to me. Thanks! :shipit:

console-subscriber/tests/support/mod.rs
hds: The previous one was probably incorrect, we should be using `Future` from the std library.

console-subscriber/tests/support/mod.rs
hds: I thought about that, but I didn't want to introduce additional complexity for the example script - also, this means that it can be copy-pasted elsewhere without non-published dependencies.

console-subscriber/tests/support/mod.rs
0xPoe: I mean we don't need to move `use tokio::task::JoinHandle;`. But I guess it is totally OK.

hds:
@hawkw would you be able to have a look at this one when you get a moment? Thanks!!

PR #527, opened 2024-02-17
feat: Adjust task and waker panel for long names

devanbenz:
close Friendly ping~ Do you have time to work on it again?

- Makes better use of waker and task panel spacing
- Puts Last woken on its own line
- wraps text within the task panel for really long location names

Updated:
task_long_name
task_no_name

### DIFF
--- a/tokio-console/src/view/task.rs
+++ b/tokio-console/src/view/task.rs
@@ -9,6 +9,7 @@ use crate::{
         help::HelpText,
     },
 };
+use ratatui::widgets::Wrap;
 use ratatui::{
     layout::{self, Layout},
     text::{Span, Spans, Text},
@@ -131,8 +132,8 @@ impl TaskView {
             .direction(layout::Direction::Horizontal)
             .constraints(
                 [
-                    layout::Constraint::Percentage(50),
-                    layout::Constraint::Percentage(50),
+                    layout::Constraint::Percentage(60),
+                    layout::Constraint::Percentage(40),
                 ]
                 .as_ref(),
             )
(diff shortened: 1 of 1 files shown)


hds (changes requested):
I haven't had a chance to look to much into the code yet. But from the screenshots, it looks like you're losing lines from the Task panel (I see the scheduled and idle times missing from the bottom).

I'm not sure if you've noticed this, but the task panel will need to have a variable height, whereas right now the height is fixed.

Also, I think that we probably don't need to put the location on its own line (after the label), it should be able to start on the line with the label and then break onto additional lines if it needs to.

I'll come back with a complete review (hopefully tomorrow) once I've had a chance to read the code changes.

devanbenz:
Ah! Yes, good call. I did not notice that. I’ll need to make some
adjustments

tokio-console/src/view/task.rs
```diff
@@ -9,6 +9,7 @@ use crate::{
         help::HelpText,
     },
 };
+use ratatui::widgets::Wrap;
```
hds, line 12: style: merge this with the other ratatui imports ok the next line.

tokio-console/src/view/task.rs
```diff
@@ -131,8 +132,8 @@ impl TaskView {
             .direction(layout::Direction::Horizontal)
             .constraints(
                 [
-                    layout::Constraint::Percentage(50),
-                    layout::Constraint::Percentage(50),
+                    layout::Constraint::Percentage(60),
```
hds, line 135: I think we might want something more dynamic than this. What do you think?

hds (commented):
A few more comments based on reading the changes this time.

devanbenz:
Will be coming back to this PR this weekend! Sorry I was out of town last week/over the weekend :D

0xPoe:
@devanbenz Friendly ping~ Do you have time to work on it again? Thanks!

devanbenz:
> @devanbenz Friendly ping~ Do you have time to work on it again? Thanks!

I definitely want to get back to this. I have an interview this week and a talk during the weekend. I have a feeling that afterwards I will likely be able to start contributing to open source more 👍

hds:
This PR has been superceded by #614, so I'll close it out.

PR #533, opened 2024-03-07
refactor: match span id earlier

0xPoe:
Just a simple refactoring to match and get span ID first. Then we don't need to check it twice.

Also, align logs to use `id`.


hds (approved):
Looks good. Thanks!

PR #535, opened 2024-03-28
refac(console): match span id earlier

0xPoe:
Follow the same pattern applied in #533 to match the span id
earlier to avoid checking twice.

### DIFF
--- a/tokio-console/src/state/async_ops.rs
+++ b/tokio-console/src/state/async_ops.rs
@@ -141,14 +141,17 @@ impl AsyncOpsState {
 
         self.async_ops
             .insert_with(visibility, update.new_async_ops, |ids, async_op| {
-                if async_op.id.is_none() {
-                    tracing::warn!(?async_op, "skipping async op with no id");
-                }
-
+                let span_id = match async_op.id.as_ref() {
+                    Some(id) => id.id,
+                    None => {
+                        tracing::warn!(?async_op, "skipping async op with no id");
+                        return None;
+                    }
+                };
                 let meta_id = match async_op.metadata.as_ref() {
                     Some(id) => id.id,
                     None => {
-                        tracing::warn!(?async_op, "async op has no metadata ID, skipping");
+                        tracing::warn!(?async_op, "async op has no metadata id, skipping");
                         return None;
                     }
                 };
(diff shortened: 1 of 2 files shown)


tokio-console/src/view/mod.rs
```diff
@@ -69,6 +69,7 @@ pub(crate) enum UpdateKind {
     SelectTask(u64),
     /// The TaskView is exited
     ExitTaskView,
+    #[allow(dead_code)]
```
0xPoe: We might use this in the future. So I just allow the dead code check. Please let me know if we want to remove it.

0xPoe (commented):

tokio-console/src/view/mod.rs
hds: I'm not sure why this isn't used but `SelectTask(u64)` is.

hds (changes requested):
This looks good.

Would you mind splitting the change to make clippy happy into another PR though? It's best to not mix concerns.

tokio-console/src/view/mod.rs
0xPoe: I think we used the task ID to monitor task details, but we don't need to use it for resources.

0xPoe:
> Would you mind splitting the change to make clippy happy into another PR though? It's best to not mix concerns.

Dropped. I will open a new PR for it.

0xPoe:
> > Would you mind splitting the change to make clippy happy into another PR though? It's best to not mix concerns.
> 
> Dropped. I will open a new PR for it.

See: https://github.com/tokio-rs/console/pull/536

hds (approved):
This looks good. Thank you very much!

Could you please make sure that the commit message follows our styling for the `console` repo. This is the basis for our release notes, so it's important. We follow [conventional commits], but checking the summary in the **Commit message guidelines** section of the [CONTRIBUTING.md] document is probably enough.

In this case, the commit message should be something like (both `refac` and `refactor` are accepted, see https://github.com/tokio-rs/console/blob/main/cliff.toml#L80):

```
refac(console): match span id earlier (#535)

Follow the same pattern applied in #533 to match the span id
earlier to avoid checking twice.
```

[conventional commits]: https://www.conventionalcommits.org/en/v1.0.0/
[CONTRIBUTING.md]: https://github.com/tokio-rs/console/blob/main/CONTRIBUTING.md#commit-message-guidelines

0xPoe:
@hds Updated. Thanks! I didn't know that we also needed to include a detailed commit body. I thought that the Pull Request body was enough, although my PR body wasn't very good either. I'll make sure to add it next time.

PR #540, opened 2024-04-09
chore: all use `.yml` for workflows

0xPoe:
We use both `.yaml` and `.yml` for workflows. It might better to use the same extension.


hds (approved):
Good idea. Thank you!

PR #541, opened 2024-04-09
test: add cases for format_location

0xPoe:
Because this function uses a regex, it's better to add some tests to make sure we don't break anything if we want to change the regex.

### DIFF
--- a/tokio-console/src/state/mod.rs
+++ b/tokio-console/src/state/mod.rs
@@ -549,7 +549,7 @@ mod tests {
     // Therefore, it's important to ensure that paths, which can differ between operating systems,
     // are displayed correctly in the console.
     #[test]
-    fn test_format_location() {
+    fn test_format_location_linux() {
         // Linux style paths.
         let location1 = proto::Location {
             file: Some(
@@ -580,6 +580,11 @@ mod tests {
             "/home/user/projects/tokio-1.0.1/src/lib.rs"
         );
 
+        assert_eq!(format_location(None), "<unknown location>");
+    }
+
+    #[test]
+    fn test_format_location_macos() {
         // macOS style paths.
         let location4 = proto::Location { file: Some("/Users/user/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.0.1/src/lib.rs".to_string()), ..Default::default() };
         let location5 = proto::Location {
@@ -603,7 +608,5 @@ mod tests {
             format_location(Some(location6)),
             "/Users/user/projects/tokio-1.0.1/src/lib.rs"
         );
-
-        assert_eq!(format_location(None), "<unknown location>");
     }
 }
--- a/tokio-console/src/state/mod.rs
+++ b/tokio-console/src/state/mod.rs
@@ -543,18 +543,29 @@ fn pb_duration(dur: prost_types::Duration) -> Duration {
 mod tests {
     use super::*;
 
+    // This test should be run on all platforms. The console can display instrumentation data
+    // from different console subscribers that may be running on different operating systems.
+    // For instance, the console could be running on Windows, while the application is running on Linux.
+    // Therefore, it's important to ensure that paths, which can differ between operating systems,
+    // are displayed correctly in the console.
     #[test]
-    #[cfg(target_os = "linux")]
     fn test_format_location() {
-        let mut location1 = proto::Location::default();
-        location1.file = Some(
-            "/home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.0.1/src/lib.rs"
-                .to_string(),
-        );
-        let mut location2 = proto::Location::default();
-        location2.file = Some("/home/user/.cargo/git/checkouts/tokio-1.0.1/src/lib.rs".to_string());
-        let mut location3 = proto::Location::default();
-        location3.file = Some("/home/user/projects/tokio-1.0.1/src/lib.rs".to_string());
+        // Linux style paths.
+        let location1 = proto::Location {
+            file: Some(
+                "/home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.0.1/src/lib.rs"
+                    .to_string(),
+            ),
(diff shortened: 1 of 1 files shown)


tokio-console/src/state/mod.rs
```diff
@@ -538,3 +538,35 @@ fn pb_duration(dur: prost_types::Duration) -> Duration {
     let nanos = u64::try_from(dur.nanos).expect("duration should not be negative!");
     Duration::from_secs(secs) + Duration::from_nanos(nanos)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_format_location() {
```
hds: This is a great idea. Could we also add tests for typical macOS (starting with `/Users/<username>`) and Windows (actually I don't know what this looks like) paths? Probably in separate test cases (separate `#[test]`).

hds (changes requested):
This is a great idea, I think it makes sense to test paths from other platform too.

tokio-console/src/state/mod.rs
0xPoe: Good idea! I guess our current regex won't work for the Windows path. I'll test it on a Windows chip.

0xPoe: image
Actually, it works well on Windows. I will figure out how we handle it on Windows.

0xPoe: I looked into the issue and found that the solution doesn't work for Windows paths. Although the output seems correct for this 'app' example, it actually isn't. The problem arises because when we use the `spawn` function, it uses `std::panic::Location::caller` to get the file location. I printed the location we received from the console-subscriber on Windows, which is `console-subscriber\\examples\\app.rs:68:10`. Unfortunately, this doesn't match the regex, so we end up calling `toString()`, which results in `console-subscriber\examples\app.rs:68:10`. 

To fix this problem, I will test more cases on Windows and improve the regex to support Windows path. But I don't think we should do it in this PR. I will open a new PR to fix it.
@hds What do you think? Does that make sense?

hds: Yep, that makes sense. Let's do it that way!

hds (approved):
Looks good. Thanks!

tokio-console/src/state/mod.rs
```diff
@@ -538,3 +538,64 @@ fn pb_duration(dur: prost_types::Duration) -> Duration {
...
+        );
+        assert_eq!(
+            format_location(Some(location3)),
+            "/home/user/projects/tokio-1.0.1/src/lib.rs"
+        );
+        assert_eq!(format_location(None), "<unknown location>");
+    }
+
+    #[test]
+    #[cfg(target_os = "macos")]
```
hawkw: is it actually necessary to only compile these tests on the OS whose paths they test? since we don't use `std::path::Path` to construct or parse the paths, i think the behavior should be the same on all operating systems — we're just testing how the console handles strings that _look_ like the paths that are produced by those operating systems. i'd prefer it if all the tests were always run, regardless of what platform a developer runs them on. 

also, it's worth noting that these paths are generated when the instrumented application is compiled, not from the system that's running the console. so, for example, if a console running on a Windows dev machine connects to an application compiled on Linux, it will see Linux-style paths, not Windows ones. therefore, i think we ought to be testing against all paths, regardless of the OS we're compiling the console for.

tokio-console/src/state/mod.rs
```diff
@@ -538,3 +538,64 @@ fn pb_duration(dur: prost_types::Duration) -> Duration {
...
+    use super::*;
+
+    #[test]
+    #[cfg(target_os = "linux")]
+    fn test_format_location() {
+        let mut location1 = proto::Location::default();
+        location1.file = Some(
+            "/home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.0.1/src/lib.rs"
+                .to_string(),
+        );
```
hawkw: style nit, take it or leave it: i think we could also express this as
```suggestion
        let location1 = proto::Location {
            file: Some(
                "/home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.0.1/src/lib.rs"
                    .to_string(),
            ),
            ..Default::default()
        };
```
which avoids the `mut` — it's entirely up to you whether you think this style is nicer, though.

hawkw (changes requested):
if possible, i would like to remove the conditional compilation from these tests. besides that, this is great! thank you!

tokio-console/src/state/mod.rs
0xPoe: Yeah. Sounds make sense.

tokio-console/src/state/mod.rs
0xPoe: Good suggestion!

0xPoe (commented):

tokio-console/src/state/mod.rs
```diff
@@ -538,3 +538,72 @@ fn pb_duration(dur: prost_types::Duration) -> Duration {
...
+        assert_eq!(
+            format_location(Some(location2)),
+            "<cargo>/tokio-1.0.1/src/lib.rs"
+        );
+        assert_eq!(
+            format_location(Some(location3)),
+            "/home/user/projects/tokio-1.0.1/src/lib.rs"
+        );
+
+        // macOS style paths.
```
hawkw: this isn't a big deal, but my preference would still be to have separate tests for the macOS and Linux paths — that way, if one test fails, the test output will still indicate whether the other succeeded, which is potentially useful for determining if a bug effects all path styles, or just one OS. 
```suggestion
    }

    #[test]
    fn test_format_location_macos() {
        // macOS style paths.
```

hawkw (approved):
one last suggestion, although it's not a blocker — i'm fine with the current code, but i think separating the tests might be a little nicer.

tokio-console/src/state/mod.rs
0xPoe: Updated.

PR #539, opened 2024-04-08
chore(deps): bump trycmd to 0.15.1

joshka:
Fixes the race condition in remove_dir all as it is no longer a
dependency. See https://github.com/advisories/GHSA-mc8h-8q98-g5hr

Before:
```
❯ cargo tree -i remove_dir_all
remove_dir_all v0.5.3
└── tempfile v3.3.0
    ├── prost-build v0.12.0
    │   └── tonic-build v0.10.0
    │       └── xtask v0.1.0 (/Users/joshka/local/tokio-console/xtask)
    │       [dev-dependencies]
    │       └── console-api v0.6.0 (/Users/joshka/local/tokio-console/console-api)
    │           ├── console-subscriber v0.2.0 (/Users/joshka/local/tokio-console/console-subscriber)
    │           └── tokio-console v0.1.10 (/Users/joshka/local/tokio-console/tokio-console)
    │   [dev-dependencies]
    │   └── console-api v0.6.0 (/Users/joshka/local/tokio-console/console-api) (*)
    └── snapbox v0.5.9
        └── trycmd v0.15.1
            [dev-dependencies]
            └── tokio-console v0.1.10 (/Users/joshka/local/tokio-console/tokio-console)
```
After:
```
❯ cargo tree -i remove_dir_all
error: package ID specification `remove_dir_all` did not match any packages
```

PR #542, opened 2024-04-09
chore: get rid of remove_dir_all

0xPoe:
close https://github.com/tokio-rs/console/pull/539

I did two things:
1. [chore: cargo update -p rustix --precise 0.38.32](https://github.com/tokio-rs/console/pull/542/commits/44d1ca11267ec5cfaa74346d8fac4e0316009d3a)
2. [chore: cargo update -p tempfile --precise 3.10.1](https://github.com/tokio-rs/console/pull/542/commits/1bfb515023bd7e89e137d22f73f746b18a063746)

See https://github.com/advisories/GHSA-mc8h-8q98-g5hr

```console
console on  rustin-patch-remove_dir-with-1.70.0 [$!] via 🦀 v1.78.0 took 21s 
❯ cargo tree -i remove_dir_all
error: package ID specification `remove_dir_all` did not match any packages
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -587,12 +587,9 @@ dependencies = [
 
 [[package]]
 name = "fastrand"
-version = "1.7.0"
+version = "2.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c3fcf0cee53519c866c09b5de1f6c56ff9d647101f81c1964fa632e148896cdf"
-dependencies = [
- "instant",
-]
+checksum = "9fc0510504f03c51ada170672ac806f1f105a88aa97a5281117e1ddc3368e51a"
 
 [[package]]
 name = "filetime"
@@ -955,15 +952,6 @@ name = "indoc"
 version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b248f5224d1d606005e02c97f5aa4e88eeb230488bcc03bc9ca4d7991399f2b5"
-
-[[package]]
-name = "instant"
-version = "0.1.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7a5bbe824c507c5da5956355e86a746d82e0e1464f65d862cc5e71da70e94b2c"
-dependencies = [
- "cfg-if",
-]
 
 [[package]]
 name = "is_terminal_polyfill"
@@ -1520,15 +1508,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dbb5fb1acd8a1a18b3dd5be62d25485eb770e05afb408a9627d14d451bae12da"
 
 [[package]]
-name = "remove_dir_all"
-version = "0.5.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3acd125665422973a33ac9d3dd2df85edad0f4ae9b00dafb1a05e43a9f5ef8e7"
-dependencies = [
- "winapi",
-]
-
-[[package]]
 name = "rustc-demangle"
 version = "0.1.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1778,16 +1757,14 @@ checksum = "20518fe4a4c9acf048008599e464deb21beeae3d3578418951a189c235a7a9a8"
 
 [[package]]
 name = "tempfile"
-version = "3.3.0"
+version = "3.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5cdb1ef4eaeeaddc8fbd371e5017057064af0911902ef36b39801f67cc6d79e4"
+checksum = "85b77fafb263dd9d05cbeac119526425676db3784113aa9295c88498cbf8bff1"
 dependencies = [
  "cfg-if",
  "fastrand",
- "libc",
- "redox_syscall 0.2.13",
- "remove_dir_all",
- "winapi",
+ "rustix",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


console-api/Cargo.toml
```diff
@@ -3,8 +3,8 @@ name = "console-api"
 version = "0.6.0"
 license = "MIT"
 edition = "2021"
-rust-version = "1.64.0"
-authors = ["Eliza Weisman <eliza@buoyant.io>", "Tokio Contributors <team@tokio.rs>",]
+rust-version = "1.70.0"
```
hawkw: why was it necessary to do a MSRV bump in this PR?

console-api/Cargo.toml
```diff
@@ -3,8 +3,8 @@ name = "console-api"
...
+authors = ["Eliza Weisman <eliza@buoyant.io>", "Tokio Contributors <team@tokio.rs>", ]
```
hawkw: why the added whitespace here?

hawkw (commented):
It would be nice to avoid the MSRV bump --- is the `clap` dependency update necessary?

console-api/Cargo.toml
0xPoe: This was added by my IDE. I'll drop it.

console-api/Cargo.toml
0xPoe: Because clap v4.4.18 set the 1.70.0 as the MSRV.

0xPoe:
> It would be nice to avoid the MSRV bump --- is the `clap` dependency update necessary?

As you can tell from the dependency tree:
```console
❯ cargo tree -i remove_dir_all
remove_dir_all v0.5.3
└── tempfile v3.3.0
    ├── prost-build v0.12.0
    │   └── tonic-build v0.10.0
    │       └── xtask v0.1.0 (/Users/joshka/local/tokio-console/xtask)
    │       [dev-dependencies]
    │       └── console-api v0.6.0 (/Users/joshka/local/tokio-console/console-api)
    │           ├── console-subscriber v0.2.0 (/Users/joshka/local/tokio-console/console-subscriber)
    │           └── tokio-console v0.1.10 (/Users/joshka/local/tokio-console/tokio-console)
    │   [dev-dependencies]
    │   └── console-api v0.6.0 (/Users/joshka/local/tokio-console/console-api) (*)
    └── snapbox v0.5.9
        └── trycmd v0.15.1
            [dev-dependencies]
            └── tokio-console v0.1.10 (/Users/joshka/local/tokio-console/tokio-console)
```

If we want to get rid of the `remove_dir_all` then we need to bump the `tempfile`. But if we try to bump it directly, we will get an error:
```console
❯ cargo update -p tempfile --precise 3.10.1
    Updating crates.io index
error: failed to select a version for `rustix`.
    ... required by package `tempfile v3.10.1`
    ... which satisfies dependency `tempfile = "^3"` of package `prost-build v0.12.0`
    ... which satisfies dependency `prost-build = "^0.12.0"` (locked to 0.12.0) of package `console-api v0.6.0 (/Volumes/t7/code/console/console-api)`
    ... which satisfies path dependency `console-api` (locked to 0.6.0) of package `console-subscriber v0.2.0 (/Volumes/t7/code/console/console-subscriber)`
versions that meet the requirements `^0.38.31` are: 0.38.32, 0.38.31

all possible versions conflict with previously selected packages.

  previously selected package `rustix v0.38.15`
    ... which satisfies dependency `rustix = "^0.38.0"` (locked to 0.38.15) of package `is-terminal v0.4.9`
    ... which satisfies dependency `is-terminal = "^0.4.1"` (locked to 0.4.9) of package `clap_builder v4.1.14`
    ... which satisfies dependency `clap_builder = "=4.1.14"` (locked to 4.1.14) of package `clap v4.1.14`
    ... which satisfies dependency `clap = "~4.1.14"` (locked to 4.1.14) of package `tokio-console v0.1.10 (/Volumes/t7/code/console/tokio-console)`

failed to select a version for `rustix` which could resolve this conflict
```

So before we bump the `tempfile`, we have to bump the clap to use the higher version `rustix`.

0xPoe (commented):

console-api/Cargo.toml
hds: @hawkw we're already going to bump the MSRV to 1.74 in #515 (which fixes an issue with input on Windows).

@hi-rustin I propose that we focus on getting the ratatui PR in and then this PR won't need an MSRV bump at all. That will avoid having multiple bumps in close changes.

0xPoe: I don't think 515 requires bumping the MSRV to 1.74.0. It also requires 1.70.0. See https://github.com/ratatui-org/ratatui/blob/efd1e476425477b0bce15e3dd96d5cdeb0e1174b/Cargo.toml#L21

But I am OK to wait for that PR to land.

Cargo.lock
```diff
@@ -1500,15 +1507,6 @@ version = "0.7.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dbb5fb1acd8a1a18b3dd5be62d25485eb770e05afb408a9627d14d451bae12da"
 
-[[package]]
-name = "remove_dir_all"
```
0xPoe, line 1534: Removed!

0xPoe (commented):

hds (changes requested):
There seem to be a lot of new transient dependencies added as part of this change.

Is it possible to make the change without doing this?

Or perhaps we could have a "chore: cargo update" PR first so that we don't get these things mixed up. What do you think?

0xPoe:
> There seem to be a lot of new transient dependencies added as part of this change.
> 
> Is it possible to make the change without doing this?

You can refer to https://github.com/tokio-rs/console/pull/542#issuecomment-2047507319

But I think we can split them into two PRs. But the change would be the same.

0xPoe:
> But I think we can split them into two PRs. But the change would be the same.

Split: https://github.com/tokio-rs/console/pull/552

0xPoe (commented):

hds (approved):
Great, thank you!

0xPoe:

PR #544, opened 2024-04-10
test(subscriber): remove very flaky tests

hds:
We have identified that some of the `console-subscriber` integration
tests are flaky (#473). This appears to be a result of the following
issue in `tracing` tokio-rs/tracing#2743.

Flaky tests are really worse than no tests. So these tests will be
removed until the flakiness can be fixed.

### DIFF
--- a/console-subscriber/tests/wake.rs
+++ b/console-subscriber/tests/wake.rs
@@ -5,35 +5,6 @@ use support::{assert_task, ExpectedTask};
 use tokio::time::sleep;
 
 #[test]
-fn sleep_wakes() {
-    let expected_task = ExpectedTask::default()
-        .match_default_name()
-        .expect_wakes(1)
-        .expect_self_wakes(0);
-
-    let future = async {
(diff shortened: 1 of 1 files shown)


console-subscriber/tests/wake.rs
```diff
@@ -4,35 +4,6 @@ use std::time::Duration;
 use support::{assert_task, ExpectedTask};
 use tokio::time::sleep;
 
-#[test]
```
0xPoe, line 7: How about we keep the tests but just ignore them by the `#[ignore]` macro? In the future, we can add it back easily.

0xPoe (approved):
Looks good to me. Thanks! 👍

I'll let you make the decision on removing it or ignoring it.

console-subscriber/tests/wake.rs
hds: I'd prefer to just remove. We can use #473 to track the need to put them back.

PR #546, opened 2024-04-15
chore: run clippy against all targets

0xPoe:
This would help us to check if there is any clippy warning in the tests.

I found it after https://github.com/tokio-rs/console/pull/544. There are some unused imports in our test code.

Use `cargo clippy --workspace --all-targets --no-deps -- -D warnings` to test all targes which include the test targets.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: ["main"]
+    branches: [ "main" ]
   pull_request:
 
 name: CI
@@ -48,8 +48,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        rust: [stable]
+        os: [ ubuntu-latest, macos-latest, windows-latest ]
+        rust: [ stable ]
         include:
           # Make 1.64 MSRV, as it's Tonic 0.10's MSRV.
           - rust: 1.64.0
@@ -100,7 +100,7 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: Run cargo clippy
-        run: cargo clippy -- -D warnings
+        run: cargo clippy --workspace --all-targets --no-deps -- -D warnings
 
   grpc_web:
     name: gRPC-web Example
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [ "main" ]
+    branches: ["main"]
   pull_request:
 
 name: CI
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: ["main"]
+    branches: [ "main" ]
```
hds: Could we avoid the unrelated formatting changes? I think this might be your editor, is that right?

hds (changes requested):
Good idea! Would be good to avoid the unrelated formatting changes in the GitHub Actions file though.

console-subscriber/tests/spawn.rs
```diff
@@ -10,6 +10,7 @@ use support::{assert_tasks, spawn_named, ExpectedTask};
 /// spawned the child task). In this scenario, that would result in the parent
 /// having 3 polls counted, when it should really be 1.
 #[test]
+#[allow(clippy::async_yields_async)]
```
hawkw, line 15: in general, it's nice if allow attributes for lints can have a comment explaining why they're allowed in this instance. it would be nice if you're able to add one here, although it's not a huge deal.

console-subscriber/tests/support/task.rs
```diff
@@ -122,6 +110,7 @@ impl ExpectedTask {
 
     /// Returns an error specifying that no match was found for this expected
     /// task.
+    #[allow(clippy::result_large_err)]
```
hawkw: nit, take out or leave it: maybe we should allow this globally for the whole `impl` block?

hawkw (commented):
overall, this looks good, thanks for the lint fixes! i agree with @hds that we should undo the YAML formatting changes before merging.

console-subscriber/tests/spawn.rs
0xPoe: Good suggestion! Added.

0xPoe (commented):

console-subscriber/tests/support/task.rs
0xPoe: Make sense. Moved.

.github/workflows/ci.yml
0xPoe: It was reformated by my IDE. Those changes were dropped.

hawkw (approved):
looks good to me, thank you!

0xPoe:
@hds This PR needs your approval before I can merge it. Could you please take a look? Thanks!
image

hds (approved):
Thanks for adding this, great idea!

hds:
> @hds This PR needs your approval before I can merge it. Could you please take a look? Thanks!
> 
> image
> 
> 

Sorry! I didn't realize I was blocking this one. 🤦‍♀️

0xPoe:
> Sorry! I didn't realize I was blocking this one. 🤦‍♀️

No worry! Thank you ❤️

PR #547, opened 2024-04-19
chore(deps): bump tonic to 0.11 and raise MSRV to 1.70

matze:


console-api/Cargo.toml
```diff
@@ -40,7 +40,7 @@ tracing-core = "0.1.17"
 futures-core = "0.3"
 
 [dev-dependencies]
-tonic-build = { version = "0.10", default-features = false, features = [
+tonic-build = { version = "0.11", default-features = false, features = [
```
hds, line 43: When updating `tonic-build`, the `console-api` interface needs to be updated. That makes this a breaking change for both `console-api` and `console-subscriber` which both include tonic in the public API.

hds (changes requested):
By bumping to a semver incompatible version of `tonic`, this change needs to be marked as incompatible in the commit message (https://www.conventionalcommits.org/en/v1.0.0/).

Also, the `console-api` protobuf files need to be updated, as do the ones `console-subscriber` grpc web example.

For `console-api` the following command needs to be run:

```sh
cargo run -p xtask -- gen-proto
```

For the example, the instructions are here:
https://github.com/tokio-rs/console/tree/main/console-subscriber/examples/grpc_web#understanding-the-code

console-api/Cargo.toml
matze: Amended the commit but
```
cargo run -p xtask -- gen-proto
```
didn't do anything ~~and I'd be surprised it should because prost is still the same and the protos left untouched~~. Seems tonic-build still outputs exactly the same.

hds: OK, it may be that this change doesn't affect the output, but it is still a breaking change for `console-api` and `console-subscriber` and we'll have to treat it as such.

matze: As I said, I amended the commit with the breaking change trailer. Anything else that's missing?

hds: Sorry, I missed that. This is good to go. However, we've got another change which is going to bump the MSRV higher, I'd prefer to put that in first and then this one won't need the MSRV bump. Hopefully we'll get that other one in soon.

Xuanwo:
Hello, I came across this PR because `console-api` and `console-subscriber` are blocking our project's upgrade to `tonic 0.11`. Is this PR ready for merge?

0xPoe (approved):
Looks good to me. Thanks! 👍

Blocked by https://github.com/tokio-rs/console/pull/515

0xPoe:
@hds Could you please help approve this PR? Then we can merge it. Thanks!

hds (approved):
Thanks for adding this one!

PR #549, opened 2024-05-09
chore: rename config to config.toml

0xPoe:
Cargo recommended using `config.toml` as the config file name.

Before change:
```console
❯ cargo b     
warning: `/Volumes/t7/code/console/.cargo/config` is deprecated in favor of `config.toml`
note: if you need to support cargo 1.38 or earlier, you can symlink `config` to `config.toml`
```


hds (approved):
Thanks for fixing this!

PR #552, opened 2024-05-16
chore: bump clap and clap_complete to the latest version

0xPoe:
This PR split from https://github.com/tokio-rs/console/pull/542.

We need to bump the clap first. Then we can bump the tempfile crate. See https://github.com/tokio-rs/console/pull/542#issuecomment-2047507319


hds (approved):
Thanks for splitting this out, I think that's clearer now!

Issue #553, opened 2024-05-23
assertion failed: meta.is_event()


dinhani:
### What crate(s) in this repo are involved in the problem?

console-api

### What is the issue?

`console-api` is panicking when running in debug mode in this [`debug_assert!`](https://github.com/tokio-rs/console/blob/60bcf87537d414b21efbd84159207f9ecda5e914/console-api/src/common.rs#L48)

No idea if this is expected or not, but I identified that an event of kind `HINT` is reaching that point and triggering the assertion to fail.

It comes from the external crate `jsonrpsee-core`, so I cannot easily change it.

```rust
Metadata { 
  name: "enabled /home/renato/.cargo/registry/src/index.crates.io-6f17d22bba15001f/jsonrpsee-core-0.22.5/src/tracing.rs:13", 
  target: "jsonrpsee_core::tracing::client", 
  level: Level(Trace), 
  module_path: "jsonrpsee_core::tracing::client", 
  location: /home/renato/.cargo/registry/src/index.crates.io-6f17d22bba15001f/jsonrpsee-core-0.22.5/src/tracing.rs:13, 
  fields: {}, 
  callsite: Identifier(0x559831eadc48), 
  kind: Kind(HINT) 
}
```

### How can the bug be reproduced?

```rust
let client = HttpClientBuilder::default().build(jsonrpc_url).unwrap();
let _ = client.request::<bool, Vec<()>>("net_listening", vec![]).await
```

### Logs, error output, etc

```text
thread 'tokio' panicked at /home/renato/projects/console/console-api/src/common.rs:50:13:
assertion failed: meta.is_event()
```

### Versions

```text
├── console-subscriber v0.2.0
│   ├── console-api v0.6.0
```

### Possible solution

Maybe filtering out invalid metadata kinds before calling the function or using `TryFrom` instead.
The function is being called from [aggregator.rs](https://github.com/tokio-rs/console/blob/main/console-subscriber/src/aggregator/mod.rs#L449-L450) in `console-subscriber`.

### Would you like to work on fixing this bug?

yes

hds:
@dinhani Thanks for your issue report!

I would say that this is definitely incorrect on `console-api` side, even in debug mode we should be more resilient.

[...]

PR #554, opened 2024-05-24
fix(subscriber): ignore metadata that is not a span or event

dinhani:
As described in #553, there are parts of code that expects tracing `Metadata` to be of type `EVENT` or `SPAN`.

This PR ensure that any tracing metadata with a different type will be ignored instead of panicking when running in debug mode like [here](https://github.com/tokio-rs/console/blob/60bcf87537d414b21efbd84159207f9ecda5e914/console-api/src/common.rs#L48).


hds (approved):
This looks great, thank you! I'm going to go ahead and merge without any tests because some substantial changes are needed to the test structure to catch this one properly, and I don't want to hold it up.

hds:
@dinhani Merged. Thank you very much for the report and the PR!

PR #555, opened 2024-05-30
fix: handle Windows path correctly

### LINKED ISSUE #550: tokio-console cannot handle window path correctly

tokio-console

### What is the issue?

image

Although the output seems correct for this 'app' example, it actually isn't. The problem arises because when we use the spawn function, it uses std::panic::Location::caller to get the file location. I printed the location we received from the console-subscriber on Windows, which is console-subscriber\\examples\\app.rs:68:10. Unfortunately, this doesn't match the regex, so we end up calling toString(), which results in console-subscriber\examples\app.rs:68:10.

Run tokio-console on the Windows.

### Logs, error output, etc

_No response_

### Versions

```text
cargo run -- -V
   Compiling tokio-console v0.1.10 (/Volumes/t7/code/console/tokio-console)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.07s
     Running `target/debug/tokio-console -V`
tokio-console 0.1.10
```

### Possible solution

To fix this problem, I will test more cases on Windows and improve the regex to support Windows path. But I don't think we should do it in this PR. I will open a new PR to fix it.

### Additional context

ref https://github.com/tokio-rs/console/pull/541#discussion_r1565637248

yes

0xPoe:
fix https://github.com/tokio-rs/console/issues/550

We need to make truncate_registry_path can handle Windows path as well. Because tokio-console can connect to any server from different platforms, so we use the same path separator to have the same experience.

Some test snapshots:
local package: 
image

crates.io package: 
image

### DIFF
--- a/tokio-console/src/state/mod.rs
+++ b/tokio-console/src/state/mod.rs
@@ -511,10 +511,8 @@ fn truncate_registry_path(s: String) -> String {
 
     static REGEX: OnceCell<Regex> = OnceCell::new();
     let regex = REGEX.get_or_init(|| {
-        Regex::new(
-            r".*(/|\\)\.cargo(/|\\)(registry(/|\\)src(/|\\)[^/\\]*(/|\\)|git(/|\\)checkouts(/|\\))",
-        )
-        .expect("failed to compile regex")
+        Regex::new(r".*[/\\]\.cargo[/\\](registry[/\\]src[/\\][^/\\]*[/\\]|git[/\\]checkouts[/\\])")
+            .expect("failed to compile regex")
     });
 
     let s = match regex.replace(&s, "<cargo>/") {
--- a/tokio-console/src/state/mod.rs
+++ b/tokio-console/src/state/mod.rs
@@ -511,15 +511,20 @@ fn truncate_registry_path(s: String) -> String {
 
     static REGEX: OnceCell<Regex> = OnceCell::new();
     let regex = REGEX.get_or_init(|| {
-        Regex::new(r".*/\.cargo(/registry/src/[^/]*/|/git/checkouts/)")
-            .expect("failed to compile regex")
+        Regex::new(
+            r".*(/|\\)\.cargo(/|\\)(registry(/|\\)src(/|\\)[^/\\]*(/|\\)|git(/|\\)checkouts(/|\\))",
+        )
+        .expect("failed to compile regex")
     });
 
-    return match regex.replace(&s, "<cargo>/") {
+    let s = match regex.replace(&s, "<cargo>/") {
         Cow::Owned(s) => s,
         // String was not modified, return the original.
-        Cow::Borrowed(_) => s.to_string(),
+        Cow::Borrowed(_) => s,
     };
+
+    // This help use the same path separator on all platforms.
+    s.replace('\\', "/")
 }
 
 fn format_location(loc: Option<proto::Location>) -> String {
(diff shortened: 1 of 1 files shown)


0xPoe (commented):

hds:
I haven't had a chance to look at the change properly yet (although what I have checked looks good). However I'm not sure what the image in the PR description is showing. Should the "local" one also show a crate path, rather than one from the application code?

0xPoe:
> Should the "local" one also show a crate path, rather than one from the application code?

Because I ran the example from the console project directly. So I guess we have to show the application code. This path does not match the regex, so we directly show it.

tokio-console/src/state/mod.rs
```diff
@@ -511,15 +511,20 @@ fn truncate_registry_path(s: String) -> String {
 
     static REGEX: OnceCell<Regex> = OnceCell::new();
     let regex = REGEX.get_or_init(|| {
-        Regex::new(r".*/\.cargo(/registry/src/[^/]*/|/git/checkouts/)")
-            .expect("failed to compile regex")
+        Regex::new(
+            r".*(/|\\)\.cargo(/|\\)(registry(/|\\)src(/|\\)[^/\\]*(/|\\)|git(/|\\)checkouts(/|\\))",
```
hds: It might make more sense to use character sets `[/\\]` instead of groups `(/|\\)` for the path separators. When I see groups I expect that we're extracting the contents.

That would make it like this:

```suggestion
            r".*[/\\]\.cargo[/\\](registry[/\\]src[/\\][^/\\]*[/\\]|git[/\\]checkouts[/\\])",
```

I think this regex is also getting **almost** too complex. I'd leave it for now, but if it becomes more so, I'd consider replacing it with either multiple regex or splitting the path into parts and then checking that way.

tokio-console/src/state/mod.rs
```diff
@@ -511,15 +511,20 @@ fn truncate_registry_path(s: String) -> String {
...
-    return match regex.replace(&s, "<cargo>/") {
+    let s = match regex.replace(&s, "<cargo>/") {
         Cow::Owned(s) => s,
         // String was not modified, return the original.
-        Cow::Borrowed(_) => s.to_string(),
+        Cow::Borrowed(_) => s,
     };
+
+    // This help use the same path separator on all platforms.
+    s.replace('\\', "/")
```
hds: I'm not sure this is a good idea, it might cause issues if people have a `/` or `\` in some directory name (although I'm not sure we should worry about that case really).

Also, while `"C:/Users/user/projects/tokio-1.0.1/src/lib.rs"` looks fine to me, maybe it looks weird to a Windows user?

0xPoe:
I am oncall this week and next week. I will update it when I get out of jail(oncall).

tokio-console/src/state/mod.rs
0xPoe: Good suggestion! Updated!

tokio-console/src/state/mod.rs
0xPoe: > Also, while `"C:/Users/user/projects/tokio-1.0.1/src/lib.rs"` looks fine to me, maybe it looks weird to a Windows user?

The reason I want to use the same path style is that users can use tokio-console to connect to servers from different servers. So maybe using the same path style would help us reduce confusion. But I also agree that it might cause some confusion for Windows users.

0xPoe: > I'm not sure this is a good idea, it might cause issues if people have a `/` or `\` in some directory name (although I'm not sure we should worry about that case really).

Yes. This is possible. At least, on macOS, I can create a dir that contains it by command: `mkdir test\\a`.

0xPoe: How about:
1. `<cargo>/tokio-1.0.1/src/lib.rs` for macOS and Linux path.
2. `<cargo>\tokio-1.0.1\src\lib.rs` for Windows path.

But I am not sure if there is an easy way to detect if the path string is from the Windows OS.

0xPoe: @hds Do you have any thoughts on this? I am not sure which way we should go.

hds: > How about:
> 
>     1. `<cargo>/tokio-1.0.1/src/lib.rs` for macOS and Linux path.
> 
>     2. `<cargo>\tokio-1.0.1\src\lib.rs` for Windows path.
> 
> 
> But I am not sure if there is an easy way to detect if the path string is from the Windows OS.

Let's go with this. I would suggest we just go with what separator we see in the path. If it's `/` we leave it as `/`, if it's `\` we leave it `\`. Naively you could probably count the number of each character in the string and only use `\` (which I would consider less likely) if it is stricly higher.

0xPoe: I used:
```rust
fn is_windows_path(path: &str) -> bool {
    use once_cell::sync::OnceCell;
    use regex::Regex;

    static REGEX: OnceCell<Regex> = OnceCell::new();
    let regex = REGEX.get_or_init(|| Regex::new(r"^[a-zA-Z]:\\").expect("failed to compile regex"));
    let has_drive_letter = regex.is_match(path);
    let slash_count = path.find('/').iter().count();
    let backslash_count = path.find('\\').iter().count();
    has_drive_letter && backslash_count > slash_count
}
```

It is not perfect, but it seems to work.

hds (approved):
Looks good. Thank you!

0xPoe:
Tested locally again:
image

PR #559, opened 2024-06-07
chore: include tonic updates in release notes

hds:
In the past we have marked commits which update tonic as `chore`, which
excludes them from the release notes. However, when we update tonic to a
new (pre-1.0) major version, such as from 0.10 to 0.11, this is a
breaking change for both the `console-api` and the `console-subscriber`
crates. As such, it shouldn't be skipped.

This change adds a new commit type `update`, to use for updating
dependencies. It will have it's own section in the release notes. A
commit preprocessing rule has been added to the Git-Cliff config file to
change commit subjects from `chore: bump tonic` to `update: bump tonic`
so that they're caught in the new category (this will fix it for the
next release which includes upgrading tonic from 0.10 to 0.11).

The contributor guide has also been updated to mention this new commit
type.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -290,6 +290,8 @@ Must be one of the following:
 * **feat**: A new feature
 * **fix**: A bug fix
 * **docs**: Documentation only changes
+* **update**: A dependency update (when a core dependency, especially
+  `tonic`, gets a breaking update)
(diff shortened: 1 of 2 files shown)


0xPoe (approved):
Looks good to me. Thanks! 👍

PR #562, opened 2024-06-20
fix: tidy Cargo.lock

0xPoe:
tokio-console would be 0.1.11.

Do we have any bugs in our release process?


hds (approved):
Nice catch!

0xPoe:

PR #567, opened 2024-07-04
feat: add temporality state to update message

### LINKED ISSUE #551: tokio-console cannot show the pause status correctly

tokio-console

### What is the issue?

There are two Tokio Console instances connected to the same stream. If you pause the stream in one instance and the status cannot be displayed correctly in the other instance.

All steps execute under the git repo:

1. Start the example app: `cargo run --example app`
2. Start one tokio console: `cargo run`
3. Start another tokio console: `cargo run`
4. Hit `space` in one of the instances
image
As you can see, all of them are paused, but only one of them showed `PAUSED`.

### Logs, error output, etc

_No response_

### Versions

```text
commit 3193fde77adf1473c4124adfa6ef8234659589c9 (origin/main, origin/HEAD)
Author: 二手掉包工程师 <rustin.liu@gmail.com>
Date:   Mon Apr 22 21:35:21 2024 +0800

    chore: run clippy against all targets (#546)
    
    This would help us to check if there are any
    clippy warning in the tests.
```

### Possible solution

Perhaps we should include a new API to allow clients to check whether the stream has been paused.

### Additional context

I believe the reason for this problem is that we only store the pause status on the client side. Therefore, there is no way to notify other instances to know the stream is paused.

yes

0xPoe:
close https://github.com/tokio-rs/console/issues/551

This PR added a new field to the console's update message. Clients can use it to check if the subscriber stream is paused.

For example:
image

### DIFF
--- a/console-api/proto/instrument.proto
+++ b/console-api/proto/instrument.proto
@@ -70,6 +70,17 @@ message Update {
 
     // Any new span metadata that was registered since the last update.
     common.RegisterMetadata new_metadata = 5;
+
+    // The time "state" of the aggregator, such as paused or live.
+    Temporality temporality = 6;
+}
+
+// The time "state" of the aggregator.
+enum Temporality {
+    // The aggregator is currently live.
+    LIVE = 0;
+    // The aggregator is currently paused.
+    PAUSED = 1;
 }
 
 // `PauseResponse` is the value returned after a pause request.
--- a/tokio-console/src/state/mod.rs
+++ b/tokio-console/src/state/mod.rs
@@ -5,6 +5,7 @@ use crate::{
     warnings::Linter,
 };
 use console_api as proto;
+use console_api::instrument::Temporality;
 use ratatui::{
     style::{Color, Modifier},
     text::Span,
@@ -71,12 +72,6 @@ pub(crate) enum FieldValue {
     Debug(String),
 }
 
-#[derive(Debug)]
-enum Temporality {
-    Live,
-    Paused,
-}
-
 #[derive(Debug, Eq, PartialEq)]
 pub(crate) struct Attribute {
     field: Field,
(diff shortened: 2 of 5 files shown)


hi-rustin:
Still WIP. I just wanted to let you know that I'm not ready for review.

hi-rustin:
Tested locally:
image

0xPoe (commented):
This is not a perfect solution to fully solve this problem.

It can only be used before the new client starts, for already running clients, it doesn't help, because there is no message sent to that second client.

tokio-console/src/state/mod.rs
```diff
@@ -5,6 +5,7 @@ use crate::{
     warnings::Linter,
 };
 use console_api as proto;
+use console_api::instrument::Temporality;
```
hds, line 8: For all other items used from the `console_api` crate, we reference by the "full" path (but starting with `proto`), I think we should do the same here to not make `Temporality` different.

```suggestion
```

tokio-console/src/state/mod.rs
```diff
@@ -107,6 +102,14 @@ impl State {
         current_view: &view::ViewState,
         update: proto::instrument::Update,
     ) {
+        match Temporality::try_from(update.temporality) {
```
hds, line 105: Perhaps best to stick to the way that we reference all other types from the `console_api` crate.
```suggestion
        match proto::instrument::Temporality::try_from(update.temporality) {
```

console-api/proto/instrument.proto
```diff
@@ -70,6 +70,17 @@ message Update {
 
     // Any new span metadata that was registered since the last update.
     common.RegisterMetadata new_metadata = 5;
+
+    // The time "state" of the aggregator, such as paused or live.
+    Temporality temporality = 6;
+}
+
+// The time "state" of the aggregator.
+enum Temporality {
```
hds, line 79: I think it would make sense to add another `rpc` for console subscriber state updates. Right now temporality would be the only thing in there, but this would allow us to do "out of band" state updates such as this case, where no data is being sent.

This would also allow us to solve the issue with already connected clients, as we could send this new server state update without needing to send a real update, so all connected tokio console instances would get the pause update.

What do you think?

hds (changes requested):
I think that putting the pause/live state into a separate update stream would make it flexible enough to update all clients that the subscriber is paused.

console-api/proto/instrument.proto
0xPoe: Sounds good! It would be a good solution.

0xPoe:
Finally, I have some time to work on it again. I will try to add a new API for it.

PR #570, opened 2024-07-08
fix(console): avoid crash when accessing selected item

### LINKED ISSUE #565: Crash when selecting a task that no longer exists
I reliably encounter the crash below while running tokio console.
To reproduce:
- Have tasks ordered in descending order.
- Select the last task with 'G'
- Wait for the task to disappear (of course, it means you need to instrument a binary with a varying number of tasks).
- Push enter.
- The crash below occurs.

## Error
```
index out of bounds: the len is 58 but the index is 18446744073709551601
```

## Metadata
|key|value|
|--|--|
|**version**|0.1.10|
|**config.subcmd**|`None`|
|**config.target_addr**|`None`|
|**config.env_filter**|`None`|
|**config.log_directory**|`None`|
|**config.retain_for**|`None`|
|**config.view_options.no_colors**|`false`|
|**config.view_options.lang**|`Some("en_US.UTF-8")`|
|**config.view_options.ascii_only**|`None`|
|**config.view_options.truecolor**|`Some(true)`|
|**config.view_options.palette**|`None`|
|**config.view_options.toggles.color_durations**|`None`|
|**config.view_options.toggles.color_terminated**|`None`|
|**location**|/home/raphael/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-console-0.1.10/src/view/table.rs:166:34|

0xPoe:
We should check the length before using the index to access it.

close https://github.com/tokio-rs/console/issues/565

Test locally:

https://github.com/tokio-rs/console/assets/29879298/5c4fd5da-e1c7-490b-bd67-1257972076d3

But it is difficult to view it, you can try it by following the steps from the issue.

### DIFF
--- a/tokio-console/src/view/mod.rs
+++ b/tokio-console/src/view/mod.rs
@@ -132,14 +132,12 @@ impl View {
                 // mutate the currently selected view.
                 match event {
                     key!(Enter) => {
-                        if let Some(task_weak) = self.tasks_list.selected_item() {
-                            if let Some(task) = task_weak.upgrade() {
-                                update_kind = UpdateKind::SelectTask(task.borrow().span_id());
-                                self.state = TaskInstance(self::task::TaskView::new(
-                                    task,
-                                    state.task_details_ref(),
-                                ));
-                            }
+                        if let Some(task) = self.tasks_list.selected_item() {
+                            update_kind = UpdateKind::SelectTask(task.borrow().span_id());
+                            self.state = TaskInstance(self::task::TaskView::new(
+                                task,
+                                state.task_details_ref(),
+                            ));
                         }
                     }
                     _ => {
@@ -151,12 +149,9 @@ impl View {
             ResourcesList => {
                 match event {
                     key!(Enter) => {
-                        if let Some(res_weak) = self.resources_list.selected_item() {
-                            if let Some(res) = res_weak.upgrade() {
-                                update_kind = UpdateKind::SelectResource(res.borrow().span_id());
-                                self.state =
-                                    ResourceInstance(self::resource::ResourceView::new(res));
-                            }
+                        if let Some(res) = self.resources_list.selected_item() {
+                            update_kind = UpdateKind::SelectResource(res.borrow().span_id());
+                            self.state = ResourceInstance(self::resource::ResourceView::new(res));
                         }
                     }
                     _ => {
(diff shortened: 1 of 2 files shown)


tokio-console/src/view/table.rs
```diff
@@ -154,18 +154,19 @@ impl<T: TableList<N>, const N: usize> TableListState<T, N> {
...
-            })
-            .unwrap_or_default()
+                    None
+                }
+            } else {
+                let adjusted_index = self.sorted_items.len().checked_sub(i + 1)?;
+                self.sorted_items.get(adjusted_index).cloned()
+            }
+        })
     }
```
hds: Given that we always call `selected_item` and then have to upgrade the the `Weak` reference, how about we perform the upgrade in `selected_item`, since we need to return an `Option` now anyway.

So it would look like:

```rust
    pub(in crate::view) fn selected_item(&self) -> Option<Rc<RefCell<T::Row>>> {
        self.table_state.selected().and_then(|i| {
            if self.sort_descending {
                if i < self.sorted_items.len() {
                    Some(self.sorted_items[i].clone())
                } else {
                    None
                }
            } else {
                let adjusted_index = self.sorted_items.len().checked_sub(i + 1)?;
                self.sorted_items.get(adjusted_index).cloned()
            }
        })
        .and_then(|weak| weak.upgrade())
    }
```

Alternatively we could use the same pattern at each call site (see other comment).

tokio-console/src/view/mod.rs
```diff
@@ -169,21 +174,24 @@ impl View {
...
-                        if let Some(op) = view.async_ops_table.selected_item().upgrade() {
-                            if let Some(task_id) = op.borrow().task_id() {
-                                let task = self
-                                    .tasks_list
-                                    .sorted_items
-                                    .iter()
-                                    .filter_map(|i| i.upgrade())
-                                    .find(|t| task_id == t.borrow().id());
+                        if let Some(op_weak) = view.async_ops_table.selected_item() {
+                            if let Some(op) = op_weak.upgrade() {
```
hds: As an alternative approach to changing `selected_item()` we could use `.and_then` here:

```suggestion
                        if let Some(op_weak) = view.async_ops_table.selected_item().and_then(|w| w.upgrade()) {
```

This removes the extra level of indentation.

hds (changes requested):
I think we can avoid that extra level of indentation one way or another.

tokio-console/src/view/table.rs
0xPoe: I picked this approach. Thank you!

0xPoe (commented):

hawkw (approved):
Thanks, this looks good to me!

0xPoe:
@hds Need your approval to merge it. Could you please take another look? Thank you!

hds (approved):
Looks good. Thank you!

0xPoe:

PR #571, opened 2024-07-10
update(api,subscriber)!: upgrade tonic to 0.12

dnaka91:
Update all dependencies on `tonic` and `tonic-build` to v0.12, along with the related `prost`, `prost-types` and `prost-build` crates to v0.13.1.


hds (changes requested):
This all looks good from my side. Thanks for putting in the work to do this!

The exception is that we need to treat this change as breaking (hence me changing the PR title).

I'm marking it as "request changes" only because I want to merge it personally a little later and make sure that the commit message correctly marks the change as breaking.

dnaka91:
I followed along #547 for how to mark it as a breaking release but it seems things have changed a bit since then.

If there's anything I can do from my side, please let me know and I'll update the PR 👍.

hds:
> I followed along #547 for how to mark it as a breaking release but it seems things have changed a bit since then.
> 
> If there's anything I can do from my side, please let me know and I'll update the PR 👍.

Actually I hadn't even seen your commit message. It's pretty much perfect, thank you!

There is a small difference in that I want to list the crates that actually have breaking changes in the scope part of the conventional commits. This doesn't do anything at the moment, but I'm hoping we can get a change to Release-plz to support that (see MarcoIeni/release-plz#1514).

The other thing is that "bump" sounds like a small change, where this is a breaking tonic change, so I've used "upgrade" as the verb instead in the PR and I'll change that in the commit message when merging.

hds (approved):
Thanks!

hds:
@dnaka91 Merged. Thank you once again for your PR and for taking the time to write a descriptive commit message that follows our conventions. We really appreciate that!

PR #573, opened 2024-07-10
refactor: use dep syntax to avoid crate rename

dnaka91:
Use the `dep:` syntax to explicitly define `parking_lot` as dependency in the feature definition, added in Rust 1.60. This allows to avoid the rename of `parking_lot` to `parking_lot_crate` internally.


hawkw (approved):
Great, thank you!

0xPoe (approved):
Looks good to me. Thanks! 👍

PR #578, opened 2024-07-29
fix(subscriber): remove unused `AggregatorHandle` and fix other lints

hds:
A number of new or updated Clippy lints in Rust 1.80.0 need to be fixed.

An update to the `dead_code` pointed out that the `AggregatorHandle` is
not used, and it is not constructable from outside the crate because it
has a private field. This struct was introduced in #451 as part of the
`Server::into_parts` method. Originally, this method was going to return
the `AggregatorHandle`, which wrapped the join handle from the task
where the `Aggregator` had been spawned. This was later replaced by
returning the `Aggregator` itself, which the user had the obligation to
spawn themselves. However, it seems that the `AggregatorHandle` wasn't
removed, even though it was never used.

A new lint is the one for unexpected `--cfg` items. We now need to
declare those in `Cargo.toml`.

An update to `needless_borrows_for_generic_args` causes a false positive
changing a `&mut` to a move, which we can't do as the same value is used
afterwards.

### DIFF
--- a/console-subscriber/README.md
+++ b/console-subscriber/README.md
@@ -94,12 +94,12 @@ runtime][Tokio] is considered *experimental*. In order to use
   level].
 
   + If you're using the [`console_subscriber::init()`][init] or
-  [`console_subscriber::Builder`][builder] APIs, these targets are enabled
-  automatically.
+    [`console_subscriber::Builder`][builder] APIs, these targets are enabled
+    automatically.
 
   + If you are manually configuring the `tracing` subscriber using the
-  [`EnvFilter`] or [`Targets`] filters from [`tracing-subscriber`], add
-  `"tokio=trace,runtime=trace"` to your filter configuration.
+    [`EnvFilter`] or [`Targets`] filters from [`tracing-subscriber`], add
+    `"tokio=trace,runtime=trace"` to your filter configuration.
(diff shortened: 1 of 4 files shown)


hds:
@hi-rustin Could you please have a look at these changes to fix new and updated clippy lints form Rust 1.80.0.

console-subscriber/README.md
```diff
@@ -94,12 +94,12 @@ runtime][Tokio] is considered *experimental*. In order to use
   level].
 
   + If you're using the [`console_subscriber::init()`][init] or
-  [`console_subscriber::Builder`][builder] APIs, these targets are enabled
-  automatically.
+    [`console_subscriber::Builder`][builder] APIs, these targets are enabled
```
0xPoe, line 97: Not sure why this was changed.

0xPoe (approved):

console-subscriber/README.md
hds: This is a new clippy lint. 2nd and subsequent lines of a list item should be indented by an extra 2 spaces so they match the text (after the "bullet") of the first line.

0xPoe: I see, It came from the doc lint.

PR #579, opened 2024-08-19
fix(console): correct the grammar issue

0xPoe:
I know the best solution is to ensure we can distinguish singular or plural here. But for now, let's correct the grammar issue first.


hds (approved):
Makes sense. Thanks!

0xPoe:

PR #582, opened 2024-09-05
feat: add the WatchState API

### LINKED ISSUE #551: tokio-console cannot show the pause status correctly

0xPoe:
close https://github.com/tokio-rs/console/issues/551
close https://github.com/tokio-rs/console/pull/567

This pull request introduces a new stream API named "WatchState." Currently, it only provides the current temporality, which will enable clients to check if the server has been paused.

See more: https://github.com/tokio-rs/console/pull/567#discussion_r1694635301

https://github.com/user-attachments/assets/51bfb1cc-3f72-4acd-94bc-6cbefa729963

### DIFF
--- a/tokio-console/src/conn.rs
+++ b/tokio-console/src/conn.rs
@@ -37,6 +37,7 @@ enum State {
     Disconnected(Duration),
 }
 
+#[allow(clippy::large_enum_variant)]
 pub(crate) enum Message {
     Update(Update),
     State(console_api::instrument::State),
--- a/tokio-console/src/conn.rs
+++ b/tokio-console/src/conn.rs
@@ -1,3 +1,4 @@
+use console_api::instrument::StateRequest;
 use console_api::instrument::{
     instrument_client::InstrumentClient, InstrumentRequest, PauseRequest, ResumeRequest,
     TaskDetailsRequest, Update,
@@ -30,9 +31,15 @@ pub struct Connection {
 enum State {
     Connected {
         client: InstrumentClient<Channel>,
-        stream: Box<Streaming<Update>>,
+        update_stream: Box<Streaming<Update>>,
+        state_stream: Box<Streaming<console_api::instrument::State>>,
     },
     Disconnected(Duration),
+}
+
+pub(crate) enum Message {
+    Update(Update),
+    State(console_api::instrument::State),
 }
 
 macro_rules! with_client {
@@ -110,9 +117,16 @@ impl Connection {
                     }
                 };
                 let mut client = InstrumentClient::new(channel);
-                let request = tonic::Request::new(InstrumentRequest {});
-                let stream = Box::new(client.watch_updates(request).await?.into_inner());
-                Ok::<State, Box<dyn Error + Send + Sync>>(State::Connected { client, stream })
+                let update_request = tonic::Request::new(InstrumentRequest {});
+                let update_stream =
+                    Box::new(client.watch_updates(update_request).await?.into_inner());
+                let state_request = tonic::Request::new(StateRequest {});
+                let state_stream = Box::new(client.watch_state(state_request).await?.into_inner());
+                Ok::<State, Box<dyn Error + Send + Sync>>(State::Connected {
+                    client,
+                    update_stream,
+                    state_stream,
+                })
             };
             self.state = match try_connect.await {
                 Ok(connected) => {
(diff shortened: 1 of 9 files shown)


0xPoe:
@hds Do you have time to take a quick look at this pull request before I continue? I want to ensure I'm on the right track and understand your feedback correctly.

Do you have any thoughts on how to use it with `tokio-console`? Currently, we use `next_update` to connect to the server and handle all the backoff stuff. I'm not sure how to use this new API in `tokio-console`. I'm unsure which step we should take first when we're trying to start the client. Perhaps we should watch the stats first? Or should we only watch the stats after we receive the first update?

hds:
@Rustin170506 I think this is the right approach.

To integrate this with Tokio-Console, I would get the `State` stream from `WatchState` and include it in the `tokio::select!` statement in `main()`:

https://github.com/tokio-rs/console/blob/f8e1bee760358f702ca8359ec3de6cb39649fe60/tokio-console/src/main.rs#L81

To make the UI "reactive" even in the event of network slow down, we may want to have some additional states in Tokio Console for "pausing" and "unpausing" which would be set in between the user pressing the space bar and the new state being received from the `State` stream.

Does that make sense?

0xPoe:
> To integrate this with Tokio-Console, I would get the `State` stream from `WatchState` and include it in the `tokio::select!` statement in `main()`:

The issue at hand is that when we call `next_event`, it attempts to connect to the server within the function. However, since we've introduced a new API, we're unsure about when we should monitor this new API. Should we have `next_event` establish the connection first, or should we update the logic to monitor the state API and establish the connection simultaneously?

hds:
My thought was to do it all simultaneously.

0xPoe (commented):

0xPoe:
@hds, I finally have some time to fix it. Could you please take a look? Thank you!

tokio-console/src/state/mod.rs
```diff
@@ -71,12 +71,6 @@ pub(crate) enum FieldValue {
     Debug(String),
 }
 
-#[derive(Debug)]
-enum Temporality {
```
hds, line 75: I think we might want to keep this internal state separate from what we receive from the `console-subscriber`, and add 2 new values for the time between the user pressing `<space>` and receiving the state update:

* Pausing
* Unpausing

For example, if we are in state `Live` and the user presses `<space>` we would transition to state `Pausing` and send the pause command to the `console-subscriber`, once we get a state update back from the `console-subscriber` that it has actually paused, we would transition to `Pause` (this bit you've already done).

This will make the UI more reactive on slower connections. My use case here is something like connecting to an application on a remote machine via an SSH pipe or something like that.

tokio-console/src/conn.rs
```diff
@@ -128,20 +143,39 @@ impl Connection {
...
-                        tracing::warn!(%status, "error from stream");
-                        self.state = State::Disconnected(Self::BACKOFF);
+            match &mut self.state {
+                State::Connected {
+                    update_stream,
+                    state_stream,
+                    ..
+                } => {
+                    tokio::select! { biased; // Always biased to update stream.
+                        update = update_stream.next() => match update {
```
hds, line 155: Is `StreamExt::next` cancel safe? I didn't see the answer immediately in the documentation.

hds (changes requested):
Everything is looking great!

I've got one suggestion regarding the user experience and a question regarding cancel safety.

tokio-console/src/conn.rs
0xPoe: Yes. I think it is.
https://docs.rs/tokio/latest/tokio/macro.select.html#cancellation-safety

image

tokio-console/src/state/mod.rs
0xPoe: Sounds good. I will add it.

0xPoe: Added.

0xPoe (commented):

hds (approved):
Looks great. Thank you!

0xPoe:
Rebaed to the latest master.

0xPoe:

PR #587, opened 2024-10-01
feat(console): add large future lints

hds:
In Tokio, the futures for tasks are stored on the stack unless they are
explicitly boxed. Having very large futures can be problematic as it can
cause a stack overflow.

This change makes use of new instrumentation in Tokio
(tokio-rs/tokio#6881) which includes the size of the future which drives
a task. The size isn't given its own column (we're running out of
horizontal space) and appears in the additional fields column. In the
case that a future was auto-boxed by Tokio, the original size of the
task will also be provided.

Two new lints have been added for large futures. The first will detect
auto-boxed futures and warn about them. The second will detect futures
which are large (over 1024 bytes by default), but have not been
auto-boxed by the runtime.

Since the new lints depend on the new instrumentation in Tokio, they
will only trigger if the instrumented application is running using
`tokio` 1.41.0 or later. The version is as yet, unreleased.

Both lints have been added to the default list.

## PR Notes

Here's a screenshot with the new lints triggered on tasks from the `app` example.

large_future_lints

### DIFF
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -63,6 +63,9 @@ pub struct Config {
     /// * `lost-waker` -- Warns when a task is dropped without being woken.
     ///
     /// * `never-yielded` -- Warns when a task has never yielded.
+    ///
+    /// * `large-future` -- Warnings when the future driving a task occupies a large amount of
+    ///                     stack space.
     #[clap(long = "warn", short = 'W', value_delimiter = ',', num_args = 1..)]
     #[clap(default_values_t = KnownWarnings::default_enabled_warnings())]
     pub(crate) warnings: Vec<KnownWarnings>,
@@ -80,9 +83,12 @@ pub struct Config {
     ///
     /// * `never-yielded` -- Warns when a task has never yielded.
     ///
+    /// * `large-future` -- Warnings when the future driving a task occupies a large amount of
+    ///                     stack space.
+    ///
     /// If this is set to `all`, all warnings are allowed.
     ///
-    /// [possible values: all, self-wakes, lost-waker, never-yielded]
+    /// [possible values: all, self-wakes, lost-waker, never-yielded, large-future]
     #[clap(long = "allow", short = 'A', num_args = 1..)]
     pub(crate) allow_warnings: Option<AllowedWarnings>,
 
@@ -143,6 +149,8 @@ pub(crate) enum KnownWarnings {
     SelfWakes,
     LostWaker,
     NeverYielded,
+    AutoBoxedFuture,
+    LargeFuture,
 }
 
 impl FromStr for KnownWarnings {
@@ -153,6 +161,8 @@ impl FromStr for KnownWarnings {
             "self-wakes" => Ok(KnownWarnings::SelfWakes),
             "lost-waker" => Ok(KnownWarnings::LostWaker),
             "never-yielded" => Ok(KnownWarnings::NeverYielded),
+            "auto-boxed-future" => Ok(KnownWarnings::AutoBoxedFuture),
+            "large-future" => Ok(KnownWarnings::LargeFuture),
             _ => Err(format!("unknown warning: {}", s)),
         }
     }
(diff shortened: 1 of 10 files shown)


hds:
I'll keep this in draft until the Tokio PR is merged, as this feature is useless without it.

hds:
The dependent change in Tokio tokio-rs/tokio#6881 has been merged, so this is good to go.

tokio-console/src/warnings.rs
```diff
@@ -258,3 +258,90 @@ impl Warn<Task> for NeverYielded {
...
+pub(crate) struct LargeFuture {
+    min_size: usize,
+    description: String,
+}
+impl LargeFuture {
+    pub(crate) const DEFAULT_MIN_SIZE_BYTES: usize = 1024;
+    pub(crate) fn new(min_size: usize) -> Self {
+        Self {
+            min_size,
+            description: format!("tasks are large (threshold {} bytes)", min_size),
```
hawkw: nitpicky: what about 
```suggestion
            description: format!("tasks are {} bytes or larger", min_size),
```

tokio-console/src/warnings.rs
```diff
@@ -258,3 +258,90 @@ impl Warn<Task> for NeverYielded {
...
+
+    fn format(&self, task: &Task) -> String {
+        let original_size = task
+            .original_size_bytes()
+            .expect("warning should not trigger if original size is None");
+        let boxed_size = task
+            .size_bytes()
+            .expect("warning should not trigger if size is None");
+        format!(
+            "This task was auto-boxed by the runtime due to its size (originally \
```
hawkw: what do you think about adding a little more detail here:
```suggestion
            "This task's future was auto-boxed by the runtime when spawning, due to its size (originally \
```

tokio-console/src/config.rs
```diff
@@ -63,6 +63,9 @@ pub struct Config {
     /// * `lost-waker` -- Warns when a task is dropped without being woken.
     ///
     /// * `never-yielded` -- Warns when a task has never yielded.
+    ///
+    /// * `large-future` -- Warnings when the future driving a task occupies a large amount of
+    ///                     stack space.
```
hawkw, line 71: It looks like the `auto-boxed-future` warning is missing from the docs here?

tokio-console/src/config.rs
```diff
@@ -80,9 +83,12 @@ pub struct Config {
     ///
     /// * `never-yielded` -- Warns when a task has never yielded.
     ///
+    /// * `large-future` -- Warnings when the future driving a task occupies a large amount of
+    ///                     stack space.
+    ///
```
hawkw, line 94: Again, `auto-boxed-future` is missing.

tokio-console/src/config.rs
```diff
@@ -80,9 +83,12 @@ pub struct Config {
...
     /// If this is set to `all`, all warnings are allowed.
     ///
-    /// [possible values: all, self-wakes, lost-waker, never-yielded]
+    /// [possible values: all, self-wakes, lost-waker, never-yielded, large-future]
```
hawkw:
```suggestion
    /// [possible values: all, self-wakes, lost-waker, never-yielded, large-future, auto-boxed-future]
```

hawkw (changes requested):
This is really cool --- I love seeing new lints I hadn't thought of being added to the console!

I had some very small suggestions about the wording of the lint text, and noticed some stuff was missing from the docs. Other than that, the implementation looks great!

tokio-console/src/config.rs
hds: Fixed.

tokio-console/src/config.rs
hds: Fixed.

hawkw (commented):
Oh, one last thing: since this change depends on new Tokio instrumentation, can we add something stating which version of Tokio these lints will be available in to the list of Tokio versions here: https://github.com/tokio-rs/console/tree/main/console-subscriber#required-tokio-versions

Also, I assume you've tested that the console still works fine with prior tokio versions that are missing the task-size fields? It seems like it ought to, but it's probably worth making sure...

hds:
@hawkw Good point about the Tokio version! I had tested an earlier version against the current version of Tokio, but it was good that I tested it again, after modifying the large blocking task I had accidentally made it so large that in the Tokio 1.40.0 (and lower) it was causing a stack overflow (just in the example application, but all the same) because that blocking task wasn't being auto-boxed.

Fixed that now and confirmed that everything works fine in the current version of Tokio, just that there is no size information and so the lints won't ever trigger.

hawkw (approved):
looks good to me whenever CI is satisfied!

PR #590, opened 2024-10-21
chore: update dependencies

hds:
For users who build tokio-console by installing it with `cargo`, it is
common to use the `--locked` flag to indicate that the dependency
versions from the lock file should be used. So it's worth having those
dependencies up to date.

This change checks in the result of running `cargo update` without
changing the manifest (`Cargo.toml`) files.

The patch (pre-1.0 minor) version change to `tonic-build` included a
deprecation (rename) and some clippy lint allow directives in the
generated Rust files.

The update to `tracing` in the lockfile from 0.1.37 to 0.1.40 includes
the changes from tokio-rs/tracing#2562. This change causes an
instrumented future to be entered one final time when the future is
dropped, so that any code run in the drop implementation will be in the
scope of the span. This causes the number of polls recorded for a task
to be incremented by 1 right before it is dropped.

Due to this change, some of the `console-subscriber` tests needed to be
updated. This behaviour was already present for many application using
the `console-subscriber`, but not yet in the tests because we hadn't
updated the lockfile for a while.


0xPoe (approved):
Looks good to me. Thanks! 👍

0xPoe:
@hds The CI failed. Need to update the proto stuff.

PR #593, opened 2024-11-04
fix(api): bump minimum version of tonic

divergentdave:
Change the minimum version of tonic from 0.12 to 0.12.3. This fixes compilation in downstream projects with old lock files, since the generated code now uses a constant only present in tonic 0.12.3.

Closes #592

Issue #592, opened 2024-11-04
Minimum version of tonic dependency is too low


divergentdave:
### What crate(s) in this repo are involved in the problem?

console-api

### What is the issue?

FYI, version 0.8.1 of console-api uses the `tonic::Status::GRPC_STATUS` constant, which was introduced in tonic v0.12.3, but its dependency on tonic only specifies 0.12.

### How can the bug be reproduced?

Update `console-api` or `console-subscriber` in a project with an old Cargo.lock.

### Logs, error output, etc

```
    Checking console-api v0.8.1
error[E0599]: no associated item named `GRPC_STATUS` found for struct `Status` in the current scope
   --> /home/david/.cargo/registry/src/index.crates.io-6f17d22bba15001f/console-api-0.8.1/src/generated/rs.tokio.console.instrument.rs:577:48
    |
577 | ...                   tonic::Status::GRPC_STATUS,
    |                                      ^^^^^^^^^^^ associated item not found in `Status`

error[E0599]: no associated item named `GRPC_STATUS` found for struct `Status` in the current scope
   --> /home/david/.cargo/registry/src/index.crates.io-6f17d22bba15001f/console-api-0.8.1/src/generated/rs.tokio.console.trace.rs:357:48
    |
357 | ...                   tonic::Status::GRPC_STATUS,
    |                                      ^^^^^^^^^^^ associated item not found in `Status`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `console-api` (lib) due to 2 previous errors
```

### Versions

console-api v0.8.1

### Possible solution

Change the dependency specification to `version = "0.12.3"`.

### Would you like to work on fixing this bug?

maybe

hds:
@divergentdave thanks for the report. If you feel like submitting a PR to fix, feel free. In any event I'll try to get a fix and a point release out tomorrow.

hds:
Adding CI tests in #596 to avoid this issue in the future.

PR #596, opened 2024-11-05
chore: check minimal versions on CI

hds:
As reported in issue #592, in our release of `console-api` 0.8.1, we
specified that we required a tonic version of 0.12 in teh Cargo.toml
file. However, since we had generated Rust code with `tonic-build`
0.12.3, we had some code that was only present in `tonic` from 0.12.3
as well.

This error was fixed in #593.

However, this showed a gap in our CI testing, we don't test against
minimum versions of our dependencies to catch this kind of error.

This change adds a CI job to check the minimal versions of our direct
dependencies. We perform a cargo update with the `-Z
direct-minimal-versions` flag and then perform cargo check. This would
have caught the previous error and will catch any new ones of a similar
type.

One downside to this approach is that we must explicitly give a minimum
version for our direct dependencies that is equal to or greater than the
minimum version that any of of transitive dependencies specify for that
same crate. However this check is probably worth the annoyance.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,8 @@ env:
   RUSTUP_MAX_RETRIES: 10
   # Don't emit giant backtraces in the CI logs.
   RUST_BACKTRACE: short
+  # Pin a nightly version for minimal versions
+  rust_nightly: nightly-2024-05-05
 
 jobs:
   check:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -25,6 +25,8 @@ env:
   RUSTUP_MAX_RETRIES: 10
   # Don't emit giant backtraces in the CI logs.
   RUST_BACKTRACE: short
+  # Pin a nightly version for minimal versions
+  rust_nightly: nightly-2024-05-05
```
0xPoe, line 29: May I ask why we need to pin the nightly version here? Is it to avoid breaking changes from Cargo's resolver?

.github/workflows/ci.yml
```diff
@@ -42,6 +44,27 @@ jobs:
...
+        with:
+          toolchain: ${{ env.rust_nightly }}
+      - name: Install cargo-hack
+        uses: taiki-e/install-action@cargo-hack
+      - uses: Swatinem/rust-cache@v2
+      - name: "check -Z direct-minimal-versions"
+        run: |
+          # Remove dev-dependencies from Cargo.toml to prevent the next `cargo update`
+          # from determining minimal versions based on dev-dependencies.
+          cargo hack --remove-dev-deps --workspace
```
0xPoe, line 63: I guess this is for https://github.com/rust-lang/cargo/issues/5657#issuecomment-536920817?

0xPoe (commented):
Thank you!

.github/workflows/ci.yml
hds: Nightly versions may not always work. So we're pinning this to avoid flaky tests depending on the most recent nightly.

.github/workflows/ci.yml
hds: Also just because we don't need to have exact versions for dev dependencies (since it doesn't affect crates that depend on us).

.github/workflows/ci.yml
0xPoe: Got it. Thank you!

0xPoe (approved):
Looks good to me. Thanks! 👍

PR #602, opened 2024-12-11
fix: bump the url to 2.5.4

### LINKED ISSUE #601: RUSTSEC-2024-0421: `idna` accepts Punycode labels that do not produce any non-ASCII when decoded
> `idna` accepts Punycode labels that do not produce any non-ASCII when decoded

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `idna`                      |
| Version             | `0.5.0`                   |
| URL                 | [https://bugzilla.mozilla.org/show_bug.cgi?id=1887898](https://bugzilla.mozilla.org/show_bug.cgi?id=1887898) |
| Date                | 2024-12-09                         |
| Patched versions    | `>=1.0.0`                  |

`idna` 0.5.0 and earlier accepts Punycode labels that do not produce any non-ASCII output, which means that either ASCII labels or the empty root label can be masked such that they appear unequal without IDNA processing or when processed with a different implementation and equal when processed with `idna` 0.5.0 or earlier.

Concretely, `example.org` and `xn--example-.org` become equal after processing by `idna` 0.5.0 or earlier. Also, `example.org.xn--` and `example.org.` become equal after processing by `idna` 0.5.0 or earlier.

In applications using `idna` (but not in `idna` itself) this may be able to lead to privilege escalation when host name comparison is part of a privilege check and the behavior is combined with a client that resolves domains with such labels instead of treating them as errors that preclude DNS resolution / URL fetching and with the attacker managing to introduce a DNS entry (and TLS certificate) for an `xn--`-masked name that turns into the name of the target when processed by `idna` 0.5.0 or earlier.

## Remedy

Upgrade to `idna` 1.0.3 or later, if depending on `idna` directly, or to `url` 2.5.4 or later, if depending on `idna` via `url`. (This issue was fixed in `idna` 1.0.0, but versions earlier than 1.0.3 are not recommended for other reasons.)

When upgrading, please take a moment to read about [alternative Unicode back ends for `idna`](https://docs.rs/crate/idna_adapter/latest).

If you are using Rust earlier than 1.81 in combination with SQLx 0.8.2 or earlier, please also read an [issue](https://github.com/servo/rust-url/issues/992) about combining them with `url` 2.5.4 and `idna` 1.0.3.

## Additional information

This issue resulted from `idna` 0.5.0 and earlier implementing the UTS 46 specification literally on this point and the specification having this bug. The specification bug has been fixed in [revision 33 of UTS 46](https://www.unicode.org/reports/tr46/tr46-33.html#Modifications).

## Acknowledgements

Thanks to kageshiron for recognizing the security implications of this behavior.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2024-0421.html) for additional details.

0xPoe:
close https://github.com/tokio-rs/console/issues/601

As `Remedy` said, because we are not directly dependent on `idna`. So we just need to bump the `url` to 2.5.4 or later.

After bump:

```console
console on  pr/Rustin170506/602:rustin-patch-url-version [$] via 🦀 v1.83.0 
❯ cargo tree --invert -p idna 
idna v1.0.3
└── url v2.5.4
    └── color-eyre v0.6.3
        ├── tokio-console v0.1.13 (/Volumes/t7/code/console/tokio-console)
        └── xtask v0.1.0 (/Volumes/t7/code/console/xtask)
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -511,6 +511,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "displaydoc"
+version = "0.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "97369cbbc041bc366949bc74d34658d6cda5621039731c6310521892a3a20ae0"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "dunce"
 version = "1.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


0xPoe:
CI should be fixed at https://github.com/tokio-rs/console/pull/603.

hds (approved):
Thank you!

0xPoe:
Just rebased on upstream, no other changes. So I will merge it.

PR #603, opened 2024-12-11
refactor: apply `clippy::needless-lifetimes` suggestions

0xPoe:
See: https://github.com/tokio-rs/console/actions/runs/12277528296/job/34257115867?pr=602

I used `cargo clippy --workspace --all-targets --no-deps --fx` to fix it.


hds (approved):
Thanks for fixing these!

hawkw (approved):

Issue #523, opened 2024-02-14
Make better use of space in task detail screen


hds:
### What problem are you trying to solve?

In the task detail screen, there are two panels which split the width of the screen in half. The Task panel and the Waker panel.

The task panel is 7 or 8 lines high (depending on whether the task has a name defined). The Waker panel is just 2 lines high. Additionally, the location of the task can sometimes overflow the width of the Task panel, causing the beginning of the location to not be visible.

See the screenshot below which illustrates how the location can't be seen yet we have plenty of space elsewhere:

![tokio-console-cant_see_location](https://github.com/tokio-rs/console/assets/89589/423c0758-3d3e-4ea7-a1e9-c33f04b6a2bb)

Additionally, if the task doesn't have a name, we don't list the `name` line at all, which results in an empty line at the bottom of the Task panel:

![tokio-console-empty_line](https://github.com/tokio-rs/console/assets/89589/ef87226b-077b-4442-8afe-69c80f2e039b)

### How should the problem be solved?

The space on that panel should be better optimised.

The Waker panel could be split onto more lines. So instead of looking like this:

```
╭Waker─────────────────────────────────────────────────────────────────────────╮
│Current wakers: 1 (clones: 10461, drops: 10460)                               │
│Woken: 5230 times, last woken: 75.831727ms ago                                │
│                                                                              │
│                                                                              │
│                                                                              │
│                                                                              │
│                                                                              │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
```

It could be split onto multiple lines:

```
╭Waker─────────────────────────────────────────────────────────────────────────╮
│Current wakers: 1                                                             |
|  Clones: 10461                                                               |
|  Drops: 10460                                                                │
│Woken: 5230 times,                                                            │
│Last woken: 75.831727ms ago                                                   │
│                                                                              │
│                                                                              │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
```

This would free up more horizontal space. Then the Task panel could be made wider and the Waker panel narrower.

Additionally, the hight should depend on how many lines are actually needed. If the task has no name, then we can make the Task and Waker panels one line shorter.

For bonus points, the Location field in the Task panel could be split onto multiple lines (up to some reasonable maximum like 2 or 3) in case it still doesn't fit into the space provided. If done, it would make sense to submit this change as a separate PR.

This could end up with the two panels looking like this:

```
[...]

PR #614, opened 2025-03-16
fix(console): add dynamic constraints layout in task details screen

vrn21-arcanist:
Changes I've made is for accommodating loong location names; ive changed the .. to the long name; it widens the task rectangle if its longer than the default (50%) and goes to the next line if its longer than what we can accommodate in a single line. This fixes issue #523  have attached screenshots of the same in the PR, before and after changes

These screenshots are what was the UI before my changes.
Screenshot 2025-03-16 at 4 15 10 PM

and these are the screenshots after my changes.
Screenshot 2025-03-16 at 4 14 25 PM
Screenshot 2025-03-16 at 4 14 19 PM
Screenshot 2025-03-16 at 4 14 15 PM

### DIFF
--- a/tokio-console/src/view/task.rs
+++ b/tokio-console/src/view/task.rs
@@ -127,7 +127,7 @@ impl TaskView {
             )
         };
 
-        let stats_area = Layout::default()
+        let stats_area_check = Layout::default()
             .direction(layout::Direction::Horizontal)
             .constraints(
                 [
@@ -137,6 +137,51 @@ impl TaskView {
                 .as_ref(),
             )
             .split(stats_area);
+
+        let mut location_lines_vector: Vec<String> = vec![];
+        let title = "Location: ";
+        let location_max_width = stats_area_check[0].width as usize - 2 - title.len(); // NOTE: -2 for the border
+        let stats_area = if task.location().len() > location_max_width {
+            let max_width_stats_area = area.width - 45;
+            if max_width_stats_area < task.location().len() as u16 {
+                location_lines_vector = task
+                    .location()
+                    .to_string()
+                    .chars()
+                    .collect::<Vec<char>>()
+                    .chunks(max_width_stats_area as usize)
+                    .map(|chunk| chunk.iter().collect())
+                    .collect();
+
+                let area_needed_to_render_location = task.location().len() as u16;
+                Layout::default()
+                    .direction(layout::Direction::Horizontal)
+                    .constraints(
+                        [
+                            layout::Constraint::Min(area_needed_to_render_location + 15), //Note: 15 is the length of "| Location:     |"
+                            layout::Constraint::Min(32),
+                        ]
+                        .as_ref(),
+                    )
+                    .split(stats_area)
+            } else {
+                let area_needed_to_render_location = task.location().len() as u16;
+                location_lines_vector.push(task.location().to_string());
+                Layout::default()
+                    .direction(layout::Direction::Horizontal)
+                    .constraints(
+                        [
+                            layout::Constraint::Min(area_needed_to_render_location + 15), //Note: 15 is the length of "| Location:     |"
+                            layout::Constraint::Min(32),
+                        ]
+                        .as_ref(),
+                    )
+                    .split(stats_area)
+            }
+        } else {
+            location_lines_vector.push(task.location().to_string());
+            stats_area_check
+        };
 
         // Just preallocate capacity for ID, name, target, total, busy, and idle.
         let mut overview = Vec::with_capacity(8);
@@ -152,17 +197,13 @@ impl TaskView {
 
         overview.push(Line::from(vec![bold("Target: "), Span::raw(task.target())]));
 
-        let title = "Location: ";
-        let location_max_width = stats_area[0].width as usize - 2 - title.len(); // NOTE: -2 for the border
-        let location = if task.location().len() > location_max_width {
-            let ellipsis = styles.if_utf8("\u{2026}", "...");
-            let start = task.location().len() - location_max_width + ellipsis.chars().count();
-            format!("{}{}", ellipsis, &task.location()[start..])
-        } else {
-            task.location().to_string()
-        };
-
-        overview.push(Line::from(vec![bold(title), Span::raw(location)]));
+        let first_line = location_lines_vector[0].clone();
+        location_lines_vector.remove(0);
+        let location_vector = vec![bold(title), Span::raw(first_line)];
+        overview.push(Line::from(location_vector));
+        for line in location_lines_vector {
+            overview.push(Line::from(Span::raw(format!("    {}", line))));
+        }
 
         let total = task.total(now);
 
@@ -185,28 +226,37 @@ impl TaskView {
 
         let mut waker_stats = vec![Line::from(vec![
             bold("Current wakers: "),
-            Span::from(format!("{} (", task.waker_count())),
-            bold("clones: "),
-            Span::from(format!("{}, ", task.waker_clones())),
-            bold("drops: "),
-            Span::from(format!("{})", task.waker_drops())),
+            Span::from(format!("{} ", task.waker_count())),
         ])];
+        let waker_stats_clones = vec![
+            bold("  Clones: "),
+            Span::from(format!("{}, ", task.waker_clones())),
+        ];
 
-        let mut wakeups = vec![
+        let waker_stats_drops = vec![
+            bold("  Drops: "),
+            Span::from(format!("{}", task.waker_drops())),
+        ];
+
+        let wakeups = vec![
             bold("Woken: "),
             Span::from(format!("{} times", task.wakes())),
         ];
 
+        let mut last_woken_line = vec![];
+
         // If the task has been woken, add the time since wake to its stats as well.
         if let Some(since) = task.since_wake(now) {
-            wakeups.reserve(3);
-            wakeups.push(Span::raw(", "));
-            wakeups.push(bold("last woken: "));
-            wakeups.push(styles.time_units(since, view::DUR_LIST_PRECISION, None));
-            wakeups.push(Span::raw(" ago"));
+            last_woken_line.reserve(3);
+            last_woken_line.push(bold("Last woken: "));
+            last_woken_line.push(styles.time_units(since, view::DUR_LIST_PRECISION, None));
+            last_woken_line.push(Span::raw(" ago"));
         }
 
+        waker_stats.push(Line::from(waker_stats_clones));
+        waker_stats.push(Line::from(waker_stats_drops));
         waker_stats.push(Line::from(wakeups));
+        waker_stats.push(Line::from(last_woken_line));
 
         if task.self_wakes() > 0 {
             waker_stats.push(Line::from(vec![
--- a/tokio-console/src/view/task.rs
+++ b/tokio-console/src/view/task.rs
@@ -139,19 +139,12 @@ impl TaskView {
             )
         };
 
-        let stats_constraints = if location_lines_vector.len() > 1 {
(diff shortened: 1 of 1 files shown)


vrn21:
#523

hds:
@vrn21 Thanks for working on this topic! So far this PR looks good. However, with a fixed number of rows, when you go to 2 lines for the location text, the "Idle" time is getting dropped off, so that needs to be resolved too. I think that there are some hints in the original issue regarding how to approach that issue.

vrn21-arcanist:
Thank you for looking into this @hds  and Sorry for the delay,

 but i'm still not sure how i'm supposed to move forward with this, i cant really determine what the height of the task_area is, before hand as in the current code, number of lines extra needed can be only computed after the each of the constraint is set. Could you help on this?

hds:
In the code that exists, the areas are computed up front, based on the constraints provided. However, those areas aren't actually used until the `Paragraph` for each of the task overview and the waker stats are actually rendered at the bottom.

If the creation of the task and waker "widgets" (that's the variable name), is performed before the area is split on constraints, then we can already compute the number of lines that each one needs and use that as input to the `Constraints::Length` for the `stats_area` which is where they sit.

Does that help?

vrn21-arcanist:
Sorry again for the delay, Does this work @hds?

I've changed the logic to check whether locations is longer than what is viewable in a single line, before hand, assigning constraints (both vertical and horizontal) afterwards.
Screenshot 2025-03-21 at 7 42 09 PM
Screenshot 2025-03-21 at 7 41 27 PM

hds:
This is looking good! There are a few style and layout things, but the functionality appears correct. I'll get you a proper review of this on Monday.

Thanks for your work!

tokio-console/src/view/task.rs
```diff
@@ -152,17 +247,13 @@ impl TaskView {
...
-            let start = task.location().len() - location_max_width + ellipsis.chars().count();
-            format!("{}{}", ellipsis, &task.location()[start..])
-        } else {
-            task.location().to_string()
-        };
-
-        overview.push(Line::from(vec![bold(title), Span::raw(location)]));
+        let first_line = location_lines_vector[0].clone();
+        location_lines_vector.remove(0);
+        let location_vector = vec![bold(location_heading), Span::raw(first_line)];
```
hds: I think that we can use the reference here, can't we? We don't need to clone or manipulate the Vec.

```suggestion
        let location_vector = vec![bold(location_heading), Span::raw(location_lines_vector[0])];
```

tokio-console/src/view/task.rs
```diff
@@ -152,17 +247,13 @@ impl TaskView {
...
+        overview.push(Line::from(location_vector));
+        for line in location_lines_vector {
+            overview.push(Line::from(Span::raw(format!("    {}", line))));
+        }
```
hds: Then here we can use the vec slice:

```suggestion
        for line in location_lines_vector[1..] {
            overview.push(Line::from(Span::raw(format!("    {}", line))));
        }
```

tokio-console/src/view/task.rs
```diff
@@ -68,76 +67,172 @@ impl TaskView {
...
+                    layout::Constraint::Percentage(50),
+                    layout::Constraint::Percentage(50),
+                ]
+                .as_ref(),
+            )
+            .split(area);
+
+        let location_heading = "Location: ";
+        let location_max_width = stats_area_check[0].width as usize - 2 - location_heading.len(); // NOTE: -2 for the border
+        let max_width_stats_area = area.width - 45;
```
hds: Can we give the `45` a name? I'm guessing this is the minimum waker area needed?

tokio-console/src/view/task.rs
```diff
@@ -68,76 +67,172 @@ impl TaskView {
...
+        let mut location_lines_vector: Vec<String> = task
+            .location()
+            .to_string()
+            .chars()
+            .collect::<Vec<char>>()
+            .chunks(max_width_stats_area as usize)
+            .map(|chunk| chunk.iter().collect())
+            .collect();
+        let no_of_lines_extra_required_to_accomadate_location = location_lines_vector.len() - 1;
```
hds: These very long variable names don't really fit in with the style of the rest of the codebase, we probably don't need the extra variable at all.

Instead, why don't we extract out the task stats height into a variable:

```suggestion
        // Id, Name, Target, Location (multiple), total, busy, scheduled, and idle times + top/bottom borders
        let task_stats_height = 9 + location_lines_vector.len();
```

tokio-console/src/view/task.rs
```diff
@@ -68,76 +67,172 @@ impl TaskView {
...
             controls_area,
             stats_area,
             poll_dur_area,
             scheduled_dur_area,
             fields_area,
             warnings_area,
-        ) = if warnings.is_empty() {
-            let chunks = Layout::default()
-                .direction(layout::Direction::Vertical)
+        ) = if task.location().len() > location_max_width {
```
hds: We definitely don't want to add another layer of `if-else` statements here. I think that be specifying the task stats height outside, we can leave this like it was (just depending on the presense of warnings).

tokio-console/src/view/task.rs
```diff
@@ -34,7 +34,6 @@ impl TaskView {
     pub(crate) fn update_input(&mut self, _event: input::Event) {
         // TODO :D
     }
-
```
hds: I don't think we want to remove this line.

tokio-console/src/view/task.rs
```diff
@@ -152,17 +247,13 @@ impl TaskView {
...
-            task.location().to_string()
-        };
-
-        overview.push(Line::from(vec![bold(title), Span::raw(location)]));
+        let first_line = location_lines_vector[0].clone();
+        location_lines_vector.remove(0);
+        let location_vector = vec![bold(location_heading), Span::raw(first_line)];
+        overview.push(Line::from(location_vector));
+        for line in location_lines_vector {
+            overview.push(Line::from(Span::raw(format!("    {}", line))));
```
hds: I think we need to add more spaces here so that the beginning of the text aligns with the first line location text.

If you leave only 4 spaces here, then the chunks for the 2nd line and onward should be bigger than the first chunk (which is also an option).

hds (changes requested):
This is looking good, thanks! There are some changes I've suggested.

vrn21-arcanist:
@hds i've made the changes you've told to

tokio-console/src/view/task.rs
```diff
@@ -68,6 +68,17 @@ impl TaskView {
...
+        let max_width_stats_area = area.width - 45; //NOTE: 45 is min width needed, this is a calculated number according to the string: 'Last woken: 75.831727ms ago' but with some more extra pixels.
+        let location_lines_vector: Vec<String> = task
+            .location()
+            .to_string()
+            .chars()
+            .collect::<Vec<char>>()
+            .chunks(max_width_stats_area as usize)
+            .map(|chunk| chunk.iter().collect())
+            .collect();
+        // Id, Name, Target, Location (multiple), total, busy, scheduled, and idle times + top/bottom borders
```
hds, line 82: I think I wrote something confusing, it's include the task stats height as a variable here:
```suggestion
        // Id, Name, Target, Location (multiple), total, busy, scheduled, and idle times + top/bottom borders
        let task_stats_height = 9 + location_lines_vector.len() as u16;
```

tokio-console/src/view/task.rs
```diff
@@ -83,7 +94,7 @@ impl TaskView {
                         // controls
                         layout::Constraint::Length(controls.height()),
                         // task stats
-                        layout::Constraint::Length(10),
+                        layout::Constraint::Length((10 + location_lines_vector.len() - 1) as u16),
```
hds: And use that variable here:

```suggestion
                        layout::Constraint::Length(task_stats_height),
```

tokio-console/src/view/task.rs
```diff
@@ -105,7 +116,7 @@ impl TaskView {
                         // warnings (add 2 for top and bottom borders)
                         layout::Constraint::Length(warnings.len() as u16 + 2),
                         // task stats
-                        layout::Constraint::Length(10),
+                        layout::Constraint::Length((10 + location_lines_vector.len() - 1) as u16),
```
hds: And here too:

```suggestion
                        layout::Constraint::Length(task_stats_height),
```

tokio-console/src/view/task.rs
```diff
@@ -127,16 +138,30 @@ impl TaskView {
...
+                .direction(layout::Direction::Horizontal)
+                .constraints(
+                    [
+                        layout::Constraint::Percentage(50),
+                        layout::Constraint::Percentage(50),
+                    ]
+                    .as_ref(),
+                )
+                .split(stats_area)
+        };
```
hds: We can simplify this a bit by just setting different constraints:

```suggestion
        let stats_constraints = if location_lines_vector.len() > 1 {
            [
                // 15 is the length of "| Location:     |"
                layout::Constraint::Min(area_needed_to_render_location + 15),
                layout::Constraint::Min(32),
            ]
        } else {
            [
                layout::Constraint::Percentage(50),
                layout::Constraint::Percentage(50),
            ]
        };
        let stats_area = Layout::default()
            .direction(layout::Direction::Horizontal)
            .constraints(stats_constraints.as_ref())
            .split(stats_area);
```

hds (changes requested):
One point here where I think I wasn't clear regarding what I meant. And one other thing that I didn't notice before, sorry about that.

vrn21-arcanist:
@hds i've updated the latest changes requested

tokio-console/src/view/task.rs
```diff
@@ -127,15 +139,22 @@ impl TaskView {
             )
         };
 
+        let stats_constraints = if location_lines_vector.len() > 1 {
+            let area_needed_to_render_location = task.location().len() as u16;
+            [
+                // 15 is the length of "| Location:     |"
+                layout::Constraint::Min(area_needed_to_render_location + 15),
```
hds: Variable names are a bit long still. (you can hit "apply suggestion" to get the change directly applied)

```suggestion
            let location_len = task.location().len() as u16;
            [
                // 15 is the length of "| Location:     |"
                layout::Constraint::Min(location_len + 15),
```

hds: What happens when the 2 `Min` constraints add up to more than the available width?

vrn21: from [this](https://github.com/ratatui/ratatui/pull/393), i think the view would overflow according to the length the view has.

hds: I think the `if` statement is incorrect then (and also back to front). In fact, I think that if instead the constraints look just like this then it might be enough (always, no if statement at all). Could you try and show some screenshots?

```rust
            [
                layout::Constraint::Min(location_lines_vector[0].len()),
                layout::Constraint::Min(32),
            ]
```

hds (commented):
What about this idea?

vrn21-arcanist:
Sorry for the delay again!

Yes, you're right, we could remove that unecessary if block. This doesnt really affect the view.

Attaching the current versions:

Screenshot 2025-03-28 at 12 22 20 AM
Screenshot 2025-03-28 at 12 22 34 AM
Screenshot 2025-03-28 at 12 22 46 AM
Screenshot 2025-03-28 at 12 23 47 AM
Screenshot 2025-03-28 at 12 29 41 AM
 
I'm commiting the changes( removing if block), i dont think this has any other problems with it.

vrn21-arcanist:
@hds seems fine?

hds (approved):
Looks good. Thank you for all your work on this issue!

PR #623, opened 2025-04-09
feat: add support for vsock connections

domenkozar:
I've hidden it behind a feature flag, since it's kind of uncommon.

### DIFF
--- a/tokio-console/src/config.rs
+++ b/tokio-console/src/config.rs
@@ -1,6 +1,7 @@
 use crate::state::tasks::Task;
 use crate::view::Palette;
 use crate::warnings;
+use cfg_if::cfg_if;
 use clap::builder::{PossibleValuesParser, TypedValueParser};
 use clap::{ArgAction, ArgGroup, CommandFactory, Parser as Clap, Subcommand, ValueHint};
 use clap_complete::Shell;
@@ -34,6 +35,9 @@ pub struct Config {
     /// On Unix platforms, this may also be a URI with the `file` scheme that
     /// specifies the path to a Unix domain socket, as in
     /// `file://localhost/path/to/socket`.
+    ///
+    /// When the `vsock` feature is enabled, this may also be a URI with the `vsock` scheme that
+    /// specifies a vsock connection, as in `vsock://2:6669` to connect to CID 2 port 6669.
     ///
     /// [default: http://127.0.0.1:6669]
     #[clap(value_hint = ValueHint::Url)]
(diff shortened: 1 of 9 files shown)


hds:
Hey! This looks good in general. Would you mind adding a bit more background and justification to your PR message (that will go into the commit history).

Basically because I didn't know what a vsock was until today, so I figure the reference would be worthwhile. 😅

domenkozar:
@hds done :)

domenkozar:
Friendly ping :)

hds:
@domenkozar Thanks for the ping. I've just had another look through the PR, however don't we need to add vsock support into tokio-console as well? Otherwise this feels like only half the feature.

domenkozar:
@hds something like this? :)

tokio-console/src/config.rs
```diff
@@ -528,11 +531,20 @@ impl Config {
             .clone();
         match target_addr.scheme_str() {
             Some("file" | "http" | "https") => {}
+            #[cfg(feature = "vsock")]
```
hds, line 535: I think that in the case that vsock is provided, but the feature isn't enabled, we should provide a concrete error message informing the user that `tokio-console` wasn't compiled with the `vsock` feature, and so it isn't available.

tokio-console/src/config.rs
```diff
@@ -528,11 +531,20 @@ impl Config {
...
+            Some("vsock") => {}
             _ => {
+                #[cfg(feature = "vsock")]
+                return Err(color_eyre::eyre::eyre!(
+                "invalid scheme for target address {:?}, must be one of 'file', 'http', 'https', or 'vsock'",
```
hds: Rather than repeating the whole error co traction, we can probably use an `if_cfg!` just to create the list of allowed protocols.

tokio-console/src/conn.rs
```diff
@@ -112,6 +114,48 @@ impl Connection {
...
+                        endpoint
+                            .connect_with_connector(tower::service_fn(move |_| {
+                                VsockStream::connect(vsock_addr).map_ok(TokioIo::new)
+                            }))
+                            .await?
+                    }
+                    #[cfg(not(feature = "vsock"))]
+                    Some("vsock") => {
+                        return Err(
+                            "vsock is not supported on this platform or feature is not enabled"
```
hds: We can probably limit this to saying the feature isn't enabled, since that's why we've hit this code path.

hds (changes requested):
Looks good! Just a few small things regarding error handling (mostly in the case where the new feature isn't enabled).

domenkozar:
PTAL :)

hds (approved):
Looks good. Just a couple of CI checks that are failing. Thanks for all your work on this PR!

hds:
@domenkozar Merged! Thanks for your contribution!

I hope to get a release of the crates out in the next week. There are just a few little changes that we'd like to make first.

PR #626, opened 2025-04-13
chore: update tokio to 1.38.2

hds:
There was a soundness defect in Tokio's broadcast channel. This change
updates to the oldest version of Tokio with the soundness defect fixed
(1.38.2 - 1.38 is the current oldest LTS version). This is preferable to
updating to the latest version of Tokio, as that would force users of
`console-subscriber` to use that version or later too.

Tokio 1.38.2: https://github.com/tokio-rs/tokio/releases/tag/tokio-1.38.2


hds:
Tokio has already been updated past this version.

PR #641, opened 2025-09-29
Fix Nix builds and update the lock file

ipetkov:
Should make it a little bit nicer for Nix/devshell users


ipetkov:
Depends on https://github.com/tokio-rs/console/pull/640 which should fix CI runs

seanmonstar (approved):

PR #658, opened 2026-03-27
chore: raise required compiler to rust 1.84

dtolnay:
Rust 1.84.0 (released in January 2025) stabilized the MSRV-aware resolver in Cargo: https://blog.rust-lang.org/2025/01/09/Rust-1.84.0/#cargo-considers-rust-versions-for-dependency-version-selection. This makes it unnecessary to disallow using `tokio-console` with more recent versions of `clap`.

Specifically, I am interested in building `tokio-console` against `clap` 4.6 &mdash; using a much newer compiler than 1.74.0 obviously. But this is prohibited by the current dependency specification.


dtolnay (commented):
"CI / Lints" job fails, but there are no new lints beyond the ones fixed in https://github.com/tokio-rs/console/pull/657.

hds (approved):
This seems reasonable. Thanks!

